PR #1, opened 2015-05-21
User session is extended for SchedulerStateRest#getSchedulerStatus only

lpellegr:
The idea is to call renewLeaseForClient in all methods exposed in the REST API.
It seems that this call can be performed in the existing checkAccess method that
is already invoked in public methods.


lpellegr:
@fviale @youribonnaffe could you please review the patch?

lpellegr:
@fviale @youribonnaffe could you please review the patch? 

I have added a test to check that renewLeaseForSession is called in the methods from SchedulerStateRest. The last method is assumed to be tested independently.

PR #3, opened 2015-05-28
SCHEDULING-2279 Node events are slow to process in Resource Manager

youribonnaffe:
The looping over nodes in RMListenerProxy resulted in many ProActive calls/use
of futures slowing down the dispatch of events.
Now only one call is performed and the lookup of nodes is done on RMInitialState
side.


lpellegr:
:+1:

PR #5, opened 2015-06-03
SCHEDULING-2022 Broken number of tasks update in job info

youribonnaffe:
The commit 6c6e8b8feeb4e5f3de08cf0e9a8734bee1dc1bee broke the update
of numberOfTasks (pending, running, finished) stored in JobInfo.

When a task finishes, those fields should be updated even if only task
events are triggered. Otherwise those fields will only be update upon
job termination when a job finished event is triggered for instance.

This was causing org.ow2.proactive_grid_cloud_portal.scheduler.SchedulerStateRest#listJobs
to return a non updated job state whereas org.ow2.proactive_grid_cloud_portal.scheduler.SchedulerStateRest#revisionAndjobsinfo
returned a update state because it reads data from DB.


youribonnaffe:
@mauriciojost Could take a look please?

mauriciojost:
Seems okay to me, approved.

youribonnaffe:
Thank you @mauriciojost

youribonnaffe:
Thank you so much @mauriciojost

PR #2251, opened 2015-06-22
Improve Scheduling Time (Speed branch)

lpellegr:
Below is a list of tasks to complete before merging the pull request into the master (please update the list when a task is finished or if a new one is required):

  - [x] Release jsr223-nativeshell version 0.3
  - [x] Move example classes from `scheduler-node` to a dedicated module
    - [x] Remove no longer used classes
  - [x] Delay input/output files replacement at last moment (during task execution not when creating job)
  - [x] Document/Cleanup dropped features like job classpath, job archive, Java API
  - [x] Check that no tmp files are left when using it, when running tests
  - [x] Working dir feature for non forked tasks?
  - [x] Native task output in case of error is a bit ugly (stack traces)
  - [x] Implement fork environment
    - [x] Add support `forkEnvironment` in `task` element
    - [x] Check that old forkEnvironment definition is still working (backward compatibility)
  - [x] Single setting to enable/disable task forking
  - [x] Remove JobEnvironment/JobClasspath
    - [x] Delete Job archives
  - [x] Remove workingDir attribute from staticCommand element
  - [x] Remove baseEnvironment from forkEnvironment
  - [x] Remove Result Preview Class
  - [x] Update Studio
    - [x] Add support for forkEnvironment 
    - [x] Remove jobClasspath
    - [x] Remove staticCommand
    - [x] Remove Result Preview Class
  - [x] Search and fix TODOs related to this pull request 
  - [x] Fix broken tests
    - [x]  TestStaticWorkingDirCommand
    - [x]  TestMultipleHostRequest
    - [x] TestProcessTreeKiller
    - [x] TestModifyPropagatedVariables

**TODO** last actions before merging into master:
  - [x] Push programming Speed branch
  - [x] Uncomment line 24 from `build.grade` to make `mavenLocal` repository optional 
  - [x] Include latest changes from master
  - [x] Rebase to hide not meaningful commit messages
  - [ ] Upload last version of schema 3.3 on ActiveEon website


lpellegr:
Is there a reason for having all Job examples in `scheduler-node` module? It seems there are used only in `scheduling-server` or in `studio`. Could I move them in a dedicated module named `scheduler-examples`?

fviale:
I don't see a reason whatsoever.

youribonnaffe:
I think the reason is to just to have them exposed in the nodes the easiest way possible. We definitely can clean up the non used Java tasks.

fviale:
Ah yes like EmptyTask or things like that, now I remember.

lpellegr:
I noticed that the following classes seem not to be used from the [existing examples](https://github.com/ow2-proactive/scheduling/tree/Speed/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/examples):

  - DenoisePreview
  - JoinPreview
  - NativeTestWithRandomDefault
  - SchedulerClientEventsLogger
  - TestTask

I will remove them if nobody is against that.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
```diff
@@ -32,1809 +32,290 @@
...
-    protected List<OutputSelector> outputFiles;
+@ActiveObject
+public class TaskLauncher {
 
-    // buffered string to store logs destined to the client (for e.g. datapspaces error/warn messages)
-    private StringBuffer clientLogs;
+    private static final Logger logger = Logger.getLogger(TaskLauncher.class);
 
-    protected Decrypter decrypter = null;
+    private ProActiveForkedTaskLauncherFactory factory = new ProActiveForkedTaskLauncherFactory();
```
lpellegr: Is it required to initialize it for all TaskLauncher instances? it seems that the factory should be passed in the constructor?

lpellegr:
I was looking at files in `scheduling/scheduler/scheduler-api/src/main/resources/org/ow2/proactive/scheduler/common/xml/schemas/jobdescriptor` for moving where the `forkEnvironment` may be defined and I had some questions:

1. Is it ok to create a new version `7.0`? or should I use `3.3`?
2. What do you think about allowing users to define a `forkEnvironment` at the job level? this way if the fork environment is the same for several tasks, the users does not have to repeat the same element multiple times. Besides, if the `forkEnvironment` is defined at the job and task level, then the more specific one (task level) will be used for the task.
3. What is the purpose of the `dev` version? is it a copy of the latest version that avoids to update all URIs in existing XML files when a new version is published?

youribonnaffe:
1. I'm not sure, because we rarely change the job schema, so even if we start at 7, there might not be might a version schema with PWS 8. But I would not mind 7 too.
2. IMHO fork environment is rarely used so I would not do it for now.
3. Not sure, but maybe the reason you suggested is correct :D

youribonnaffe:
I would suggest to do  _Deprecate pre/post script and allow a list of executables instead?_ later.

lpellegr:
I have some new questions:

  1. What is the purpose of property `pa.scheduler.task.allowjavatasks`? I mean is it used by someone? in which context?
  2. Regarding the task *Single setting to enable/disable task forking*, is it not already handled by the scheduler property `pa.scheduler.task.scripttasks.fork`? should we add tests? I also suggest to rename the property to `pa.scheduler.task.fork.force` or simply to `pa.scheduler.task.fork`. What do you think?
  3. I noticed that new configuration files have been put in `config/scheduler/forkedJavaTask`:

```
├── forkedTask.java.policy
├── forkedTask-log4j
└── forkedTask-paconf.properties
```

If you are not against the following, I will move files as explained below in order to be consistent with other configurations:

  - `forkedTask.java.policy` to `config/nodes/task-forked.java.policy`
  - `forkedTask-log4j` to `config/log/task-forked.properties`
  - `forkedTask-paconf.properties` to `config/nodes/settings-task-forked.ini`

youribonnaffe:
1. The property `pa.scheduler.task.allowjavatasks` is used to forbid the usage of non forked Java tasks (that is you must set a forkEnvironment tag in your workflow).

2. I agree, we can remove  `pa.scheduler.task.allowjavatasks` and  `pa.scheduler.task.scripttasks.fork` (used to make the script tasks forked) and introduce a new property. `pa.scheduler.task.fork` looks fine to me, the default would be true, to make all tasks forked by default.

3. actually those files are not needed anymore in the speed branch so we can even remove them :)

lpellegr:
While looking at actions to perform for task *Remove JobEnvironment/JobClasspath*, Youri and I noticed that `rest-client` makes use of JobClasspath because of Job archives. We were wondering if it still make sense to support Job archives since there is the studio to create/import/export jobs. Besides, it seems there is no tool nor documentation to create such archives.

@bamedro, @fviale What do you think about :fire: job archives?

fviale:
I think there was two basic use for job archive :
- it allows to pass along scripts when you submit jobs (otherwise, the
script must be inlined)
- it indeed allows to add jars to the Job Classpath from the client side
(jars are copied inside the archive, and unzipped on the server side, with
a TaskClassServer started on top).

lpellegr:
@fviale According to what I have understood, the goal is to replace job classpath by paths defined in classpath variable from a fork environment. These paths are supposed to reference files in USERSPACE that will be copied automatically to compute nodes.

For this reason, I don't think that job archives are still required regarding point 2. However, for point 1 I have no idea. Is it not possible to simply put the scripts in the classpath? 

@youribonnaffe please correct me if I am wrong.

fviale:
If this is for the speed branch, then I guess it makes sense to remove it.

Similarly, I think that we should allow referencing scripts located in the
user space.

youribonnaffe:
It would be possible, using input files to copy the scripts to the local space and then execute them by using the path to the current folder (./myscript).
There is still the issue that the script is read upon submission...

lpellegr:
@youribonnaffe could you please explain the issue about read upon submission?

youribonnaffe:
When a workflow is submitted, if scripts are defined as files (not inline code), those scripts will be read upon submission, i.e when using CLI, the client must have the scripts locally accessible. When using the Scheduler portal, the scripts need to exist on the server side... because the REST server will perform the submission...

fviale:
But this will not work for selection scripts.

lpellegr:
@fviale why ?

fviale:
Because selection scripts are not implied by the input files transfer. They
are executed by the Resource Manager directly on multiple nodes, before the
resource is selected and the Task Launcher deployed (which does the file
transfer)

youribonnaffe:
Do we need to give access to the user to dataspaces or are input/output files enough?
Given we try to avoid to start a ProActive runtime in the forked task, we won't have access to dataspaces. However we could expose the VFS URLs and somehow (I don't really know how), let the task starts a ProActive runtime on its own and connect to dataspaces.
Right now, localspace, userspace, inputspace, outputspace and globalspace objects contains a VFS URL (as a string).

fviale:
I think this behavior is fine, direct use of dataspaces from a task is very
rare, so it's understandable that the user will start a ProActive runtime
in that case. Additionnaly, it will in most cases be directly available on
a shared file system.

lpellegr:
@bamedro @fviale When an environment variable is defined for a ForkEnvironment in the studio, it is possible to define whether is it appended and if a specific character should be appended. Youri and I have no idea about the usage of such a feature. Do you remember a specific usecase? could it be removed?

mebubo:
:beers:

Issue #2261, opened 2015-07-09
Add CommonAttributes properties to Job toString() (debug)


fviale:
Currently, properties inherited from superclass CommonAttributes is not printed in the Job debug toString() method.

fviale:
done

PR #2262, opened 2015-07-09
Add CommonAttributes properties to Job toString() (debug) #2261

fviale:


lpellegr:
The change seems ok for me. One thing I noticed is that the formatting is not taken into account (left spaces for collection elements) but I don't if it matters since the method is used for debugging purposes   only if I am not wrong.

fviale:
I've done the formatting by using the gradle target format (which was maybe a mistake).

lpellegr:
I was speaking about the String that is produced.

PR #2264, opened 2015-07-10
Fix an issue with precious logs option which is no longer enabled by …

lpellegr:
…default

Precious logs option is no longer enabled by default on tasks. Consequently,
when a user tries to retrieve full logs for a task, an exception was occuring.
This is because the file containing logs is not found.

The workaround consists of retrieving truncated logs from the database if a task
was not executed using precious logs.


lpellegr:
@youribonnaffe Could you please have a look?

Issue #2265, opened 2015-07-16
Database deadlock with concurrent job insertions and deletions


fviale:
When using automatic job remove, it may happen that the mechanism tries to remove an already removed job. When this occur, the following error happens.

This problem completely paralyses the scheduler, which must be restarted in order to work again.

``` 
[2015-07-09 12:11:28,951 INFO     o.o.p.s.c.SchedulingService] job 13490 removing
[2015-07-09 12:11:28,951 INFO     o.o.p.s.c.SchedulingService] job 13492 removing
[2015-07-09 12:11:28,951 INFO     o.o.p.s.c.SchedulingService] job 13493 removing
[2015-07-09 12:11:28,951 INFO     o.o.p.s.c.SchedulingService] job 13491 removing
[2015-07-09 12:11:28,951 INFO     o.o.p.s.c.SchedulingService] job 13494 removing
[2015-07-09 12:11:28,951 INFO             o.o.p.s.u.JobLogger] job 13490 does not exist
[2015-07-09 12:11:28,952 INFO             o.o.p.s.u.JobLogger] job 13494 does not exist
[2015-07-09 12:11:28,951 INFO             o.o.p.s.u.JobLogger] job 13491 does not exist
[2015-07-09 12:11:28,951 INFO             o.o.p.s.u.JobLogger] job 13493 does not exist
[2015-07-09 12:11:28,951 INFO             o.o.p.s.u.JobLogger] job 13492 does not exist
[2015-07-09 12:11:33,079 INFO         o.o.p.c.SessionsCleaner] cleaning session started, 0 existing session(s) 
[2015-07-09 12:11:33,079 INFO         o.o.p.c.SessionsCleaner] cleaning session ended, 0 session(s) removed
[2015-07-09 12:11:59,285 WARN    o.h.e.j.s.SqlExceptionHelper] SQL Error: 30000, SQLState: 40001
[2015-07-09 12:11:59,285 ERROR   o.h.e.j.s.SqlExceptionHelper] A lock could not be obtained due to a deadlock, cycle of locks and waiters is:
Lock : TABLE, FORKED_JAVA_TASK_DATA, Tablelock
[...]

PR #2270, opened 2015-07-22
Concurrent operations with the DB produce deadlocks

lpellegr:
This pull request is mainly to produce a new release 6.1.1 that fixes issue #2265.

@fviale @youribonnaffe Could you please have a look?


youribonnaffe:
Looks ok to me,although it might slow down submission (but it will probably still be quite fast and sufficient for our use cases)

PR #2272, opened 2015-07-27
Avoid database deadlock upon job termination/job creation

youribonnaffe:
This deadlock behavior was observed during manual stress tests of the Scheduler.
It seems related to #2265 however the synchronized methods in SchedulerDBManager
are still required.

The task executable (i.e a script) is now stored inside TaskData, similar to
pre/post/clean scripts. ScriptTaskData has then been removed and it seems to help
preventing the database deadlock that was observed when submitting and scheduling
a lot of tasks.

Pre/post/executable/clean scripts are mapped with a OneToOne relation with the task
whereas selection scripts are mapped with a OneToMany relation. Since the removal of
scripts has been simplified (to avoid the in query, that breaks with a lot of tasks),
selection scripts needs to be stored in a different table. Otherwise all scripts linked
to a task in ScriptData are considered as selection scripts for this task.

TaskData needs to be updated as well when removing scripts (removeJobScripts) otherwise
recovery fails (observed with TestIfTaskRestore).


youribonnaffe:
@lpellegr related to the recent fix around DB deadlock

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -1470,13 +1460,13 @@ private ExecutableContainer loadExecutableContainer(Session session, InternalTas
             ExecutableContainer container = null;
 
             if (task.getClass().equals(InternalForkedScriptTask.class)) {
```
lpellegr: Is it not possible to merge the body of the two conditions?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
```diff
@@ -185,7 +190,7 @@ public void setAdditionalClasspath(List<String> additionalClasspath) {
     @Cascade(org.hibernate.annotations.CascadeType.ALL)
     @OneToOne(fetch = FetchType.LAZY)
     @JoinColumn(name = "ENV_SCRIPT_ID")
-    @ForeignKey(name = "none")
+    @ForeignKey(name = "none") // to be able to remove script without removing task
```
lpellegr: Is it not possible to put a more meaningful name? why "none"?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
```diff
@@ -196,6 +201,19 @@ public void setEnvScript(ScriptData envScript) {
     }
 
     @Cascade(org.hibernate.annotations.CascadeType.ALL)
+    @OneToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "SCRIPT_ID")
+    @ForeignKey(name = "none") // to be able to remove script without removing task
```
lpellegr: Same question as before.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
```diff
@@ -389,23 +411,27 @@ static TaskData createTaskData(JobData jobRuntimeData, InternalTask task) {
...
             }
             taskData.setSelectionScripts(scripts);
         }
+        if (task.getExecutableContainer() != null) {
+            taskData.setScript(ScriptData.createForScript(
+                    ((ScriptExecutableContainer) task.getExecutableContainer()).getScript(), taskData));
+        }
         if (task.getPreScript() != null) {
-            taskData.setPreScript(ScriptData.createForScript(task.getPreScript()));
+            taskData.setPreScript(ScriptData.createForScript(task.getPreScript(), taskData));
```
lpellegr: Please put `task.getX` in a variable in order to factorize  `taskData.setPreScript(ScriptData.createForScript(...)`

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
```diff
@@ -617,7 +643,7 @@ public void setDataspaceSelectors(List<SelectorData> dataspaceSelectors) {
     @Cascade(CascadeType.ALL)
     @OneToOne(fetch = FetchType.LAZY)
     @JoinColumn(name = "PRE_SCRIPT_ID")
-    @ForeignKey(name = "none")
+    @ForeignKey(name = "none") // to be able to remove script without removing task
```
lpellegr: Same as before.

PR #2273, opened 2015-07-27
Improve Scheduling speed (server side)

youribonnaffe:
Manual performance tests showed improvements with those two changes applied when trying to improve the scheduling speed.
Node selection is a linear process and implies going over all the existing nodes, every saved millisecond can make the scheduling loop faster.

During those tests, it was also observed that tweaking some Scheduler settings helped (pa.scheduler.core.starttask.threadnumber and pa.scheduler.policy.nbtaskperloop).

Other experiments consisted in replacing active objects in RMCore with plain Java objects (monitoring, selection manager).

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
@@ -452,13 +452,16 @@ public abstract class SelectionManager {
         }
 
         List<RMNode> filteredList = new ArrayList<>();
-
+        HashSet<Permission> clientPermissions = new HashSet<>();
         for (RMNode node : freeNodes) {
             // checking the permission
             try {
-                client.checkPermission(node.getUserPermission(), client +
-                    " is not authorized to get the node " + node.getNodeURL() + " from " +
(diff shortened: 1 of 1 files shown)


youribonnaffe:
@lpellegr could you please take a look?

lpellegr:
If performance is really critical, maybe logging could be disabled by default for this specific class. 

Regarding the last point, [line 335](https://github.com/ow2-proactive/scheduling/pull/2273/files#diff-6d406cafea56319ce4ef002e726e8a9cR335) should be put in a branch that checks if logging is enabled. 

Besides, as a micro/tiny optimization, calls to `logger.isXEnabled` could be cached in a boolean variable defined at the beginning of the method. 

Finally, in you pull-request comment you say that you have tweaked some scheduler properties and  made some experiments by replacing active objects by plain Java object. However, I don't see these changes in the pull-request. Is it normal?

youribonnaffe:
The last com
The last commit tries to avoid logging if not necessary as suggested. With info level, it does not log that much now so I will keep it enabled for now, we can always tweak the log level if it is a problem.

I did not include the scheduler properties in the commit as it was maybe a bit too specific for my performance test, it seems safer to me to tweak those only if needed. I mentioned them just to leave a trace that those exists :).

PR #2303, opened 2015-08-28
Change handling of URIs in TaskProActiveDataspaces #2300

### LINKED ISSUE #2299: fix TestSmartProxy on jenkins Windows
The test fails because it uses the class TestScheduler to start a scheduler and there is a problem with Windows file name handling

"C:\jenkins\tools\hudson.model.JDK\JDK7u79\jre\bin\java": CreateProcess error=206, The filename or extension is too long
       at java.lang.ProcessBuilder.start(ProcessBuilder.java:1047)
       at functionaltests.utils.TestScheduler.start(TestScheduler.java:147)
       at functionaltests.utils.SchedulerTHelper.startScheduler(SchedulerTHelper.java:159)
       at functionaltests.utils.SchedulerTHelper.getSchedulerAuth(SchedulerTHelper.java:204)
       at functionaltests.TestSmartProxy.init(TestSmartProxy.java:102)

       Caused by:
       java.io.IOException: CreateProcess error=206, The filename or extension is too long
           at java.lang.ProcessImpl.create(Native Method)
           at java.lang.ProcessImpl.<init>(ProcessImpl.java:385)
           at java.lang.ProcessImpl.start(ProcessImpl.java:136)
           at java.lang.ProcessBuilder.start(ProcessBuilder.java:1028)
           ... 4 more

1 test completed, 1 failed

### LINKED ISSUE #2300: Change handling of URIs in TaskProActiveDataspaces
Following the speed branch refactor, the Dataspaces URI returned use the VirtualURIs (internal syntax) instead of RealURI (like ftp://). Besides that, the computing of the scratch space folder is broken on windows (missing drive letter in path).

fviale:
Change handling of URIs in TaskProActiveDataspaces #2300
fix TestSmartProxy on jenkins Windows #2299

fix #2299 Changed the way the java command was started to avoid command path which contains dots.

fix #2300 switched to the use of RealURIs and added a method which converts them to file if possible.

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
@@ -36,17 +36,8 @@
  */
 package org.ow2.proactive.scheduler.task;
 
-import java.io.File;
-import java.io.Serializable;
-import java.security.KeyPair;
-import java.security.KeyPairGenerator;
-import java.security.NoSuchAlgorithmException;
-import java.security.PublicKey;
-import java.security.SecureRandom;
-import java.util.Collections;
-import java.util.Map;
-import java.util.concurrent.TimeUnit;
-
+import com.google.common.base.Stopwatch;
+import org.apache.log4j.Logger;
 import org.objectweb.proactive.Body;
 import org.objectweb.proactive.InitActive;
 import org.objectweb.proactive.annotation.ImmediateService;
@@ -71,8 +62,13 @@ import org.ow2.proactive.scheduler.task.executors.TaskExecutor;
 import org.ow2.proactive.scheduler.task.utils.Decrypter;
 import org.ow2.proactive.scheduler.task.utils.TaskKiller;
 import org.ow2.proactive.scheduler.task.utils.WallTimer;
-import com.google.common.base.Stopwatch;
-import org.apache.log4j.Logger;
+
+import java.io.File;
+import java.io.Serializable;
+import java.security.*;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 
 /**
@@ -128,7 +124,7 @@ public class TaskLauncher implements InitActive {
     }
 
     public void doTask(ExecutableContainer executableContainer, TaskResult[] previousTasksResults,
-            TaskTerminateNotification terminateNotification) {
+                       TaskTerminateNotification terminateNotification) {
         logger.info("Task started");
 
         WallTimer wallTimer = new WallTimer(initializer.getWalltime(), taskKiller);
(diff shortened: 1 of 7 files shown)


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
```diff
@@ -148,12 +144,18 @@ public void doTask(ExecutableContainer executableContainer, TaskResult[] previou
...
-                dataspaces.getScratchURI(), dataspaces.getInputURI(), dataspaces.getOutputURI(),
-                dataspaces.getUserURI(), dataspaces.getGlobalURI(), progressFileReader.getProgressFile()
-                        .toString(), getHostname());
+                    dataspaces.getScratchURI(), dataspaces.getInputURI(), dataspaces.getOutputURI(),
+                    dataspaces.getUserURI(), dataspaces.getGlobalURI(), progressFileReader.getProgressFile()
+                    .toString(), getHostname());
 
             File workingDir = getTaskWorkingDir(context, dataspaces);
 
+            logger.info("Task working dir : " + workingDir);
```
lpellegr: Have you added these information to improve debugging? if yes, I suggest to use debug logging level (also I think there is no space before `:` in English). 

Besides, it could be interesting to wrap all logger calls in a condition which checks if logging is enabled for the specified level (thus avoiding a few calls and String concatenations when not required).

marcocast: Agree with Laurent, no need to put INFO level to display debugging information.
Regarding wrapping logs with isDebugEnabled method, I would suggest to simply upgrade to last version of log4j 2 http://logging.apache.org/log4j/2.x/maven-artifacts.html or to use slf4j to completely avoid the need to check if Debug level is enabled and keep the code clean. 
On the cost or benefit of checking debug levels see https://logging.apache.org/log4j/2.x/performance.html .

lpellegr: @marcocast Using SLF4J and its associated logback implementation is something that is already planned (probably with the upgrade dependencies). However, it is not possible to make such big updates each time we saw it, especially if _stories_ must take less than 3 days. 

You might be interested by the next link, which gives arguments to prefer logback over log4j:

http://logback.qos.ch/reasonsToSwitch.html

lpellegr:
@fviale I have taken a look as you asked. Only a few comments, otherwise it's okay for me.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
marcocast: Yes, logback + slf4j will do the job when we will upgrade the libs. We have only 400 debug logging statements in the code, should take way less than 3 days to change. 
In the meantime I would say it's fine to leave DEBUG level logging without isDebugEnabled unless it is done to avoid really expensive operations.

fviale: The log level in the scheduler-node execution is a bit confusing. Those logs are never set to debug. They might be set this way if ever we really want to debug the indepth functioning of the TaskLaunchers. The info level is related to all information they are useful when we receive logs from a client installation. In that case, knowing what are the location of dataspaces, which file transfers occurred etc is mandatory to understand any problem which might have happened several days ago.

lpellegr: Regarding the time required, have you checked there is no specific implementation of log4j interfaces or related features that would require more work (e.g. custom appender implementations, etc.)? I think it's not just a matter of adapting logging statements but I am maybe wrong...

Besides, IMHO, the upgrade should be done for scheduling and associated projects so that it has a global benefit along with JAR files embedded in a release package.

marcocast: @lpellegr  Good point, in the case we see it can take more than 3 days we can split the stories.
@fviale looks like those logs are needed at INFO level so :)

lpellegr:
Test this please.

lpellegr:
Retest this please.

lpellegr:
Retest this please.

fviale:
Retest this please

lpellegr:
Retest this please

marcocast:
Retest this please

marcocast:
Retest this please

marcocast:
Retest this please

fviale:
Retest this please

fviale:
Retest this please

fviale:
And again, Retest this please

fviale:
Retest this please

marcocast:
Test this please

marcocast:
Test this please

PR #2320, opened 2015-09-25
Add new variable PA_USER

### LINKED ISSUE #2280: Add new variable PA_USER
It is currently not possible to get the username of the user that submitted the task from a task itself.

The only workaround is to run a task as runAsMe. Then, since the process will run under the same username as the one used for authentication in the scheduler, the information can be retrieved with the following piece of code:

```java
System.getProperty("user.name")
```

lpellegr:
Resolves #2280


lpellegr:

fviale:
ok for me

PR #2326, opened 2015-10-01
fix1997: Making "dependences" transient in ClientTaskState.

tobwiens:
This will have a big and unforeseeable effect on the execution of the scheduler server. I submitted this pull request to check if all test run and to collect comments about what can be affected by this change.

So what I have found is that the ClientTaskState is serialized inside the ProActive scheduler:
- ClientTaskState has an ```ArrayList<TaskState>``` -> Custom arbitrary tree
- Java serializes custom arbitrary trees recursively. 
- Around 600 recursive method calls will throw a ```StackOverflowError``` exception

What are the solutions?
- Write own serialization code
- Make it versatile 
- Change internal data-structures Task, TaskState, .....

*Write own serialization code*
I investigated how to manually serialize an arbitrary tree.
- It is possible by flattening the class through custom code -> That has a huge disadvantage because the track of objects gets lost.
That means that an ArrayList, which references two nodes in the tree, it will not reference the same objects after serialization. Because objects will be deeply copied and not re-used, even if the initial references were to the same object.

-> So if a ClientJobState object is serialized (it holds a list of all tasks), the amount deserialized objects (with n tasks) will be n(n+1)/2. That means we will have n*n/2 more traffic/memory utilization. – I think that is too much increase

*Make it versatile*
Just changing the serialization behavior in an environment where nearly everything depends on serialization does not seem very safe?!

*Change internal data-structure* 
We can change the data structure but it is a lot of work.

### DIFF
--- a/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/ClientJobState.java
+++ b/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/ClientJobState.java
@@ -1,21 +1,19 @@
 package org.ow2.proactive.scheduler.job;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
 import org.ow2.proactive.scheduler.common.Scheduler;
 import org.ow2.proactive.scheduler.common.job.JobInfo;
 import org.ow2.proactive.scheduler.common.job.JobState;
 import org.ow2.proactive.scheduler.common.job.JobType;
-import org.ow2.proactive.scheduler.common.task.RestartMode;
-import org.ow2.proactive.scheduler.common.task.TaskId;
-import org.ow2.proactive.scheduler.common.task.TaskInfo;
-import org.ow2.proactive.scheduler.common.task.TaskState;
-import org.ow2.proactive.scheduler.common.task.TaskStatus;
+import org.ow2.proactive.scheduler.common.task.*;
 import org.ow2.proactive.scheduler.task.ClientTaskState;
 import org.ow2.proactive.scheduler.task.TaskInfoImpl;
+
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 
 /**
@@ -169,6 +167,15 @@ public final class ClientJobState extends JobState {
     public RestartMode getRestartTaskOnError() {
         throw new RuntimeException(
             "Not implemented: the restart task on error property is not available on client side.");
+    }
+
+    private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IOException {
+        ois.defaultReadObject();
+        for(TaskState task : this.getTasks()) {
+            if (task instanceof ClientTaskState) {
+                ((ClientTaskState)task).restoreDependences(this.tasks);
+            }
+        }
     }
 
 }
--- a/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/ClientJobState.java
+++ b/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/ClientJobState.java
@@ -29,14 +29,13 @@ import java.util.Map;
  */
 public final class ClientJobState extends JobState {
 
+    private final ClientJobSerializationHelper clientJobSerializationHelper;
     private JobInfoImpl jobInfo;
     private String owner;
     private JobType type;
     private Map<TaskId, TaskState> tasks = new HashMap<>();
-
     private boolean cancelJobOnError;
     private int maxNumberOfExecution;
-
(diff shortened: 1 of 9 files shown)


marcocast:
The first thing we need to do now, is to understand the Why. Why do we serialise the all the task states? who is deserialising them? Do we need to serialise all the tasks data? 
Answering correctly to all of those questions will bring the solution to light.

tobwiens:
Who is serializing the ClientJobState and ClientTaskState?:
The ```SchedulerFrontend implements InitActive, Scheduler, RunActive``` has a method which is ```JobState getJobState```. It accesses the frontendState member and returns the JobState.

Why do we access the JobState?:
The JobState is requested by any type of frontend through the ```SchedulerProxyUserInterface implements Scheduler, Serializable```. 
The ```SchedulerProxyUserInterface ``` has a member ```protected Scheduler uischeduler;``` on which the getJobState(String) is executed. That will forward (through rmi) the method call to the ```SchedulerFrontend```, which returns the ClientJobState object. That will serialize the ClientJobState object which holds all ClientTaskState objects.

Do we need to serialize all task data?:
No. That will change the REST API.

tobwiens:
I tried to escalate the issue, by creating a xml workflow with 500, 1000, 1500, 2000, 2500 concatenated tasks.

The submission works up to 2000 tasks and fails for 2500 with a error code 413 (HTTP errror: Entity too large). 

So.... the serialization issue only exists in the frontend and with loops which blow the data-structures up. :+1:

fviale:
Are you sure the JobState is involved when you fail to submit ? Submission
only returns a jobid (and not a jobstate).

I would rather think the "Entity too large" error is due to jetty limiting
incoming http transfer.

tobwiens:
retest this please.

tobwiens:
retest this please.

tobwiens:
retest this please.

tobwiens:
When I fail to submit the jetty server is involved and the the InternalJob. I tested if one can push a too big data-structure into the scheduler and make it crash. But it was not possible to submit large workflows. Therefore the issue probably won't appear during submission. Furthermore the InternalJob is never serialized and the InternalTask has its dependencies transient.

tl;dr:
It only appears with ClientJobState and ClientTaskState. Everything else seems fine.

tobwiens:
retest this please.

tobwiens:
retest this please.

tobwiens:
retest this please.

tobwiens:
retest this please.

tobwiens:
retest this please.

tobwiens:
retest this please.

lpellegr:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
retest this please

tobwiens:
retest this please.

tobwiens:
retest this please.

tobwiens:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
Retest!

tobwiens:
Retest this please!

tobwiens:
Come on! Retest this please!

tobwiens:
Retest this please!

tobwiens:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
This is fixed in #2352

scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/ClientJobState.java
```diff
@@ -1,22 +1,20 @@
...
 import org.ow2.proactive.scheduler.common.Scheduler;
 import org.ow2.proactive.scheduler.common.job.JobInfo;
 import org.ow2.proactive.scheduler.common.job.JobState;
 import org.ow2.proactive.scheduler.common.job.JobType;
-import org.ow2.proactive.scheduler.common.task.RestartMode;
-import org.ow2.proactive.scheduler.common.task.TaskId;
-import org.ow2.proactive.scheduler.common.task.TaskInfo;
-import org.ow2.proactive.scheduler.common.task.TaskState;
-import org.ow2.proactive.scheduler.common.task.TaskStatus;
+import org.ow2.proactive.scheduler.common.task.*;
```
marcocast: normally it is better to import the specific classes needed instead of *

marcocast:
:+1:

scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/ClientJobState.java
tobwiens: Thank you. I will fix that!

PR #2329, opened 2015-10-02
Job submission with variables from CLI does not work with Java 8

### LINKED ISSUE #2252: Job submission with variables from CLI does not work with Java 8
If a submission of a job, with variables, is performed from interactive CLI, using Java 8, it will fail with the following error:
> $> submit('job.xml', [foo='bar'])
> Cannot cast jdk.internal.dynalink.beans.StaticClass to java.lang.Class

Probably an issue with Nashorn that is now used with Java 8.

cdiop:
Modification of submit command syntax
```submit('pathWorkflow.xml', "var1=value1 var2=value2")```

This pull-request closes #2252.


fviale:
what if we need that var1=value with space   ?

fviale:
I mean var1 = "value with space"

cdiop:
I keep the same syntax 
-s,--submit <job-descriptor [var1=value1 var2=value2 …]>

and I remove the [ ]

We can replace the space by , or ;

fviale:
Ideally in non interactive mode, I would see something like :

proactive-client -s ... var1="value1" var2="value2"

But I don't know how argument are parsed (and especially across different
OS shells) and how that would translate in interactive mode.

cdiop:
The syntax is now:
submit('job.xml', '{"var1":"value1","var2":"value2"}') with possibiulity to have space

fviale:
I think it's a bit "heavy"

why switching from "=" to ":" ? I would prefer to choose the syntax
collectively, other people might have elegant ideas

fviale:
I suppose it's kind of tricky as the argument is parsed as a whole and
split into tokens

fviale:
ah that's a JSON syntax maybe ? Then in that case, yes it makes sense.

marcocast:
Yes, it is pure json format

cdiop:
retest this please

fviale:
How will be the syntax from the CLI in non-interactive mode ?

cdiop:
Could two people please review the changes?

cdiop:
retest this please

cdiop:
Retest this please.

cdiop:
Retest this please

lpellegr:
@cdiop I have added some suggestions.

cdiop:
Retest this please

cdiop:
Retest this please.

cdiop:
Retest this please.

cdiop:
Retest this please

lpellegr:
Retest this please

lpellegr:
Retest this please

PR #2337, opened 2015-10-08
Grouping slow tests

marcocast:
Moved slow running tests (over 1 minute per class test) into regression folder. 
Those slow tests will not run by default when using ./gradlew functionalTest.
They will be triggered only from the scheduling jenkins job with ./gradlew functionalTest -Dregression=true


marcocast:
Please 2 reviewer here. It is not really a code change, but only a small structure + gralde test execution change

lpellegr:
Is it not possible to exclude tests based on their fully qualified name instead of moving class files? I ask because these will need to be moved again later...

marcocast:
It is possible, but I prefer to group them in a specific folder (funtionaltest.regression) as it leaves the gradle file easier to read and modify. Then Later we will decide what's best.

build.gradle
```diff
@@ -155,6 +155,12 @@ ext.encoding = {
 ext.schedulingFunctionalTestConfiguration = {
 
     include 'functionaltests/**'
+    
+    boolean isRegression = System.getProperty("regression") != null ? Boolean.valueOf(System.getProperty("regression")) : false;
```
lpellegr: Why not using Gradle project properties which seems done for this purpose (instead of `System.getProperty(...)`)? 

```gradle
project.hasProperty('regression')
```

then invoke with:

`-Pregression` ?

marcocast: :+1:

marcocast:
retest this please

marcocast:
Test please

PR #2349, opened 2015-10-14
fixTestsWindows

tobwiens:
This PR updates a Workflow which is executed in the TestPropagatedVariablesWalltime test. The worklow runs a native task with a sleep 30 command with a walltime of 5 seconds. It tests if the variables are propagated even though the walltime killed the task. 

That is problematic since it fails on Windows with: Cannot run program "sleep": CreateProcess error=2, The system cannot find the file specified. 

The Workflow is updated to use a python task with time.sleep(30) instead of a native task.

I am wondering why this test could pass on the Windows machine sometimes? It parses the task output for var_value, which is an actual value of 'var'. That might have been randomly matched, but I don't see how at the moment. Does anyone have an idea?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -159,7 +159,7 @@ ext.schedulingFunctionalTestConfiguration = {
     
     if(!project.hasProperty('regression')){
 	    def regression_tests = new File(project.rootDir, "/regression_tests.txt")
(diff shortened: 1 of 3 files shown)


tobwiens:
Retest this please.

tobwiens:
This is close in favor for: #2353

marcocast:
I think you committed .travis.yml by mistake?

PR #2369, opened 2015-10-28
Added necessary accessors in  CookieBasedPTK

fviale:


lpellegr:
@fviale If it is okay that tests fail, then :+1:

PR #2382, opened 2015-11-16
Fix dates parameters in task-centric REST resources

paraita:
The parameters `from` and `to` could not be used independently from each other. This has been fixed.
Functional tests are also added to cover the new SchedulerDBManager methods introduced by those REST resources.


tobwiens:
Looks okay for me. What does sonar have to say? Is sonar blocked?

paraita:
Everything's fine except for those scriptengine errors (on the jenkins slave side).

Issue #2391, opened 2015-11-29
In scheduling 6.3, script dataspace bindings should use LocalSpace / RemoteSpace API


fviale:
Currently script dataspace bindings use the DataspaceFileObject API, which is complicated to use and understand, from a customer point of view. It does not allow to easily manipulate files by patterns, etc.

LocalSpace/RemoteSpace api should be used instead in script bindings

fviale:
fixed in version 6.3.3

PR #2392, opened 2015-11-30
Script dataspace bindings should use LocalSpace / RemoteSpace API #2391

fviale:
- replaced the current DataSpaceFileObject bindings with the LocalSpace / RemoteSpace adapters. This allow a cleaner and easier control of file transfers in scripts.
- had to move Credentials API from scheduler-client to scheduler-api
- removed NodeSet bindings in scripts. Node objects will have to be retrieved using proactive api and the nodes urls.


lpellegr:
:+1:

PR #2393, opened 2015-11-30
Make semantic consistent for range parameter used with pagination 

lpellegr:


lpellegr:

tobwiens:
Looks good, you made the code more readable :+1:!

scheduler/scheduler-api/src/test/java/org/ow2/proactive/scheduler/common/util/PaginationTest.java
```diff
@@ -0,0 +1,59 @@
+package org.ow2.proactive.scheduler.common.util;
+
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assertThat;
```
marcocast: Why Truth instead of Hamcrest here?

marcocast:
:+1:

scheduler/scheduler-api/src/test/java/org/ow2/proactive/scheduler/common/util/PaginationTest.java
lpellegr: @marcocast Truth is at least as readable as Hamcrest, sometimes more, but in all cases more simple to use. Since new tools and library have been introducing by some people in their developments without any consensus, I do the same when I think it can ease my work.

PR #2401, opened 2015-12-04
Update tasks filtering in DB

lpellegr:


lpellegr:

paraita:
:+1:

lpellegr:
Thank you for the reviews.

@mbenguig Below are some answers:

1. Setter functions return the _Builder_ instance in order to have the possibility to chain method calls, thus providing a fluent interface. This last is quite common with builder pattern.
2. Suggestions have been applied.
3. The value _86400000_ depicts the number of milliseconds in a day. This last is multiplied to the _shift_ parameter of method _getTimeRelativeToReference_. As said in the comment above the method, the _shift_ is a number of days. Since the reference timestamp is in milliseconds, it explains why this value is used.

PR #2404, opened 2015-12-08
Remove definition of web app specific properties from JettyStarter

lpellegr:
The properties which have been removed are Java system properties that have precedence over properties defined in configuration file. Consequently, it was impossible to change the value of these properties without editing code and recompiling.

Since the properties which have been removed are Web apps specific, they should be defined by them only. Although Web apps are deployed by default on the same host where the scheduler is, it is not always the case in real world scenarios.


lpellegr:

marcocast:
ok to me

PR #2415, opened 2015-12-16
changing the tests methods called run

marcocast:
The test methods called run are badly handled in Jenkins causing tests outputs to be unreadeble


marcocast:
2 reviews please? (it's just renaming some tests methods)

paraita:
LGFM

lpellegr:
:+1: Looks fine to me

I think it would be good that we define (or use and improve existing one) IDE profiles in the future to avoid package declarations reordering and expansion.

PR #2426, opened 2016-01-04
Add installation scripts in tools for redhat-based linux os

fviale:


lpellegr:
@fviale This script seems to require root access. If true, it would be nice to check first that the script is executed as root (i.e. `$(id -u) == "0"`). Ideally, checking sudo would even be better.

tools/install.sh
```diff
@@ -0,0 +1,80 @@
...
+PA_FOLDER_NAME="$(basename "$CURRENT_PADIR")"
+
+/usr/bin/cp -f -R $CURRENT_PADIR $PA_ROOT
+
+ln -s -f $PA_ROOT/$PA_FOLDER_NAME $PA_ROOT/default
+
+# creation of the proactive user
+
+read -e -p "Name of the user starting the ProActive service: " -i "proactive" USER
+
```
lpellegr: Shouldn’t we ask for default group associated to the user that is created? I think that by default it would be nice to use the same group and user as the one used for proactive-agent distribution .

tools/install.sh
```diff
@@ -0,0 +1,80 @@
...
+echo "Do you want to create this user ?"
+select yn in "Yes" "No"; do
+    case $yn in
+        Yes ) 
+		useradd -m $USER;
```
lpellegr: Shouldn’t home directory be set to `$PA_ROOT`?

tools/install.sh
```diff
@@ -0,0 +1,80 @@
...
+    esac
+done
+
+chown -R $USER:$USER $PA_ROOT/$CURRENT_PA_DIR
+
+# installation of the service script
+
+cp $SCRIPT_DIR/proactive-scheduler /etc/init.d/
+
+read -e -p "Protocol used by the proactive server: " -i "http" PROTOCOL
```
lpellegr: I think it could be helpful to specify the options which are allowed between square brackets.

tools/install.sh
```diff
@@ -0,0 +1,80 @@
...
+
+/usr/bin/cp -f -R $CURRENT_PADIR $PA_ROOT
+
+ln -s -f $PA_ROOT/$PA_FOLDER_NAME $PA_ROOT/default
+
+# creation of the proactive user
+
+read -e -p "Name of the user starting the ProActive service: " -i "proactive" USER
+
+echo "Do you want to create this user ?"
```
lpellegr: No space before question mark.

tools/install.sh
```diff
@@ -0,0 +1,80 @@
...
+
+
+chmod 700 /etc/init.d/proactive-scheduler
+
+mkdir -p /var/log/proactive
+touch /var/log/proactive/scheduler
+
+chown -R $USER:$USER /var/log/proactive
+
+echo "Start the proactive-scheduler service at startup ?"
```
lpellegr: No space before question mark.

tools/proactive-scheduler
```diff
@@ -0,0 +1,187 @@
...
+	sleep 3
+}
+
+
+
+start() {
+	CMD="nohup "
+	if [ $(daemonRunning) == 1 ]
+	then
+			find $PROACTIVE_HOME -mtime +50 -name '*.log' -exec rm {} \;
```
lpellegr: I think that logs removal value should be configurable (i.e. value defined as a global variable)? why 50 days by default?

tools/proactive-scheduler
```diff
@@ -0,0 +1,187 @@
...
+}
+
+waitForCompletion() {
+
+	echo "Service $NAME is being stopped"
+	while [ $(daemonRunning) == 0 ]
+	do
+		sleep 1
+	done
+	sleep 3
```
lpellegr: Is this sleep really required?

tools/proactive-scheduler
```diff
@@ -0,0 +1,187 @@
...
+			#echo $CMD
+			# CMD="$CMD echo \$!"
+			if [ "$USER" != "" ]
+			then
+				$(su $USER -c "$CMD")
+			else
+				$(su -c "cd $PROACTIVE_HOME/bin && $CMD")
+			fi
+			echo Scheduler command executed, waiting for start
+			export https_proxy=
```
lpellegr: Should not _https_proxy_ configuration be a case by case system administrator configuration based on system specificities?

tools/proactive-scheduler
```diff
@@ -0,0 +1,187 @@
...
+			sleep 1
+			CURRPID=$(ps aux|grep SchedulerStarter|grep -v "grep"|awk '{ print $2 }')
```
lpellegr: Is it not possible to use _$!_ variable to get the PID of the last process started? by adding for instance `&& echo $!` on line _101_ and _103_ and retrieving the value in a variable, thus avoiding the sleep.

tools/proactive-scheduler
```diff
@@ -0,0 +1,187 @@
...
+			echo $CURRPID > $PIDFILE
+			until curl --noproxy '*' --insecure $REST_PUBLIC_URL 2>/dev/null;
+        		do
+                		sleep 0.5
+        		done
+			echo Web interfaces started at $PROTOCOL://${ALIAS}:$PORT
```
lpellegr: There is no guarantee that all Web interfaces are started as suggested by the message. Only Web app related to REST API has the guarantee to be up.

tools/proactive-scheduler
```diff
@@ -0,0 +1,187 @@
...
+
+	else
+		echo -e "Service $NAME already running, pid: \e[1;34m$(cat $PIDFILE)\e[0m"
+	fi
+
+}
+
+
+
+# Restart the service FOO
```
lpellegr: service FOO?

tools/proactive-scheduler
```diff
@@ -0,0 +1,187 @@
...
+stop() {
+	if [ -f $PIDFILE ] 
+	then
+		CURRPID=$(cat $PIDFILE)
+	fi
+
+	if [ $(daemonRunning) == 0 ]
+	then
+		# kill cheduler process
```
lpellegr: missing _s_ to scheduler

tools/install.sh
fviale: ProActive agent uses proactive/proactive as default username and group:
AGENTUSER="proactive"
AGENTGROUP="proactive"

I could eventually ask the user if he wants a custom group

tools/install.sh
fviale: What would that bring ?

tools/install.sh
fviale: makes sense

tools/proactive-scheduler
fviale: That's true, it should be asked in the install script.

tools/proactive-scheduler
fviale: The last sleep 3 is not relevant, you are right

tools/install.sh
lpellegr: If a user doesn't know where installation directory is but knows under which user the process is executing (e.g. `ps aux`) he can easily identify the installation directory of proactive service.

tools/proactive-scheduler
fviale: proxy is used for outgoing http connections. Here we are connecting on the local machine and it must be disabled in order to allow the http connection.

tools/proactive-scheduler
fviale: It does not work, I tried. The PID of the process is the one of the shell process executing the bash script. We need the PID of the JVM subprocess to be able to kill the scheduler properly.

tools/proactive-scheduler
fviale: That's correct. Do you want me to test against the scheduler/rm/studio ? That's a lot of tests, is it really necessary. There didn't seem to be any issue related to that following my experience with this script.

tools/proactive-scheduler
fviale: service proactive-scheduler, why FOO ?

lpellegr: yes, it was a question :) make more sense to put `proactive-scheduler`.

fviale: NAME is set to "proactive-scheduler" :-)

lpellegr: but not for the text used as comment.

fviale: Ahhhhhhhh :-)

fviale:
Superseded by another pull request

PR #2429, opened 2016-01-08
fix for #2426

fviale:
- The install script asks various configuration questions to the user
- It copies the scheduling distribution to a specified folder
- It eventually creates the user account starting the scheduler
- It installs the service.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+
+if [[ $(id -u) -ne 0 ]] ; then echo "Please run the installation script as root" ; exit 1 ; fi
+
+SCRIPT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
+
+# installation of the ProActive distribution
+ 
+read -e -p "Directory where to install the scheduler: " -i "/opt/proactive" PA_ROOT
+
+PA_ROOT=$(echo $PA_ROOT | xargs)
+
+mkdir -p $PA_ROOT
+rm -f $PA_ROOT/default
+
+CURRENT_PADIR="$(dirname "$SCRIPT_DIR")"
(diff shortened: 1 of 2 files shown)


lpellegr:
I have tried the script with success on Fedora 23. 

However, I noticed that _install.sh_ script is currently installing the distribution from the sources. It means that a user who  clones the scheduling repository and uses the install script will install a ProActive version that does not deploy Scheduler and RM web portals. Asking for the distribution to install (i.e. a path to the ZIP file to use or support for remote downloading of a distribution) would be more practical and help us to upgrade an existing installation.

tools/install.sh
```diff
@@ -0,0 +1,80 @@
...
+    esac
+done
+
+chown -R $USER:$USER $PA_ROOT/$CURRENT_PA_DIR
+
+# installation of the service script
+
+cp $SCRIPT_DIR/proactive-scheduler /etc/init.d/
+
+read -e -p "Protocol used by the proactive server: " -i "http" PROTOCOL
```
lpellegr: You said that it makes sense to specify the options which are allowed between square brackets but you don't did it. Did you forget?

tools/install.sh
```diff
@@ -0,0 +1,80 @@
...
+    case $yn in
+        Yes ) 
+		useradd -m $USER;
+		passwd $USER; 
+		break;;
+        No ) break;;
+    esac
+done
+
+chown -R $USER:$USER $PA_ROOT/$CURRENT_PA_DIR
```
lpellegr: Group to use is not asked by installation script as we discussed.

fviale: There is a version problem. This code didn't include my changes.

lpellegr:
@fviale Available protocols are not displayed with `install.sh`. Otherwise it looks okay for me.

PR #2434, opened 2016-01-19
Add new version of docker compose script engine to scheudling project

tobwiens:
Add new version of docker compose script engine to scheudling project


marcocast:
:+1:

Issue #2425, opened 2015-12-22
Provide scripts inside the "tools" folders to install the scheduler as a service (linux only)


fviale:
It can be:

1) on linux, a bash script install.sh which copies a generic "proactive-scheduler" startup script to /etc/init.d and then use chkconfig to install it on startup (with some adaptations for different distributions redhat, debian) 
2) on windows, a bat script calling the command SC.exe could be interesting, but this would be for later
https://support.microsoft.com/en-us/kb/251192

This is a very helpful feature when we install the scheduler in production

lpellegr:
+:100: 

For Linux systems it would be nice to have a systemd script since systemv scripts tend to be deprecated.

It seems to be a duplicate of the existing issue #2015.

paraita:
upstart script too ? (Ubuntu)

lpellegr:
@paraita Good idea, although systemd is now supported since 15.04 but not yet in LTS version apparently:

https://wiki.ubuntu.com/SystemdForUpstartUsers

fviale:
install.sh and uninstall.sh scripts added in the tools folder.

PR #2440, opened 2016-01-21
improvement #2425 : checking for java installation and using rsync instead of cp

fviale:


lpellegr:
It would be nice to explain what is the purpose of your patch for commit entitled with "#2425".

Otherwise the changes related to `install.sh` script look okay to me.

PR #2447, opened 2016-01-26
fix for #2443 : added read/write lock in TransactionHelper

fviale:
- added readwrite lock in TransactionHelper to prevent concurrency problems
  - removed synchronization in the SchedulerDBManager


lpellegr:
@fviale It would be nice to:

  - add a small comment in `SchedulerDBManager` that states that synchronization is performed at transaction level through a reentrant read/write lock with `runWithout/With*` methods only. Besides, I think it is important to mention the fact that `runWithoutTransaction` method assumes that read only actions are performed whereas `runWithTransaction` allows read and write.   
  - add some tests, at least one for the original issue the fix was initiated for .

lpellegr:
:+1:

Issue #2448, opened 2016-01-27
When building locally the remote java command SSHInfrastructureV2 search addon jars in the remote addons directory (which does not exist locally)


fviale:
The following error appears in the logs.

The workaround/hack is to create locally the addons folder, using the remote scheduler path.
```
[2016-01-27 18:26:54,211 INFO                o.o.p.r.c.RMCore] Node source HPC-master has been successfully created by "lclaude"
[2016-01-27 18:26:54,210 ERROR .p.r.n.i.InfrastructureManager] Could not acquire node on host hpc1.chuv.ch/155.105.223.15. NS's state refreshed regarding last checked exception: #freeHosts:3 #registered: 0
java.lang.NullPointerException
        at org.ow2.proactive.resourcemanager.utils.CommandLineBuilder.buildCommandLineAsList(CommandLineBuilder.java:299)
        at org.ow2.proactive.resourcemanager.utils.CommandLineBuilder.buildCommandLine(CommandLineBuilder.java:239)
        at org.ow2.proactive.resourcemanager.nodesource.infrastructure.SSHInfrastructureV2.startNodeImpl(SSHInfrastructureV2.java:195)
        at org.ow2.proactive.resourcemanager.nodesource.infrastructure.HostsFileBasedInfrastructureManager$1.run(HostsFileBasedInfrastructureManager.java:121)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

fviale:
commit merged

PR #2452, opened 2016-01-28
fix for #2448 add wildcard of addons dir to the classpath instead of …

fviale:
…searching inside the addons folder for jar files


lpellegr:
:+1:

PR #2453, opened 2016-02-02
work on #2425 : use git to track configuration modifications

fviale:


tools/install.sh
```diff
@@ -5,17 +5,49 @@ if [[ $(id -u) -ne 0 ]] ; then echo "Please run the installation script as root"
 SCRIPT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
 CURRENT_PADIR="$(dirname "$SCRIPT_DIR")"
 PA_FOLDER_NAME="$(basename "$CURRENT_PADIR")"
+OLD_PADIR=
+OS=
+if which yum > /dev/null 2>&1; then
+   OS="RedHat"
```
lpellegr: I would love to have support for _dnf_ which is the successor of _yum_ (yum is now deprecated on Fedora at least).

tools/install.sh
```diff
@@ -5,17 +5,49 @@ if [[ $(id -u) -ne 0 ]] ; then echo "Please run the installation script as root"
...
+	if ! which java > /dev/null 2>&1; then
                 if [ ! -f "$CURRENT_PADIR/jre/bin/java" ]; then
                         echo "JAVA_HOME is not set, no 'java' command could be found in your PATH, and this ProActive distribution does not contain an embedded jre, please install a JRE or JDK." ; exit 1 ;
                 fi
         fi
 fi
 
+if ! which git > /dev/null 2>&1; then
+     echo "GIT is not installed on this computer and is required by the ProActive installation."
+     if confirm "Do you want to install it ? [Y/n] " ; then
```
lpellegr: No space before `?`.

tools/install.sh
```diff
@@ -5,17 +5,49 @@ if [[ $(id -u) -ne 0 ]] ; then echo "Please run the installation script as root"
 SCRIPT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
 CURRENT_PADIR="$(dirname "$SCRIPT_DIR")"
 PA_FOLDER_NAME="$(basename "$CURRENT_PADIR")"
+OLD_PADIR=
+OS=
+if which yum > /dev/null 2>&1; then
+   OS="RedHat"
+elif which apt-get > /dev/null 2>&1; then
+   OS="Debian"
+fi
```
lpellegr: I think that a _else_ block should be added to stop the execution of the script (`exit`) but also to inform the user that its OS is not supported by this installation script.

tools/install.sh
```diff
@@ -125,12 +154,57 @@ touch /var/log/proactive/scheduler
...
 if confirm "Start the proactive-scheduler service at startup? [Y/n] " ; then
-    if chkconfig proactive-scheduler on 2> /dev/null ; then
-        echo "proactive-scheduler service added as a startup daemon"
+    if [[ "$OS" == "RedHat" ]]; then
+       chkconfig proactive-scheduler on
+    elif [[ "$OS" == "Debian" ]]; then
+       update-rc.d proactive-scheduler defaults
     else
-        update-rc.d proactive-scheduler defaults
+        echo "Installation as a startup script not supported on this operating system."
```
lpellegr: As suggested before, this check should be performed much earlier.

tools/install.sh
```diff
@@ -125,12 +154,57 @@ touch /var/log/proactive/scheduler
...
     fi
 fi
 
+if which git > /dev/null 2>&1; then
+    # Porting of older version configuration and data files
+
+    cd $PA_ROOT/default/config && git init
+    cd $PA_ROOT/default/config && git config user.email "support@activeeon.com"
```
lpellegr: You could move to `$PA_ROOT/default/config` only once. The next repetitions look useless.

tools/install.sh
```diff
@@ -125,12 +154,57 @@ touch /var/log/proactive/scheduler
...
+    if [[ "$OLD_PADIR" != "" ]]; then
+
+        echo "Detected an existing ProActive Scheduler installation at $OLD_PADIR, porting configuration into new installation."
+
+        rsync --info=progress2 -a $OLD_PADIR/{addons,data} $PA_ROOT/default/
+
+        OLD_PADIR_NAME="$(basename "$OLD_PADIR")"
+
+        # Commit uncommitted changes in the old repository
+        cd $OLD_PADIR/config && git add -A .
```
lpellegr: Same about `cd $OLD_PADIR/config &&` which is repeated.

lpellegr:
:+1:  Commits with same title could be squashed...

lpellegr:
:+1:  Commits with same title could be squashed...

PR #2455, opened 2016-02-05
Fix for start script generation to use the existing script templates

marcocast:


build.gradle
```diff
@@ -237,18 +237,11 @@ task cleanScripts
 clean.dependsOn cleanScripts
 
 class CustomCreateStartScripts extends CreateStartScripts {
+
+	//We override the generate() method because Gradle adds "bin/" in the classpath by default.
```
fviale: "lib/"

PR #2462, opened 2016-02-11
work on #2425 : added an uninstall.sh script and network configuration in install.sh

fviale:
- The uninstall script removes all versions of the scheduler installed.
- install.sh now display the network interfaces information and let the user select one.
- improved the install.sh script to give a more next/next/next feel

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -3,8 +3,8 @@
 if [[ $(id -u) -ne 0 ]] ; then echo "Please run the installation script as root" ; exit 1 ; fi
 
 SCRIPT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
-CURRENT_PADIR="$(dirname "$SCRIPT_DIR")"
-PA_FOLDER_NAME="$(basename "$CURRENT_PADIR")"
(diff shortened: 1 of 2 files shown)


tools/install.sh
```diff
@@ -122,15 +128,24 @@ read -e -p "Group of the user starting the ProActive service: " -i "proactive" G
...
     else
-        useradd $USER -d $PA_ROOT -g $GROUP
+        exit 1
     fi
-    passwd $USER;
+
 fi
 
+chown $USER:$GROUP $PA_ROOT
 chown -R $USER:$GROUP $PA_ROOT/$CURRENT_PA_DIR
```
paraita: is `CURRENT_PA_DIR` a typo ? You declare `CURRENT_PADIR` at the beginning of the file.

paraita:
the rest looks fine to me ! :+1:

tools/install.sh
fviale: Well spotted paraita !

Indeed it was a mistake.

paraita:
:+1:

PR #2491, opened 2016-03-15
remove default credentials

ShatalovYaroslav:
remove default credentials


dist/war/getstarted/index.html
```diff
@@ -51,7 +51,6 @@
         <hr>
         <p class="lead">Use the command line client via the <a id="restUrlLink" href=""
                                                                title="ProActive REST API">REST API</a>
-            (default credentials: <b>admin/admin</b>)
         </p>
         <pre>$> ./bin/proactive-client -u <span id="restUrlCli"></span> -l admin</pre>
```
marcocast: Do we leave "admin" here or we replace with $USER ?

PR #2506, opened 2016-03-22
Change studio icon

ShatalovYaroslav:
Change studio icon


marcocast:
:+1:   Are you going to replace all the studio icons with this new one?

PR #2520, opened 2016-04-13
Preserve numberOfExecutionLeft after Scheduler restart

tobwiens:
The numberOfExecutionLeft is laoded at Scheduler startup from the database. But when the database data structure(Task Data) is converted into the internal data structure (InternalTask) the number of executions left are overwritten. 

This PR changes that behavior by providing two more setter methods which allow to set the number of executions left on node failure and task failure. Both setters are the utilized in the conversion process.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
@@ -542,14 +542,20 @@ public abstract class InternalTask extends TaskState {
         this.taskInfo = (TaskInfoImpl) taskInfo;
     }
 
-    /**
-     * @see org.ow2.proactive.scheduler.common.task.CommonAttribute#setMaxNumberOfExecution(int)
-     */
+
     @Override
     public void setMaxNumberOfExecution(int numberOfExecution) {
         super.setMaxNumberOfExecution(numberOfExecution);
         this.taskInfo.setNumberOfExecutionLeft(numberOfExecution);
         this.taskInfo.setNumberOfExecutionOnFailureLeft(maxNumberOfExecutionOnFailure);
+    }
+
+    public void setNumberOfExecutionLeft(int numberOfExecutionLeft) {
+        this.taskInfo.setNumberOfExecutionLeft(numberOfExecutionLeft);
+    }
+
+    public void setNumberOfExecutionOnFailureLeft( int numberOfExecutionOnFailureLeft) {
+        this.taskInfo.setNumberOfExecutionOnFailureLeft(numberOfExecutionOnFailureLeft);
     }
 
     /**
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
@@ -531,6 +531,8 @@ public class TaskData {
         internalTask.setRunAsMe(isRunAsMe());
         internalTask.setWallTime(getWallTime());
         internalTask.setMaxNumberOfExecution(getMaxNumberOfExecution());
+        internalTask.setNumberOfExecutionLeft(getNumberOfExecutionLeft());
+        internalTask.setNumberOfExecutionOnFailureLeft(getNumberOfExecutionOnFailureLeft());
         internalTask.setRestartTaskOnError(getRestartMode());
         internalTask.setFlowBlock(getFlowBlock());
         internalTask.setIterationIndex(getIteration());


tobwiens:
Retest this please.

tobwiens:
@ow2-proactive/developers: Two reviews please.

tobwiens:
Retest this please.

tobwiens:
Retest this please.

tobwiens:
Retest this please.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
```diff
@@ -542,16 +542,22 @@ public synchronized void update(TaskInfo taskInfo) {
...
         super.setMaxNumberOfExecution(numberOfExecution);
         this.taskInfo.setNumberOfExecutionLeft(numberOfExecution);
         this.taskInfo.setNumberOfExecutionOnFailureLeft(maxNumberOfExecutionOnFailure);
     }
 
+    public void setNumberOfExecutionLeft(int numberOfExecutionLeft) {
+        this.taskInfo.setNumberOfExecutionLeft(numberOfExecutionLeft);
+    }
+
+    public void setNumberOfExecutionOnFailureLeft( int numberOfExecutionOnFailureLeft) {
```
lpellegr: Remove space before `int`.

lpellegr:
@fviale Do you remember if there was a particular reason to reset the number of execution left on restart?

lpellegr:
@tobwiens Even if TaskData#createTaskData is not called after a restart of the Scheduler, do you understand why this method resets the number of execution left?

tobwiens:
@lpellegr  The TaskData#createTaskData sets the numberOfExecionsOnFailureLeft https://github.com/ow2-proactive/scheduling/blob/master/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java#L417.

lpellegr:
@tobwiens I mean all TaskData fields are updated based on values available in the _InternalScriptTaskData_, except `setNumberOfExecutionOnFailureLeft` and `setNumberOfExecutionLeft`, which looks really odd to me. 

Precisely, the constant value is used for the number left which is not a contant:

```
taskData.setNumberOfExecutionOnFailureLeft(
                PASchedulerProperties.NUMBER_OF_EXECUTION_ON_FAILURE.getValueAsInt());
```

Fields/getters/setter called `maxNumberOfExecutionLeft` seems not to exist in _TaskData_, are we speaking about the same class?

tobwiens:
@lpellegr: You are right. TaskData has a setter setNumberOfExecutionLeft. I have mistakenly put the max in front of it. 
To clarify: 
There is the maximum number of executions called (maxNumberOfExecution)
https://github.com/tobwiens/scheduling/blob/master/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java#L763

Here is the maximum number of executions on failure called (PASchedulerProperties.NUMBER_OF_EXECUTION_ON_FAILURE.getValueAsInt())
https://github.com/tobwiens/scheduling/blob/master/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java#L414

Here is the maximum number of execution left:
https://github.com/tobwiens/scheduling/blob/master/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java#L128

Here is the max number of execution on failure left:
https://github.com/tobwiens/scheduling/blob/master/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java#L130

This setter did set all execution values.
```
@Override
     public void setMaxNumberOfExecution(int numberOfExecution) {
         super.setMaxNumberOfExecution(numberOfExecution);
         this.taskInfo.setNumberOfExecutionLeft(numberOfExecution);
         this.taskInfo.setNumberOfExecutionOnFailureLeft(maxNumberOfExecutionOnFailure);
     }
```

So at initialization the number setNumberOfExecutionLeft was set to the numberOfExecution = maxNumberOfExecution. That was basically resetting the number at each restart to maximum.
Nevertheless it was no representing the state in the database (I am not sure why). I assume that the next execution will then overwrite the inconsistent database state.

I added
```
public void setNumberOfExecutionLeft(int numberOfExecutionLeft) {
+        this.taskInfo.setNumberOfExecutionLeft(numberOfExecutionLeft);
+    }
```
In order to be able to set the numberOfExecutionLeft consistent with the database state. This setter was not exposed before. 

When you say that 
```
taskData.setNumberOfExecutionOnFailureLeft(
                PASchedulerProperties.NUMBER_OF_EXECUTION_ON_FAILURE.getValueAsInt());
```
is not a constant, you mean that it changes according to the settings? And the settings can be changed during a restart of the Scheduler.


When the Scheduler restarts it calls the [toInternalTask](https://github.com/ow2-proactive/scheduling/pull/2520/commits/5d5b3253fd300ba6ac21b8263527b5222b8de6fa#diff-dcc638954b0bf4c84b495f0517c4d510R503). Because it loads the TaskData object from the database and creates the internalTask from it. So the TaskData#toInternalTask, resets the number of executions. (Inside the runtime but not the database).

When a task is created it calls the TaskData#createTaskData. That will have effect when a new task is scheduled. 

I hope it is more clear.

lpellegr:
@tobwiens Thank you for the clarifications.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
```diff
@@ -542,16 +542,22 @@ public synchronized void update(TaskInfo taskInfo) {
...
-    /**
-     * @see org.ow2.proactive.scheduler.common.task.CommonAttribute#setMaxNumberOfExecution(int)
-     */
+
     @Override
     public void setMaxNumberOfExecution(int numberOfExecution) {
         super.setMaxNumberOfExecution(numberOfExecution);
         this.taskInfo.setNumberOfExecutionLeft(numberOfExecution);
         this.taskInfo.setNumberOfExecutionOnFailureLeft(maxNumberOfExecutionOnFailure);
     }
```
mbenguig: Since this function does more than its title "setMaxNumberOfExecution", why not removing it, and replacing all setMaxNumberOfExecution(..) calls by setMaxNumberOfExecution(..), setNumberOfExecutionLeft(..) and setNumberOfExecutionOnFailureLeft(..)?

PR #2523, opened 2016-04-19
Allow parallel environments with 1 node.

tobwiens:
Allow parallel environments with 1 node.

Tasks which cannot be scheduled in parallel, because e.g. their selection scripts calculate resource allocations (and those change after the task is scheduled, so in the next loop) profit from being part of a parallel environment.
A parallel environment forbids to schedule other tasks at the same time, that means that the scheduling decision including running a task's selection script happen for one task at a time. Running the selection scripts for only one task lowers the likelyhood to allocate more resources than a host actually has due to race conditions.

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/task/ParallelEnvironment.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/task/ParallelEnvironment.java
@@ -75,7 +75,7 @@ public class ParallelEnvironment implements Serializable {
      * @param nodesNumber the number of nodes required by task.
      */
     public ParallelEnvironment(int nodesNumber) {
-        if (nodesNumber <= 1) {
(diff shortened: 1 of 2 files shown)


tobwiens:
Fixes issue #1747

fviale:
+1

marcocast:
no tests?

fviale:
I thought so as well, but topology selection with value 1 is already tested in the rm testsuite:
https://github.com/ow2-proactive/scheduling/blob/83db99300753117b3b97d8e46e1df32abbb0a800/rm/rm-server/src/test/java/functionaltests/topology/LocalSelectionTest.java

tobwiens:
@marcocast @fviale I can add some more tests, the current ones don't seem to cover much, otherwise those had detected a change in the code.

tobwiens:
Fixed: 2518

tobwiens:
The parallel environment is already tested. No tests are added.

Issue #2527, opened 2016-04-25
When accessing files in user/global space from the rest api, a desynchronization can occur 


fviale:
This issue appeared randomly in long sequence of jobs using dataspaces.

At the job termination, the file was requested from the REST API but could not be found, even though it was on the file system.

It seems to be an issue with VFS cache mechanism. Explicit re-synchronization should be used whenever possible.

lpellegr:
@fviale Is it fixed by your last two commits?

fviale:
I cannot say as the patch has not been yet applied at CEA

fviale:
it seems fixed as no more incident as been reported since several months

PR #2528, opened 2016-04-25
#2527 : When accessing files in user/global space from the rest api a desynchronization can occur

fviale:


tobwiens:
:+1: Looks good

Issue #2532, opened 2016-04-29
Add new filterAfterSelection method to scheduling policy


fviale:
This method will allow the scheduling policy to filter nodes are tasks which were selected and scheduled. Useful when stateless selection scripts are not sufficient to make scheduling decisions.

PR #2533, opened 2016-04-29
Add new filterAfterSelection method to scheduling policy #2532

fviale:
Added the new method, empty by default, which needs to be overriden to perform post-selection filtering.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
@@ -36,14 +36,6 @@
  */
 package org.ow2.proactive.scheduler.core;
 
-import java.security.PrivateKey;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -78,6 +70,10 @@
 import org.ow2.proactive.utils.Criteria;
 import org.ow2.proactive.utils.NodeSet;
 
+import java.security.PrivateKey;
+import java.util.*;
```
tobwiens: Can you setup IntelliJ to not merge imports with *

tobwiens:
Looks good: :+1:

tobwiens:
I think we can discuss the design a little bit more: 
The method is a generic method which allows us to implement all sorts of filtering patterns.

Nevertheless only one scheduling method can be used and we can't combine scheduling methods. So we won't be able to use that generic freedom because it will create a mess.

tobwiens:
We could allow the Scheduling Policy to hold filters which are applied by it. That way we remove the filtering logic from the Scheduling policy.

PR #2537, opened 2016-05-03
Update database schema for efficient jobs deletion

lpellegr:


lpellegr:
Retest this please.

lpellegr:
Retest this please.

lpellegr:
Retest this please.

lpellegr:

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
```diff
@@ -6,21 +6,7 @@
...
-import javax.persistence.FetchType;
-import javax.persistence.Index;
-import javax.persistence.JoinColumn;
-import javax.persistence.Lob;
-import javax.persistence.ManyToOne;
-import javax.persistence.OneToMany;
-import javax.persistence.OneToOne;
-import javax.persistence.Table;
-import javax.persistence.Transient;
+import javax.persistence.*;
```
tobwiens: You can setup your idea to not .* import.

tobwiens:
OK :+1:

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
lpellegr: I use our [coding-rules configuration](https://github.com/ow2-proactive/coding-rules/blob/master/settings.jar) that uses the wildcard if there are more than 15 imports for the same package. I agree that wildcard may lead to confusion. We should keep that in mind for our new coding rules. I think Google profile don't use/allow wildcard by default.  
capture d ecran de 2016-05-09 11-59-04

mbenguig:
In the updated code sections, is there any "unique value field" that is currently not indexed ?...

mbenguig:
Ok for me.

lpellegr:
@mbenguig No, unique field values should be indexed by default whatever the database engine is.

lpellegr:
Retest this please.

tobwiens:
@lpellegr I see new commits (after the review ok comments) Do you like another review?

lpellegr:
@tobwiens There are new commits after the review because I have rebased with master. It should be ok, thank you.

PR #2538, opened 2016-05-03
Add Postgres stored procedure example for housekeeping

lpellegr:
as requested by Brian

### DIFF
--- a/config/scheduler/settings.ini
+++ b/config/scheduler/settings.ini
@@ -76,7 +76,10 @@ pa.scheduler.core.removejobdelay=0
 # Set this time to 0 if you don't want the job to be remove automatically.
 pa.scheduler.core.automaticremovejobdelay=0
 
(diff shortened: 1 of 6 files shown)


lpellegr:

tobwiens:
Looks good :+1:, good explanation of the code and config setting. 
Can we guard this with tests somehow? Or can we put this somewhere aside. I think that might outdated soon and we won't remove it. 
What is your opinion on that??

lpellegr:
@tobwiens This is an example working with 6.0.5 only. It is already outdated since the data model changed between version 6 and 7. I already made the same comment to Brian but he wants to have the example to be included in the release.

lpellegr:
I forgot to mention that it is planned to write performance tests soon but they will not be against version 6.0.X.

mbenguig:
ok for me.

Issue #2539, opened 2016-05-03
Allow multiple-node-per-jvm deployment in HostsFileBasedInfrastructure


fviale:
Currently HostsFileBasedInfrastructures (SSH, SSHv2, CLI, AutoUpdate) deploy one node per JVM. 

Similarly to the previous changes for LocalInfrastructure, the SSH infrastructures should be able to deploy multiple nodes per JVM.

fviale:
fixed by be8da44

PR #2540, opened 2016-05-03
Allow multiple-node-per-jvm deployment in HostsFileBasedInfrastructure #2539

fviale:


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/HostsFileBasedInfrastructureManager.java
```diff
@@ -83,6 +85,11 @@
     private Hashtable<String, InetAddress> registeredNodes = new Hashtable<>();
 
     /**
+     * Nodes previously removed
+     */
+    final ConcurrentHashMap<InetAddress, AtomicInteger> removedNodes = new ConcurrentHashMap<InetAddress, AtomicInteger>();
```
lpellegr: You could use the diamong operator, i.e. replace:

```
new ConcurrentHashMap<InetAddress, AtomicInteger>();
```

by:

```
new ConcurrentHashMap<>();
```

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/HostsFileBasedInfrastructureManager.java
```diff
@@ -269,27 +276,54 @@ public void removeNode(Node node) {
...
-            try {
-                this.killNodeImpl(node, host);
-            } catch (Exception e) {
-                logger.trace("An exception occurred during node removal", e);
+            // remember the node removed
+            removedNodes.putIfAbsent(host, new AtomicInteger(0));
+            removedNodes.get(host).incrementAndGet();
+            if (!registeredNodes.containsValue(host)) {
+                // in case all nodes relative to this host were removed kill the JVM
+                freeHosts.put(host, removedNodes.remove(host).intValue());
```
lpellegr: I would have moved this line after `this.killNodeImpl(node, host);` to be sure that a node is deleted and forgotten only if the runtime on the remote node has been really killed, otherwise the information at the Scheduler level may be wrong. 

Same apply for other infrastructures.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/HostsFileBasedInfrastructureManager.java
```diff
@@ -269,27 +276,54 @@ public void removeNode(Node node) {
...
             logger.error("Node " + nodeName +
-                " is not known as a node belonging to this infrastructure manager");
+                    " is not known as a node belonging to this infrastructure manager");
+        }
+    }
+
+    protected boolean anyTimedOut(List<String> nodesUrl) {
+        boolean timeout = false;
+        for (String nodeUrl : nodesUrl) {
+            timeout = timeout || pnTimeout.get(nodeUrl);
```
lpellegr: Would be much better to short circuit as soon as possible:

```java
for (String nodeUrl : nodesUrl) {
    if (pnTimeout.get(nodeUrl)) {
        return true;
    }
}

return false;
```

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/InfrastructureManager.java
```diff
@@ -592,6 +618,14 @@ protected final boolean checkNodeIsAcquiredAndDo(String nodeName, Runnable toRun
 		}
 	}
 
+	protected final boolean checkAllNodesAreAcquiredAndDo(List<String> nodeNames, Runnable toRunWhenOK, Runnable toRunWhenKO) {
+		boolean answer = true;
+		for (String nodeName : nodeNames) {
+			answer = answer && checkNodeIsAcquiredAndDo(nodeName, toRunWhenOK, toRunWhenKO);
```
lpellegr: Would be much better to short circuit as soon as possible the loop:

```java
for (String nodeName : nodeNames) {
    if (!checkNodeIsAcquiredAndDo(nodeName, toRunWhenOK, toRunWhenKO)) {
        return false;
    }
}

return true
```

samples/scripts/deployment/localDeployment.bat
```diff
@@ -7,6 +7,7 @@ set HOST_NAME=%1
 set NODE_NAME=%2
 set NODE_SOURCE_NAME=%3
 set RM_URL=%4
+set NB_NODES=%5
```
lpellegr: This change could break existing scripts that use _localDeployment.bat_. To prevent that, you could check if $5 is defined, if not then a default value should be set.

samples/scripts/deployment/sshDeployment
```diff
@@ -9,6 +9,7 @@ HOST_NAME=$1
 NODE_NAME=$2
 NODE_SOURCE_NAME=$3
 RM_URL=$4
+NB_NODES=$5
```
lpellegr: This change could break existing scripts that use _sshDeployment_. To prevent that, you could check if $5 is defined, if not then a default value should be set.

samples/scripts/deployment/sshDeploymentOneJar
```diff
@@ -9,6 +9,7 @@ HOST_NAME=$1
 NODE_NAME=$2
 NODE_SOURCE_NAME=$3
 RM_URL=$4
+NB_NODES=$5
```
lpellegr: This change could break existing scripts that use _sshDeploymentOneJar_. To prevent that, you could check if $5 is defined, if not then a default value should be set.

lpellegr:
I have tried with success the following Infrastructure:
  - LocalInfrastructure
  - SSHInfrastructure
  - SSHInfrastructureV2

I haven't tested AutoUpdateInfrastructure neither CliInfrastrusture but the changes look ok for me.

lpellegr:
@fviale The title of the Pull Request and your commit title are misleading. Your changes include other infrastructures than SSH infrastructures.

lpellegr:
@fviale Since you are using the existing "number of nodes" one JVM and multiple nodes at the same time, instead of multiple JVMs, we loose a feature. The fact that it is no longer possible to deploy nodes in their own JVM with the Infrastructured which have been altered is for me a regression. 

Since we still have support for enabling non forked tasks, there may have reasons for having each task running in their own JVM, especially to prevent the crash of one task affecting others. Another reason in this configuration is to avoid tasks to share the same JVM settings and pools (e.g. memory pools).  

Also, if I remember correctly, console scripts are executed directly on the node JVM. In that case a company that is using such scripts may prefer a single JVM per node for safety. Or at least, we should fork a process for executing such scripts.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/InfrastructureManager.java
fviale: Not in that case, because of the actions which must be taken when the node is not acquired

lpellegr: Ah ok, I see.

fviale:
Your remark is correct, there can be use cases where separate JVM is required and fork-mode is disabled.

But:
- Multiple customers major concern is that multiple JVM use too much RAM
- Most use cases use forked tasks
- In use cases where non-forked tasks were required, a single JVM with multiple nodes was also required
- The general case where it would be possible to configure both the number of JVMs and number of nodes would make this already complex code even more complex

As a result, we can leave the implementation of the general case for a later time when it's really needed, and focus at the moment to what's most practical for our current use cases.

lpellegr:
It's a way to proceed.

PR #2541, opened 2016-05-04
Remove TestKillWhenNodeHangs test and associated utility class

lpellegr:
The test is ignored by default. Since there are some system tests which cover
what is tested in this test, it can be removed safely.

Besides, the utility class associated to the test causes compilation errors on
Windows 10 with a recent JDK installation (1.8.0_91). It is due to the import
which reference _com.sun_ internal classes which seem not found because
lib/tools.jar is not put in the classpath by default.


lpellegr:

tobwiens:
:+1:

Issue #2544, opened 2016-05-11
Low priority tasks can be scheduled over high priority ones


fviale:
Due to jobs locking mechanism, SchedulingMethodImpl can receive an incomplete number of jobs.

In that case, a low priority job can have its tasks scheduled, while high priority jobs are simply locked during the current scheduling loop.

Additionnaly the InternalPolicy filter can remove high-priority tasks from the list of tasks to be scheduled, making priorities irrelevant.

fviale:
fixed by a group of changes

PR #2549, opened 2016-05-16
Low priority tasks can be scheduled over high priority ones #2544

fviale:
The InternalPolicy filter can remove high-priority tasks from the list of tasks to be scheduled, making priorities irrelevant.

This change :
- removes the InternalPolicy filter
- added a check for priority in SchedulingTaskComparator (to ensure that high and low priority tasks are not deployed concurrently)
- added various tests which check that priority is preserved and that famine does not occur in case high-priority tasks cannot be scheduled


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -157,6 +154,8 @@ public int schedule() {
         //get job Descriptor list with eligible jobs (running and pending)
         Map<JobId, JobDescriptor> jobMap = schedulingService.lockJobsToSchedule();
 
+        logger.debug("jobs selected to be scheduled : " + jobMap);
```
lpellegr: Should be guarded with `if (logger.isDebugEnabled) { ... }` to prevent jobMap String to be created if debug is not enabled. It can really be an issue with 100k jobs in the queue.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -167,23 +166,26 @@ public int schedule() {
...
+
+            LinkedList<EligibleTaskDescriptor> taskRetrievedFromPolicy = new LinkedList<>(policyTasks);
 
             //if there is no task to scheduled, return
             if (taskRetrievedFromPolicy == null || taskRetrievedFromPolicy.size() == 0) {
                 return numberOfTaskStarted;
             }
 
-            logger.debug("eligible tasks : " + taskRetrievedFromPolicy.size());
+            logger.debug("eligible tasks : " + taskRetrievedFromPolicy);
```
lpellegr: Same as for Jobs, please use `if (logger.isDebugEnabled) { ... }`.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -199,6 +201,9 @@ public int schedule() {
                     neededResourcesNumber = getNextcompatibleTasks(jobMap, taskRetrievedFromPolicy,
                             freeResourcesNb, tasksToSchedule);
                 }
+                logger.debug("tasksToSchedule : " + tasksToSchedule);
```
lpellegr: Same as before, please use `if (logger.isDebugEnabled) { ... }`.

scheduler/scheduler-server/src/test/java/functionaltests/job/scheduling/TestJobSchedulingFamineAndPriority.java
```diff
@@ -0,0 +1,212 @@
...
+import org.ow2.proactive.scheduler.examples.EmptyTask;
+import org.ow2.proactive.scripting.SelectionScript;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Several tests related to job scheduling priorities and famine
+ */
+public class TestJobSchedulingFamineAndPriority extends SchedulerFunctionalTestWithRestart {
```
lpellegr: The usual name for "famine" is _starvation_.

lpellegr:
I think `SchedulingTaskComparator` should be unit tested.

marcocast:
I think the code is not formatted properly (using the proactive intellij/eclipse formatter)

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -199,13 +210,35 @@ public int schedule() {
...
+
                 logger.debug("required number of nodes : " + neededResourcesNumber);
-                if (neededResourcesNumber == 0) {
+                if (neededResourcesNumber == 0 || tasksToSchedule.isEmpty()) {
                     break;
                 }
 
                 NodeSet nodeSet = getRMNodes(jobMap, neededResourcesNumber, tasksToSchedule);
 
+                if (nodeSet != null) {
```
marcocast: Could you wrap this logic into a separate private method with a meaningful name? It will help the readability

marcocast: Could you wrap this logic into a separate private method with a meaningful name? It will help the readability

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -234,6 +267,7 @@ public int schedule() {
                             //get back unused nodes to the RManager
                             if (!nodeSet.isEmpty()) {
                                 releaseNodes(currentJob, nodeSet);
+                                freeResourcesNb += (nodeSet.size() + (nodeSet.getExtraNodes() != null ? nodeSet.getExtraNodes().size() : 0));
```
marcocast: Looks like this code is repeated 3 times and it could be wrapped in a private method

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -167,28 +168,38 @@ public int schedule() {
...
+
+            Vector<EligibleTaskDescriptor> policyTasks = currentPolicy
+                    .getOrderedTasks(descriptors);
+
             //ask the policy all the tasks to be schedule according to the jobs list.
             //and filter them using internal policy
-            LinkedList<EligibleTaskDescriptor> taskRetrievedFromPolicy = internalPolicy.filter(currentPolicy
-                    .getOrderedTasks(descriptors));
+
+            LinkedList<EligibleTaskDescriptor> taskRetrievedFromPolicy = new LinkedList<>(policyTasks);
```
lpellegr: Why not changing the signature of Policy#getOrderedTasks to return a linked list? It will prevent to create a copy. Using a non synchronized linked list seems ok since the list is not used by multiple threads.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -167,28 +168,38 @@ public int schedule() {
...
 
             //if there is no task to scheduled, return
             if (taskRetrievedFromPolicy == null || taskRetrievedFromPolicy.size() == 0) {
```
lpellegr: Replace `askRetrievedFromPolicy.size() == 0` by `askRetrievedFromPolicy.isEmpty()`.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -167,28 +168,38 @@ public int schedule() {
...
                 return numberOfTaskStarted;
             }
 
-            logger.debug("eligible tasks : " + taskRetrievedFromPolicy.size());
+            if (logger.isDebugEnabled()) {
+                logger.debug("eligible tasks : " + taskRetrievedFromPolicy);
```
lpellegr: I'm boring but there should be no space before `:`

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
lpellegr: Please factorize this piece of code that is duplicated.

lpellegr:
Please squash commits with the same title.

It would be a good idea to have a Jenkins build that pass before merging.

I noticed that you trying to detect node addition/removal and behave accordingly by counting the number of free nodes. Unfortunately, storing the number of free PANode resources looks not enough to me. You should store the reference to the PANodes or an ID of these nodes that are free, so that you can then compare the number but also the identity. 
It looks possible that 2 nodes are removed and then 2 others (that don't satisfy the same selection scripts) are added between your two checks. In that case, the set of identified free nodes should be passed to `getRMNodes` in order to book nodes you identified with you first snapshot only!

fviale:
Retest me please

fviale:
Retest this please

PR #2551, opened 2016-05-18
Replace return type of scheduling Policy's getOrderedTasks by LinkedList<EligibleTaskDescriptor> #2552

fviale:
Current return type is Vector and implies a list copy in the SchedulingMethodImpl class.

By directly returning a LinkedList, we would avoid this copy


lpellegr:
As discussed, you don't like the idea to change the signature of `Policy#getOrderedTasks` to return a LinkedList. However, this could prevent a copy of lot of Jobs in a method that is invoked really often. 

You mentionned that using directly the returned collection may create issues if the custom policy was using the returned collection for caching values. I understand the issue but the API could be documented to explain that a copy should be explicity returned in that case.

In any case we cannot prevent any crap that is done in the custom policy to affect SchedulingMethodImpl.

lpellegr:
The title of the commit and the pull-request is misleading (does not reflect what you changed), otherwise the change by itself LGTM.

Issue #2553, opened 2016-05-18
Add methods to return the list of free/available/total nodes urls in RMState


fviale:
Currently the from the RMState, the user can only know the number of free/available/total nodes. 

This proposed change adds as well getters which return the above sets with urls.

As there is currently no easy way to get the list of free/available/total nodes urls from the rest api.

fviale:
change merged

PR #2554, opened 2016-05-18
Low priority tasks can be scheduled over high priority ones #2544

fviale:
and
Add methods to return the list of free/available/total nodes urls in RMState #2553

- added getFreeNodes(), getAliveNodes() and getAllNodes() methods in RMState
- added getAllNodesUrls() in NodeSet which returns a set containing all node urls in this node set
- disabled testGetState in RMCoreTest which cannot be run because mocked nodes are not compatible with equals or hashCode use.
- added a new criteria which only select nodes belonging to a given list of nodes urls
- SchedulingMethodImpl now stores the set of free nodes urls at the beginning of the loop.

 - we provide this list to the resource manager at node selection. This way, we are sure that newly available nodes will not be returned by the resource manager and will not impact the scheduling algorithm.
  - with this #feature, we don't need to interrupt the scheduling loop as soon as a new node appear and can safely continue until the end

- refactored the TestJobSchedulingStarvationAndPriority class to make use of replicated tasks jobs (highlights better the priority issues)

### DIFF
--- a/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/RMState.java
+++ b/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/RMState.java
@@ -55,9 +55,9 @@ import java.util.Set;
 @XmlRootElement
 public class RMState implements Serializable {
 
-    private Set<String> freeNodesUrls;
-    private Set<String> aliveNodesUrls;
-    private Set<String> allNodesUrls;
+    private final Set<String> freeNodesUrls;
+    private final Set<String> aliveNodesUrls;
+    private final Set<String> allNodesUrls;
 
     public RMState(Set<String> freeNodesUrls, Set<String> aliveNodesUrls, Set<String> allNodesUrls) {
         this.freeNodesUrls = freeNodesUrls;
--- a/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/RMState.java
+++ b/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/RMState.java
@@ -36,12 +36,12 @@
  */
 package org.ow2.proactive.resourcemanager.common;
 
-import java.io.Serializable;
-
-import javax.xml.bind.annotation.XmlRootElement;
-
 import org.objectweb.proactive.annotation.PublicAPI;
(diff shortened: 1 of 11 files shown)


fviale:
Retest this please

fviale:
Can two people review this last change regarding priority issue ?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -223,11 +226,11 @@ public int schedule() {
                 NodeSet nodeSet = getRMNodes(jobMap, neededResourcesNumber, tasksToSchedule);
```
lpellegr: Is it not possible to prevent most conflicts by passing `freeResources` to  `getRMNodes` and then filtering returned nodes from `SelectionManager#doSelectNodes` by using `freeResources`?

rm/rm-client/src/main/java/org/ow2/proactive/utils/NodeSet.java
```diff
@@ -123,6 +125,24 @@ public void setExtraNodes(Collection<Node> extraNodes) {
...
+    }
+
+    /**
+     * Returns a set containing all nodes urls (standard + extra) included in this node set
+     *
+     * @return set of urls
+     */
+    public Set<String> getAllNodesUrls() {
+        HashSet<String> nodesUrls = new HashSet<>(size() + (extraNodes != null ? extraNodes.size() : 0));
+        for (Node node : this) {
```
lpellegr: Duplicated for loop could be extract as a private method to factorize the code.

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/RMState.java
```diff
@@ -55,14 +55,14 @@
 @XmlRootElement
 public class RMState implements Serializable {
 
-    private int freeNodesNumber;
-    private int totalAliveNodesNumber;
-    private int totalNodesNumber;
+    private Set<String> freeNodesUrls;
```
lpellegr: Could be final and immutable?

fviale:
Retest this please

lpellegr:
LGTM.

PR #2557, opened 2016-05-20
Limit the maximum number of Nodes the Resource Manager accepts.

tobwiens:
Limit the maximum number of Nodes the Resource Manager accepts.

This design limits the maximum number of Nodes by rejecting the Node in the internalAddNodeToCore method. That method is called after the Node is registered inside the Resource Manager and an AddNode event is send.

When the Node limit is reached the Node is rejected and removed from the Node list. One will see an exception inside the Scheduler.log. The Node executable will just be killed from the Resource Manger, without any error message, or sometimes with a stacktrace which indicates that the node was interrupted.

This solution is different to the solution of adding the check, of how many nodes are allowed, into the addNode method. In that method the error message is forwarded and displayed in the node output. That is understandable for the user, but it introduces race conditions and more nodes than wanted can register at the Resource Manager.

Furthermore, the code is compiled with a maxNumberOfNodes = -1; That ensures that this feature is deactivated all of the times. Only if the release script explicitly changes the value inside the .class file it will active. That hinders us from using a final static, because the compiler will do optimizations inside the bytecode, which compiles the constant away and it removes the constant from any if.
As a result, a change of the variable inside the byte code has no effect on the node limitation.


tobwiens:
retest this please

tobwiens:
Retest this please

tobwiens:
Retest this please

build.gradle
```diff
@@ -1,6 +1,6 @@
 buildscript {
     repositories {
-        if (project.hasProperty('local')) mavenLocal()
+        mavenLocal()
```
fviale: you should revert to the original code here

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -577,6 +594,14 @@ public void internalAddNodeToCore(RMNode configuredNode) {
         internalSetFree(configuredNode);
     }
 
+    private boolean isNumberOfNodesLimited() {
+        return maximumNumberOfNodes != null && maximumNumberOfNodes > 0;
```
fviale: zero should be a legal value, so I would write the test rather:

return maximumNumberOfNodes != null && maximumNumberOfNodes >= 0;

fviale:
two comments, the rest is ok

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -560,6 +569,14 @@ public void internalAddNodeToCore(RMNode configuredNode) {
             return;
         }
 
+        //noinspection ConstantConditions
+        if(isNumberOfNodesLimited() && isMaximumNumberOfNodesReached()) {
```
lpellegr: It would be nice to add unit tests in order to check your assertions and ensure that an _AddingNodesException_ is raised or not depending of the situation. Using Mockito, it should be easy.

lpellegr: It looks like you don't have applied formatting (no space between `if` and `(`).

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
lpellegr: don't forget to update the comment on field `private static Long maximumNumberOfNodes;` if you change the semantic.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -147,6 +147,15 @@
 @ActiveObject
 public class RMCore implements ResourceManager, InitActive, RunActive {
 
+    /** Limits the number of nodes the Resource Manager accepts. >-1 means UNLIMITED, <0 enforces the limit.
```
lpellegr: Based on the implementation, the comment seems wrong. `<= -1` is unlimited and `>0` enforces the limit?

rm/rm-server/build.gradle
```diff
@@ -70,4 +70,7 @@ artifacts {
 }
 
 functionalTest.dependsOn rootProject.dist, dist
-clean.dependsOn rootProject.cleanDist
\ No newline at end of file
+clean.dependsOn rootProject.cleanDist
+
+task limitTheMaximumNumberOfNodes(type: maxNodes.ReplaceMaxNumberOfNodesTask) {
```
lpellegr: It would be nice to have an example somewhere about how to invoke the task, and especially how to specify the maximum number of nodes that is desired.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
tobwiens: I understand the documentation writes about the total number of nodes. But the implementation assumes the currently added node to be registered inside the allNodes list. That creates confusion. 
Documentation is clarified accordingly and code is clarified as well.

sonarqube-activeeon:

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. MAJOR [RMCore.java#L552](https://github.com/ow2-proactive/scheduling/blob/bb0cca3bbe4e0323505d358f1735fb272f883123/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java#L552): The Cyclomatic Complexity of this method "internalAddNodeToCore" is 12 which is greater than 10 authorized. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

PR #2561, opened 2016-05-26
Activate task which limit nodes if maxNumberOfNodes is present.

tobwiens:
Activate task which limit nodes if maxNumberOfNodes is present.


fviale:
:+1:

Issue #2565, opened 2016-05-31
Add variables bindings and genericInformation to selection scripts


fviale:
Currently selection scripts contain no binding. It is though useful to access up-to-date job variables, "generic information" and standard info like job name, task id, etc.

lpellegr:
@fviale Why are you creating an issue when you provide a pull-request for this issue on the same day? on Github, issues and pull-requests share the same sequence ID. If you use the search field from the "Issues" tab without specific filter you will see that Issues and Pull-requests are displayed:

https://github.com/ow2-proactive/scheduling/issues?utf8=%E2%9C%93&q=is%3Aopen+

PR #2566, opened 2016-05-31
Add variables bindings and genericInformation to selection scripts #2565

fviale:
- add bindings to the Criteria class (which already contain the selection scripts list)
- add the bindings parameter to the RMNode interface (called on the node from the rm server)
- changed ScriptExecutor behavior to display the selection script output inside the rm logs by default
- in the SchedulingMethodImpl fill, the bindings with the task and job info, generic information, propagated variables if they are available or default job variables
- added a test which verifies that the bindings are correctly set


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -460,6 +467,53 @@ protected NodeSet getRMNodes(Map<JobId, JobDescriptor> jobMap, int neededResourc
...
+            if ((resultSize > 0) && task.handleResultsArguments()) {
+                List<TaskId> parentIds = new ArrayList<>(resultSize);
+                for (int i = 0; i < resultSize; i++) {
+                    parentIds.add(etd.getParents().get(i).getTaskId());
+                }
+                Map<TaskId, TaskResult> taskResults = schedulingService.getInfrastructure().getDBManager()
+                        .loadTasksResults(
+                                job.getId(), parentIds);
+                for (TaskResult taskResult : taskResults.values()) {
+                    if (taskResult.getPropagatedVariables() != null) {
```
sonarqube-activeeon: MAJOR Refactor this code to not nest more than 3 if/for/while/switch/try statements. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS134)

sonarqube-activeeon:
* MAJOR 1 major

lpellegr:
Generic Information defined at the task level only are available from selection scripts. It would be nice to also import Generic information from the Job level. If there is a conflict between some defined at the task level, these last should probablyt have priority.

Please squash or rename commits with same names.

It would be also nice to update the documentation to reference the new genericInformation binding and the usage of variables from selection scripts.

fviale:
all agreeing. Regarding documentation, that was the plan. I need to first add genericInformation bindings to other relevant scripts and then I will do a documentation pull request

PR #2568, opened 2016-06-03
Add Generic Information to task scripts (pre/post/executable/fork/con…

fviale:
…trolFlow)

 This adds a binding genericInformation to all scripts used inside a task

 Added a test to check that default variable bindings and generic information are available and correctly set on all task scripts (pre/post/executable/fork/controlFlow)

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncherInitializer.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncherInitializer.java
@@ -36,15 +36,8 @@
  */
 package org.ow2.proactive.scheduler.task;
 
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
+import com.google.common.collect.ImmutableMap;
 import org.objectweb.proactive.extensions.dataspaces.core.naming.NamingService;
-
 import org.ow2.proactive.scheduler.common.task.ForkEnvironment;
 import org.ow2.proactive.scheduler.common.task.TaskId;
 import org.ow2.proactive.scheduler.common.task.dataspaces.InputSelector;
@@ -52,6 +45,13 @@ import org.ow2.proactive.scheduler.common.task.dataspaces.OutputSelector;
 import org.ow2.proactive.scheduler.common.task.flow.FlowScript;
 import org.ow2.proactive.scheduler.common.util.VariableSubstitutor;
 import org.ow2.proactive.scripting.Script;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 
 /**
@@ -82,13 +82,15 @@ public class TaskLauncherInitializer implements Serializable {
     /** iteration index: task was replicated sequentially */
     private int iterationIndex = 0;
 
+    private ImmutableMap<String, String> genericInformation;
+
     /** DataSpaces needed parameter */
     private List<InputSelector> taskInputFiles = null;
     private List<OutputSelector> taskOutputFiles = null;
     private NamingService namingService;
     private boolean preciousLogs;
 
-    private Map<String, String> variables;
+    private ImmutableMap<String, String> variables;
     private int pingPeriod;
     private int pingAttempts = 1;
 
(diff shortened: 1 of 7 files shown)


fviale:
Could two enthusiasts review this change ?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncherInitializer.java
```diff
@@ -82,6 +81,8 @@
     /** iteration index: task was replicated sequentially */
     private int iterationIndex = 0;
 
+    private Map<String, String> genericInformation;
```
tobwiens: Are generic information keys unique? Can we have several equal keys?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncherInitializer.java
```diff
@@ -274,6 +275,20 @@ public int getReplicationIndex() {
     }
 
     /**
+     * @param genericInformation the generic information of this task
+     */
+    public void setGenericInformation(Map<String, String> genericInformation) {
```
tobwiens: It were nice to have it only set a construction time. What about the executable container, doesn't task related information belong more into it the executable container rather than the tasklauncher?

scheduler/scheduler-server/src/test/java/functionaltests/scripts/TestJobScriptVariables.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * This test checks that variable bindings are available and correctly set in various scripts (pre/post/fork/task)
+ */
+public class TestJobScriptVariables extends SchedulerFunctionalTestNoRestart {
+
+    private static URL jobDescriptor = TestJobScriptVariables.class
+            .getResource("/functionaltests/descriptors/Job_script_variables.xml");
+
+    @Test
+    public void testJobScriptVariables() throws Throwable {
```
tobwiens: Does this test validate that all bindings are inside the JobScript? It seems that it tests the job submission and then of all nodes a re clean afterwards. 
Do you test that the GenericInformation are inserted correctly?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
```diff
@@ -1044,6 +1043,11 @@ protected TaskLauncherInitializer getDefaultTaskLauncherInitializer(InternalJob
         tli.setIterationIndex(getIterationIndex());
         tli.setReplicationIndex(getReplicationIndex());
 
+        HashMap<String, String> gInfo = new HashMap<>();
+        gInfo.putAll(job.getGenericInformation());
+        gInfo.putAll(getGenericInformation());
```
tobwiens: So a task overwrites the generic information of a job. In order to make that obvious you could put that into a private method named with a description.

tobwiens:
A few comments, :+1:

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncherInitializer.java
fviale: yes they are unique and stored in the server in a HashMap. We can have a definition at job level and at task level for the same key. In that case, the task definition overrides the job definition

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
fviale: why not

scheduler/scheduler-server/src/test/java/functionaltests/scripts/TestJobScriptVariables.java
fviale: this test validates that all bindings are present in various task scripts (pre/post/fork/flow). The job fails if a binding is not present in either script

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncherInitializer.java
```diff
@@ -274,6 +275,20 @@ public int getReplicationIndex() {
...
+        this.genericInformation = genericInformation;
+    }
+
+    /**
+     * @return the generic information of this task
+     */
+    public Map<String, String> getGenericInformation() {
```
marcocast: Should we return a copy of this map so we don t risk that the client will change its state by mistake?

fviale: I could make the map immutable

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
```diff
@@ -1044,6 +1043,11 @@ protected TaskLauncherInitializer getDefaultTaskLauncherInitializer(InternalJob
         tli.setIterationIndex(getIterationIndex());
         tli.setReplicationIndex(getReplicationIndex());
 
+        HashMap<String, String> gInfo = new HashMap<>();
```
marcocast: Map<String, String> gInfo = Maps.newHashMap()

marcocast:
Few comments as well 👍

PR #2574, opened 2016-06-16
Update credential file inside node.jar.

tobwiens:
Update credential file inside node.jar.

Starting the node.jar with the command pnp command provided by the rm-portal fails with an authentication exception. The rm.cred file which is included inside the node.jar does not correctly authenticate.

 This change copied the rm.cred inside the resources folder of the rm-node, that will include the certificate inside the node.jar.

 So the node.jar can authenticate successfully.


fviale:
I don't completely understand this change, you seem to only change the content of the credential file.

tobwiens:
Yeah you're right. The node.jar could not connect to the Scheduler because its credentials file was wrong. So I changed it so that the node.jar is able to connect to the RM.

There is another PR which completely removes the cred file for the node.jar (rm-mode gradle project), that makes this here obsolete.

tobwiens:
Credential file was delete with other PR.

PR #2581, opened 2016-06-20
data folder is not copied when installing new ProActive version

fviale:
The data folder was not copied when installing new ProActive version.


lpellegr:
@fviale I am not sure its a good idea to copy the whole data folder. It contains files related to the database. Until now, almost all versions were incompatible regarding database models. We don't ensure backward compatibility nor provide tools to perform the upgrade.

fviale:
You are probably right, but it is still possible to manually delete the database before you start the scheduler, in case there is a database incompatibility. When doing a maintenance version upgrade, compatibility should be enforced. This is more like a best effort, working in a lot of cases.

PR #2582, opened 2016-06-20
Progress file and folder permissions are not open enough for RunAsMe …

fviale:
…mode

 The permissions of progress file and folder should be opened to allow it to work in RunAsMe mode.

The error received (inside a R task) was the following:
Writing progress to file '/tmp/PA_JVM323121899/SSH-xstoocky-3_12/309/-590803201/.tasks-progress/job-309-task-0-ada46071-b960-4179-8f70-ab927f1b5084.progress' with value '100' failed 

When looking at the folder permission and content, it showed the following:
[305t0@xstoocky13.mgps.inra.fr;17:34:18] total 8 
[305t0@xstoocky13.mgps.inra.fr;17:34:18] drwxr-xr-x 2 proagent MGPSgenmic 4096 Jun 20 17:34 . 
[305t0@xstoocky13.mgps.inra.fr;17:34:18] drwxrwxrwx 3 proagent MGPSgenmic 4096 Jun 20 17:34 .. 
[305t0@xstoocky13.mgps.inra.fr;17:34:18] -rw-r--r-- 1 proagent MGPSgenmic 0 Jun 20 17:34 job-305-task-0-6dd94948-7315-4d5d-bb3b-7605cf5eec16.progress 

The progress file does not have write permissions for other users.

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/ProgressFileReader.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/ProgressFileReader.java
@@ -2,6 +2,7 @@ package org.ow2.proactive.scheduler.task;
 
 import org.apache.log4j.Logger;
 import org.ow2.proactive.scheduler.common.task.TaskId;
+import org.ow2.proactive.scheduler.task.utils.ForkerUtils;
 
 import java.io.File;
 import java.io.IOException;
@@ -81,10 +82,10 @@ public class ProgressFileReader {
 
         try {
             Files.createDirectories(progressFileDir);
-
+            ForkerUtils.setSharedPermissions(progressFileDir.toFile(), true);
             progressFile = progressFileDir.resolve(progressFileName);
-
             Files.createFile(progressFile);
+            ForkerUtils.setSharedPermissions(progressFile.toFile(), false);
         } catch (FileAlreadyExistsException e) {
             // ignore file already exists exception
         }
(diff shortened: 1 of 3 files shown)


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/ProgressFileReader.java
```diff
@@ -82,8 +82,13 @@ private void createProgressFile(File workingDir, String progressFileName) throws
         try {
             Files.createDirectories(progressFileDir);
 
+            progressFileDir.toFile().setExecutable(true, false);
```
lpellegr: It would be safer (for preventing some race conditions) to give the permissions when files are created. Static methods `Files.createDirectories` and `Files.createFile` support a second argument for defining permissions. Below is an example:

```
Files.createDirectories(progressFileDir, 
    PosixFilePermissions.asFileAttribute(EnumSet.of(
        PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_WRITE, ...)
```

fviale: Thanks @lpellegr, more elegant indeed.

fviale:
Actually using POSIX permissions is failing on Windows, so I will have to rely on the setWritable etc method. 
More info can be found here: http://stackoverflow.com/questions/14415960/java-lang-unsupportedoperationexception-posixpermissions-not-supported-as-in

fviale:
There was a mistake in this pull request, commit 6d9db52 also fixes issue #2585

PR #2588, opened 2016-06-23
Make TaskLauncher close TaskProActiveDataspaces and its underlying thread pool

lpellegr:
The executor service started by TaskProActiveDataspaces uses non-deamon threads.
Since the executor was not closed, the JVM associated to the task was not
terminated, thus leading to a lot of threads and resources kept in background.

Below is threads usage before the fix, using a Job with 1200 replicated tasks that use dataspaces in intput and output:

capture d ecran de 2016-06-23 08-45-27

and after the fix:

capture d ecran de 2016-06-23 09-09-19

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
@@ -569,7 +569,15 @@ public class TaskProActiveDataspaces implements TaskDataspaces {
     }
 
     @Override
-    public void cleanScratchSpace() {
+    public void close() {
+        if (!executorTransfer.shutdownNow().isEmpty()) {
+            logger.error("Remaining tasks to execute while closing thread pool used for data transfer");
+        }
+
+        cleanScratchSpace();
+    }
+
+    private void cleanScratchSpace() {
         try {
             File folder = getScratchFolder();
             FileUtils.deleteQuietly(folder);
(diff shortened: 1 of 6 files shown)


fviale:
:+1:

lpellegr:
@ow2-proactive/developers Could one more person please review the changes?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
```diff
@@ -569,7 +569,15 @@ public void copyScratchDataToOutput(List<OutputSelector> outputFiles) throws Fil
     }
 
     @Override
-    public void cleanScratchSpace() {
+    public void close() {
+        if (!executorTransfer.shutdownNow().isEmpty()) {
+            logger.error("Remaining tasks to execute while closing thread pool used for data transfer");
```
tobwiens: The error is logged but not acted on. So can it be that there are still files to transfer and the Dataspace is closed? Can't we "notify" the calling code about the inconsistency?

tobwiens:
Wow the improvement is huge, good job :+1:. One remark about possible inconsistencies.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
lpellegr: It should never happen with the current code. There is a synchronization barrier in `TaskProActiveDataspaces` that forces to wait for all tasks submitted in the thread pool. The log is there to notice there is a problem if in the future the code is changed (e.g. the synchronization is changed or removed).

lpellegr:
Thank you for the reviews.

Issue #2597, opened 2016-07-01
In RunAsMe mode, transfered files from User/Global Space cannot be modified


fviale:
As transferred files belong to the ProActive Node user, they cannot be modified by the impersonated account (with standard umask permissions)

PR #2599, opened 2016-07-04
#2597 In RunAsMe mode, transfered files from User/Global Space cannot be modified

fviale:
Set read/write permissions on transferred folders and files when task is RunAsMe.
On linux, set as well execute permissions on folder.
In case of multiple folder hierarchy, set permission recursively on all parents up to the space root.


lpellegr:
:+1:

PR #2608, opened 2016-07-11
Enhancements for TaskProActiveDataspaces

lpellegr:


lpellegr:

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
```diff
@@ -94,13 +94,20 @@
             Executors.newFixedThreadPool(getFileTransferThreadPoolSize(),
                     new NamedThreadFactory("FileTransferThreadPool"));
 
+    /**
+     * Mainly for testing purposes
```
MaelAudren: If the aim of the visibility is to test then the visibility "package" is more accurate than the visibility "protected".

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
```diff
@@ -458,10 +455,39 @@ private void createFolderHierarchySequentially(
...
+            DataSpacesFileObject parent = target.getParent();
+
+            if (isDebugEnabled) {
+                logger.debug("Creating folder " + parent.getRealURI());
+            }
+
+            parent.createFolder();
+        }
+    }
+
```
MaelAudren: Why if the property value is null then you return true ?

MaelAudren:
review done

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
```diff
@@ -94,13 +94,20 @@
             Executors.newFixedThreadPool(getFileTransferThreadPoolSize(),
                     new NamedThreadFactory("FileTransferThreadPool"));
 
+    /**
+     * Mainly for testing purposes
+     */
+    protected TaskProActiveDataspaces() {
+
+    }
+
```
MaelAudren: If the aim of the visibility is to test then the visibility "package" is more accurate than the visibility "protected".

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
```diff
@@ -458,10 +455,39 @@ private void createFolderHierarchySequentially(
...
+            parent.createFolder();
+        }
+    }
+
+    private boolean isCreateFolderHierarchySequentiallyEnabled() {
+        String property = System.getProperty(PA_NODE_DATASPACE_CREATE_FOLDER_HIERARCHY_SEQUENTIALLY);
+
+        return property == null || "true".equalsIgnoreCase(property);
+    }
+
```
MaelAudren: Why if the property value is null then you return true ?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
lpellegr: Suggestion applied.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/data/TaskProActiveDataspaces.java
lpellegr: By default the property is not defined. In such a case I want the create folder hierarchy sequentially feature to be enabled. The property is there to have a mean to disable this last  feature.

lpellegr:
@MaelAudren Thank you for the review. 

@ow2-proactive/developers One more review please.

tobwiens:
:+1: Looks good. 
I have one remark: I see one test with @Test(expected = Exception.class). Might it be important to test specific exceptions under specific conditions?

Issue #2605, opened 2016-07-08
Variables in selection script may contain information of the wrong task


fviale:
This especially appears in a replicate block, where all replicated tasks receive the same task variables (the one corresponding to the first task) instead of different ones.

fviale:
fixed

PR #2618, opened 2016-07-26
Variables in selection script may contain information of the wrong task #2605

fviale:
- declared tasks which contain selection scripts using variables as incompatible.
- Added a unit test which verifies this assertion

 This fix solves the problem without impacting the whole selection mechanism performance. I.e, only tasks having selection scripts using variables will be scheduled at a slower pace than other tasks (similarly to task using topology.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingTaskComparator.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingTaskComparator.java
@@ -100,6 +100,8 @@ public class SchedulingTaskComparator {
         //if the parallel environment is specified for any of tasks => not equal
         boolean isParallel = task.isParallel() || tcomp.task.isParallel();
 
+        boolean selectionScriptUseVariables = (doesSelectionScriptsUseVariables(task) || doesSelectionScriptsUseVariables(tcomp.task));
+
         boolean requireNodeWithTokern = task.getGenericInformation().containsKey(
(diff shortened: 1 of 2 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingTaskComparator.java
```diff
@@ -44,7 +44,7 @@
 
 /**
  * SchedulingJobComparator is used to compare jobs for scheduling.
- * The comparison is made on some particular fields described in this class.
+ * The comparison is made on some particular fields described in this class.ST
```
marcocast: can you remove ST?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingTaskComparator.java
```diff
@@ -109,7 +111,18 @@ public boolean equals(Object obj) {
         // checked before
 
         //add the 6 tests to the returned value
-        return sameSsHash && sameNodeEx && sameOwner && samePriority && !isParallel && !requireNodeWithTokern;
+        return sameSsHash && sameNodeEx && sameOwner && samePriority && !isParallel && !selectionScriptUseVariables && !requireNodeWithTokern;
+    }
+
+    private boolean selectionScriptsUseVariables(InternalTask task) {
```
marcocast: can you rename this method with does in front :)

marcocast:
👍

PR #2640, opened 2016-08-25
Port TaskProActiveDataSpaces integration tests from branch 7.7.X

sophiesongge:


lpellegr:
Sophie, the pull request includes some changes that seem not related to the subject of this PR (e.g. _change project name for the notification templates_). You should start from a clean branch of your fork that is synchronized with the current master. Then, you could fetch the branch _7.7.X_ and [cherry-pick](https://ariejan.net/2010/06/10/cherry-picking-specific-commits-from-another-branch/) the commit associated to the integration tests. 

As you have noticed, there are some compilation errors when intergration tests are imported to _master_. It is due to changes in the API which have been introduced between _7.7.X_ and _master_. I think it would be better to fix the compilation error without changing the API (i.e. using value _false_ for _runAsMe_ parameter in the test and not add a new constructor if possible).

Let me know if you need some help.

Issue #2644, opened 2016-08-31
Node is throwing an exception when stopped


paraita:
The expected behavior when exiting a node is the node to cleanily exit without any exception, but the following exception occurs on a Ctrl-C:
```
Joining Resource Manager at pnp://139.59.212.130:64738/
^CCannot close dataSpace configuration !
java.lang.NullPointerException
	at org.ow2.proactive.resourcemanager.nodesource.dataspace.DataSpaceNodeConfigurationAgent.closeNodeConfiguration(DataSpaceNodeConfigurationAgent.java:276)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.objectweb.proactive.core.mop.MethodCall.execute(MethodCall.java:353)
	at org.objectweb.proactive.core.body.request.RequestImpl.serveInternal(RequestImpl.java:214)
	at org.objectweb.proactive.core.body.request.RequestImpl.serve(RequestImpl.java:160)
	at org.objectweb.proactive.core.body.BodyImpl$ActiveLocalBodyStrategy.serveInternal(BodyImpl.java:552)
	at org.objectweb.proactive.core.body.BodyImpl$ActiveLocalBodyStrategy.serve(BodyImpl.java:485)
	at org.objectweb.proactive.core.body.AbstractBody.serve(AbstractBody.java:426)
	at org.objectweb.proactive.Service.serve(Service.java:128)
	at org.objectweb.proactive.Service.blockingServeOldest(Service.java:214)
	at org.objectweb.proactive.Service.blockingServeOldest(Service.java:189)
	at org.objectweb.proactive.Service.fifoServing(Service.java:150)
	at org.objectweb.proactive.core.body.ActiveBody$FIFORunActive.runActivity(ActiveBody.java:337)
	at org.objectweb.proactive.core.body.ActiveBody.run(ActiveBody.java:175)
	at java.lang.Thread.run(Thread.java:745)
Cannot close dataSpace configuration !
java.lang.NullPointerException
[...]

PR #2645, opened 2016-08-31
Set the cacheserver to be static

paraita:
As the cacheserver can be accessed via static calls the cacheserver need to be declared static.
This is to fix issue #2644


fviale:
:+1:

PR #2665, opened 2016-09-19
Add submit job methods which can use local files or urls in rest sche…

fviale:
…duler client

The Rest scheduler client's submit method can only use a Job object. In a lot of cases, it's useful to submit a job from a local file or a url instead.


marcocast:
👍

PR #2673, opened 2016-09-28
Adding HandleOnlyMyJobsPermission as generic permission to execute me…

marcocast:
…thods of the SchedulerFrontend on other user's jobs

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
@@ -673,14 +673,13 @@ public class SchedulerFrontend implements InitActive, Scheduler, RunActive {
     @Override
     public void listenJobLogs(JobId jobId, AppenderProvider appenderProvider)
             throws NotConnectedException, UnknownJobException, PermissionException {
+        // checking permissions        
+        frontendState.checkPermissions("listenJobLogs", frontendState.getIdentifiedJob(jobId),
+                "You do not have permission to listen the log of this job !");
 
         if (!schedulingService.getListenJobLogsSupport().isEnabled()) {
             throw new PermissionException("Listening to job logs is disabled by administrator");
         }
-
-        // checking permissions        
-        frontendState.checkPermissions("listenJobLogs", frontendState.getIdentifiedJob(jobId),
-                "You do not have permission to listen the log of this job !");
 
         schedulingService.listenJobLogs(jobId, appenderProvider);
     }
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
@@ -36,6 +36,32 @@
  */
 package org.ow2.proactive.scheduler.core;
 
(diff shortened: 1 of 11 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
```diff
@@ -1114,9 +1110,7 @@ public String getTaskServerLogsByTag(String jobId, String taskTag)
     @Override
     @ImmediateService
     public List<SchedulerUserInfo> getUsers() throws NotConnectedException, PermissionException {
-        UserIdentificationImpl ident = frontendState.checkPermission("getUsers",
-                "You don't have permissions to get users");
-        frontendState.checkOwnStatePermission(false, ident);
+        frontendState.checkPermission("getUsers", "You don't have permissions to get users");
```
fviale: how are modified the getUsers method calls ?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -288,13 +288,9 @@ synchronized SchedulerState getState(boolean myJobsOnly)
...
         ListeningUser ui = identifications
                 .get(PAActiveObject.getContext().getCurrentRequest().getSourceBodyID());
-        try {
-            checkOwnStatePermission(myJobsOnly, ui.getUser());
-            return myJobsOnly ? sState.filterOnUser(ui.getUser().getUsername()) : sState;
-        } catch (PermissionException ex) {
-            logger.info(ex.getMessage());
-            throw ex;
-        }
+
```
fviale: I'm afraid, that by this modification, a user will be able to see all other users jobs (task list)

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -547,6 +538,25 @@ synchronized void checkChangeJobPriority(JobId jobId, JobPriority priority) thro
         }
     }
 
+    synchronized void checkPermissions(String methodName, IdentifiedJob identifiedJob, String errorMessage)
+            throws NotConnectedException, UnknownJobException, PermissionException {
+        try {
+            checkJobOwner(methodName, identifiedJob, errorMessage);
```
fviale: as this method is now called only at one place, I think it's not necessary any more to handle it with exceptions.

marcocast: This method should throw an exception, like it was before. so all the existing code will not be changed.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
marcocast: what you mean?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
marcocast: you mean this particular method or this change in general?

marcocast:
@fviale ok, i ll externalise the error messages in a class

PR #2676, opened 2016-09-30
add disable monitoring option when start proactive node

ShatalovYaroslav:


rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/RMNodeStarter.java
```diff
@@ -1059,43 +1071,50 @@ protected ResourceManager registerInRM(final Credentials credentials, final Stri
...
+                            JMXTransportProtocol.RMI).toString());
+                    node.setProperty(JMX_URL + JMXTransportProtocol.RO, sigarExposer.getAddress(
+                            JMXTransportProtocol.RO).toString());
+                } catch (Exception e) {
+                    logger.error("", e);
+                }
             }
-        });
+        }else {
+            logger.info("The monitoring is disabled!");
```
fviale: "JMX monitoring is disabled."

fviale:
LGTM

PR #2680, opened 2016-10-03
add option to disable JMX monitoring

ShatalovYaroslav:


rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/RMNodeStarter.java
```diff
@@ -1062,43 +1074,50 @@ protected ResourceManager registerInRM(final Credentials credentials, final Stri
...
+                } catch (Exception e) {
+                    logger.error("", e);
+                }
             }
-        });
+        }else {
+            logger.info("JMX monitoring is disabled.");
+        }
 
         for (final Node node : nodes) {
```
mbenguig: In a performance point of view, and in the case there are several nodes, it could have been better to let this for loop out of the disabledMonitoring condition (like it currently is), but add a if (!disabledMonitoring)...  Inside. To be discussed/confirmed..

mbenguig: Ok!

marcocast:
LGTM, it would be good if you can create a unit test around the change

fviale:
:+1:

PR #2685, opened 2016-10-06
restart function improvement

yinan-liu:


calendar-service/calendar-service.sh
```diff
@@ -33,25 +33,24 @@ start() {
...
+    echo " *** Starting Calendar Service *** "
+
+    #copy scheduler users to radicale server
+    loginFile="$cwd/config/authentication/login.cfg"
+    rm -rf ~/.config/radicale/users
+    touch ~/.config/radicale/users
+
+    while IFS=':' read -r user password
+    do
+        htpasswd -bs ~/.config/radicale/users "$user" "$password"
```
lpellegr: Are you sure that _htpasswd_ is a common command? It seems not to be the case on my distribution. I checked with a default Docker image for Ubuntu and same problem.

You could add a check and suggest installation of the required packages as it is done in _install.sh_ script:

https://github.com/ow2-proactive/scheduling/blob/master/tools/install.sh#L90

otherwise invoking your script may fail. Or at least explain what are the required packages in the documentation.

calendar-service/calendar-service.sh
```diff
@@ -33,25 +33,24 @@ start() {
...
+    rm -rf ~/.config/radicale/users
+    touch ~/.config/radicale/users
+
+    while IFS=':' read -r user password
+    do
+        htpasswd -bs ~/.config/radicale/users "$user" "$password"    
+    done < "$loginFile"
+
+    # Redirects default and error output to a log file
+    java -Dpa.scheduler.home="$cwd" -Dspring.config.location="$cwd/config/calendar-service/application.properties" -jar "$cwd"/calendar-service/calendar-service*.jar > /dev/null 2>&1 &
```
lpellegr: Since the calendar service is not started by the Scheduler Jetty server, why is the calendar-service configuration file in scheduler _config_ folder? is it because you need to have access to some of its properties from scheduler portal and rm portal?

calendar-service/calendar-service.sh
yinan-liu: i install htpasswd in the privous command

calendar-service/calendar-service.sh
yinan-liu: because I would like to allow user to configure calendar service and scheduler in one place, make the user feel like calendar and scheduler are still a unique application

PR #2694, opened 2016-10-14
Sort files list when creating the ListFile object.

fviale:
Files received when using the VFS api is not correctly sorted. Accordinly it is necessary to sort the listings prior to return them to the user.

### DIFF
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/dataspace/dto/ListFile.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/dataspace/dto/ListFile.java
@@ -70,6 +70,4 @@ public class ListFile {
     public List<String> getFullListing() {
         return fullListing;
     }
-
-
 }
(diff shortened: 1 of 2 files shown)


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/dataspace/dto/ListFile.java
```diff
@@ -38,35 +38,34 @@
...
+        this.fullListing = new ArrayList<>(fullListing);
+        Collections.sort(this.directoryListing);
+        Collections.sort(this.fileListing);
+        Collections.sort(this.fullListing);
     }
 
-    public void setDirectoryListing(List<String> directoryListing) {
-        this.directoryListing = directoryListing;
+    public List<String> getDirectoryListing() {
+        return directoryListing;
```
tobwiens: Can you return a copy of the directoryListing to prevent the caller from getDirectoryListing to access (and possible change) the sorted List.

sonarqube-activeeon:
* MINOR 1 minor

PR #2698, opened 2016-10-17
Task variables: model update

laurianed:


scheduler/scheduler-api/src/main/resources/org/ow2/proactive/scheduler/common/xml/schemas/jobdescriptor/dev/schedulerjob.rnc
```diff
@@ -74,6 +74,7 @@ task = ## A task is the smallest computation unit for the scheduler
         resultPreviewClass?,
         taskDescription?,
         genericInformation?,
+        variables?,
```
fviale: variables order should be placed before genericInformation to respect the same order as the one defined for the job element. (validated with Brian).

laurianed:
Retest this please

PR #2707, opened 2016-10-21
IncreaseFunctionalTestTimeouts

tobwiens:
Increase Functional Test Timeouts

Problem:
- Windows Server 2012 Jenkins Slaves take 1m:51s to execute the flow_loop_out.xml
- The timeout is set to 100 seconds 1m:30s
- try.activeeon.com takes 1m:27s, so the timeout might be too tight
- The RestSmartProxyTest have the TEN_MINUTES (in miliseconds) set to 36000

Solution:
- Increase the timeout to 500 seconds (factor 5)
- The RestSmartProxyTest TEN_MINUTES (in miliseconds) set to 600000


tobwiens:
Retest this please

Issue #2715, opened 2016-11-04
FileNotFoundException in scheduler output


fviale:
The following exception appears in scheduler output when executing a job. The job 2052 was just submitted.

```
java.io.FileNotFoundException: H:\Users\fviale\workspace\fork\scheduling\bin\..\logs\jobs\2052\2052t0 (Le chemin d?accès spécifié est introuvable)
        at java.io.FileOutputStream.open0(Native Method)
        at java.io.FileOutputStream.open(FileOutputStream.java:270)
        at java.io.FileOutputStream.<init>(FileOutputStream.java:213)
        at java.io.FileOutputStream.<init>(FileOutputStream.java:133)
        at org.apache.log4j.FileAppender.setFile(FileAppender.java:294)
        at org.apache.log4j.RollingFileAppender.setFile(RollingFileAppender.java:207)
        at org.apache.log4j.FileAppender.<init>(FileAppender.java:110)
        at org.apache.log4j.RollingFileAppender.<init>(RollingFileAppender.java:79)
        at org.ow2.proactive.utils.appenders.FileAppender.append(FileAppender.java:98)
        at org.ow2.proactive.utils.appenders.MultipleFileAppender.append(MultipleFileAppender.java:65)
        at org.apache.log4j.AppenderSkeleton.doAppend(AppenderSkeleton.java:251)
        at org.apache.log4j.helpers.AppenderAttachableImpl.appendLoopOnAppenders(AppenderAttachableImpl.java:66)
        at org.apache.log4j.Category.callAppenders(Category.java:206)
        at org.apache.log4j.Category.forcedLog(Category.java:391)
        at org.apache.log4j.Category.info(Category.java:666)
        at org.ow2.proactive.resourcemanager.selection.SelectionManager.doSelectNodes(SelectionManager.java:200)
        at org.ow2.proactive.resourcemanager.selection.SelectionManager.selectNodes(SelectionManager.java:179)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:497)
        at org.objectweb.proactive.core.mop.MethodCall.execute(MethodCall.java:353)
[...]

PR #2716, opened 2016-11-04
fix for https://github.com/ow2-proactive/scheduling/issues/2715

fviale:
- Let FileAppender class create the provided file if it does not exist already.
- Remove io exception printing to stdout.


marcocast:
👍

PR #2724, opened 2016-11-15
Refactor RestartMode

lpellegr:
The purpose is to keep the existing behaviour but to provide accessors to index
and description fields. It is a requirement for the new Scheduling API.

Equals method has been rewritten to prevent usage of exceptions which are costly.
However, the semantic of the method is still the same. Since equals was defined,
the missing hashcode method has also been added.


lpellegr:
Retest this please.

lpellegr:
Retest this please.

PR #2729, opened 2016-11-21
remove the pompt of starting calendar service in bat file, as did in …

yinan-liu:
emove the pompt of starting calendar service in bat file, as did in bash file


calendar-service/calendar-service-setup.sh
```diff
@@ -2,44 +2,44 @@
...
+  sudo rm -rf "$script_dir/radicale/linux/Radicale-$radicale_version"
+else
+  echo "Radicale is already installed"  
+  source "$script_dir/radicale.sh" stop
+  rm -rf ~/.config/radicale/*
+fi
 
-echo "Radicale server is installed, starting the server..."
-cd ~/.config/radicale/log
+mkdir -p ~/.config/radicale/collections
```
lpellegr: Extract common prefix that is repeated more than 5 times in a variables.

PR #2730, opened 2016-11-23
Using double for the ram value in the RamPolicy

marcocast:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/ram/RamCompute.java
```diff
@@ -32,21 +32,21 @@ public RamCompute() {
      * 
      * @return the total GIGA of RAM available in a machine minus the RAM preallocated by a task.
      */
-    public long getAvailableRAMInGB() {
+    public double getAvailableRAMInGB() {
 
         OperatingSystemMXBean bean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
 
-        long totRAMAvailable = bean.getTotalPhysicalMemorySize() / GIGABYTE;
+        double totRAMAvailable = bean.getTotalPhysicalMemorySize() / GIGABYTE;
```
fviale: you should add a cast : ((double) bean.getTotalPhysicalMemorySize()) / GIGABYTE

marcocast: ok

PR #2737, opened 2016-11-28
improve test for email notification in order to test event filter

sophiesongge:


scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/core/JobEmailNotificationTest.java
```diff
@@ -107,6 +107,9 @@ public void setUp() throws KeyException {
     public void testIncompleteFrom() throws Exception {
         setSenderAddress(INCOMPLETE_EMAIL);
         InternalJob job = createJob(USER_EMAIL);
+        Map<String, String> genericInfo = job.getGenericInformation();
+        genericInfo.put("NOTIFICATION_EVENTS", "Job running to finished");
+        job.setGenericInformation(genericInfo);
```
yinan-liu: is it necessary ? as the local variable genericInfo is already the genericInformation of the job

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -72,9 +72,8 @@ public boolean doCheckAndSend() throws JobEmailNotificationException {
             logger.debug("Notification emails disabled, doing nothing");
             return false;
         }
-        if (jobStatusList != null && jobStatusList.contains(eventType.toString().toLowerCase())) {
-            logger.debug("Event type is not mentioned to be noticed, doing nothing");
-            return true;
+        if(jobStatusList!=null && !jobStatusList.contains(eventType.toString().toLowerCase())){
```
yinan-liu: jobStatusList is assigned with a new ArrayList<> in line 51, maybe the check jobStatusList != null is no longer necessary here

scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/core/JobEmailNotificationTest.java
sophiesongge: Yes, it is necessary, if we don't check here, it will give a null pointer exception in the test, because the generic information can be empty

sophiesongge: In the test, the job is simulated by setting everything including job id, name generic information etc. If we don't set it here, the generic information is empty

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
sophiesongge: Yes, it is necessary, if we don't check here, it will give a null pointer exception in the test, because the generic information can be empty

sophiesongge:
retest this please

sophiesongge:
retest this please

tobwiens:
Retest this please

PR #2738, opened 2016-11-29
Skip OS Name test

tobwiens:
Skip OS Name test

What does this test test?:
It seems that the test gets os.name (System.getProperty) and takes it as a regular expression.
Then the test gets os.name (System.getProperty) again and matches that against the regular expression.

Problem:
Windows Server 2016 return Windows NT (unknown)
"(unknown)" is not interpreted as a simple string because the brackets are not escaped
Matching against the same string fails.

Solution:
Ignoring the test for now


tobwiens:
Retest this please

tobwiens:
@ow2-proactive/developers two reviews please.

lpellegr:
@tobwiens Why not fixing the test by escaping the String with [Pattern#quote](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#quote(java.lang.String))?

tobwiens:
@lpellegr Very good! Thank you!

fviale:
@tobwiens Why not fixing actually the SelectionUtils.checkOSName() method by escaping the String with Pattern.quote ?

lpellegr:
@fviale I don't think it is a good idea to escape the String inside `SelectionUtils#checkOSName()`. This will break backwards compatibility and prevent users to use regular expressions.

fviale:
@lpellegr , my mistake, if it's supposed to accept regular expressions, then I agree

tobwiens:
Restest this please

tobwiens:
Retest this please

tobwiens:
Retest this please

PR #2742, opened 2016-12-05
Fix SQL error on shutdown when embedded HSQLDB server is used

lpellegr:
The default embedded database must not be closed in a shutdown hook.

Multiple shutdown hooks are defined. Some are used for instance by the
Resource Manager to remove node sources before termination. If the database is
closed before completing the last operation, then some errors will occur
(see logs).

Besides, checking the state of the Scheduler and RM by registering an event
listener or even by polling with periodic method calls will not work since
Scheduler and RM are Active Objects and these are terminated by other shutdown
hooks.

The database connection should be closed by Hibernate when the datasource is
closed. If not, the HSQLDB server will handle in the worst case the JVM shutdown
as an accidental machine failure.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulerStarter.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulerStarter.java
@@ -328,7 +328,21 @@ public class SchedulerStarter {
                     discoveryService.stop();
                 }
 
-                hsqldbServer.stop();
+                // WARNING: do not close embedded HSQLDB server in a shutdown hook.
+                //
+                // Multiple shutdown hooks are defined. Some are used for instance
+                // by the Resource Manager to remove node sources before termination.
+                // If the database is closed before completing the last operation, then
+                // some errors will occur (see logs).
+                //
+                // Besides, checking the state of the Scheduler and RM by registering
+                // an event listener or even by polling with periodic method calls will
+                // not work since Scheduler and RM are Active Objects and these are
+                // terminated by other shutdown hooks executed before the current one.
+                //
+                // The database connection should be closed by Hibernate when
+                // the datasource is closed. If not, the HSQLDB server will handle
+                // in the worst case the JVM shutdown as an accidental machine failure.
             }
         }));
     }


lpellegr:
@ow2-proactive/developers Could 2 people please review the changes?

sonarqube-activeeon:

1. MAJOR [SchedulerStarter.java#L322](https://github.com/ow2-proactive/scheduling/blob/03667d7c82cb11fe30ba34a00db496abd9b4e161/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulerStarter.java#L322): Reduce this anonymous class number of lines from 26 to at most 20, or make it a named class. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS1188)

sonarqube-activeeon:

1. MAJOR [SchedulerStarter.java#L322](https://github.com/ow2-proactive/scheduling/blob/34734e513beaaa2836882c4e2314c769c785adf1/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulerStarter.java#L322): Reduce this anonymous class number of lines from 26 to at most 20, or make it a named class. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS1188)

yinan-liu:
i was thinking the issue happened only on server mode, not on embedded file mode (as it has never had such issue before), so why remove it completely even for embedded mode ? if we dont close it, what will happen on embedded file mode?

lpellegr:
@yinan-liu If you look at the implementation of [SchedulerHsqldbStarter](https://github.com/lpellegr/scheduling/blob/34734e513beaaa2836882c4e2314c769c785adf1/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulerHsqldbStarter.java) you will see that `hsqldbServer.stop();` is doing something when HSQLDB server is used only. In that case, I still call it _embedded_ since this mode is the one enabled by default in the Scheduler without any additional configuration. If HSQLDB file mode is configured in configuration files of the Scheduler and/or RM, then the behaviour is exactly the same as in 7.19.X and below.

sonarqube-activeeon:

1. MAJOR [SchedulerStarter.java#L322](https://github.com/ow2-proactive/scheduling/blob/5c040d9dd7b3ead5269074b24169d939742cc7ac/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulerStarter.java#L322): Reduce this anonymous class number of lines from 26 to at most 20, or make it a named class. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS1188)

PR #2754, opened 2017-01-03
Update the jline version

### LINKED ISSUE #2466: Arrow keys don't work in proactive-client interactive mode on Windows
They display instead UTF8 characters like: ┐¢

It means as well that the history is not available (up/down does not work)

paraita:
This fixes the problem of having JLine not detecting properly the underlying terminal mode, as a result, some character keys can be altered and not function anymore (e.g the backspace key). This also fixes #2466


paraita:
According to what's been discussed with @fviale, I should check quickly that no regression has been introduced to the scriptengines before merging my PR

paraita:
I just did a quick check of usual tasks (groovy, python, javascript, bash) and I didn't noticed any regression

Issue #2743, opened 2016-12-06
When a Task is hanging, it can prevent other tasks from running on the same JVM


fviale:
The attached ThreadDump shows how several Tasks can be blocked, waiting for a Task which is currently executing.

The issue seems related to the fact that a Thread is waiting for a DataSpaceNodeConfigurationAgent.ReadWriteLock.writeLock acquisition. And this prevents all other threads DataSpaceNodeConfigurationAgent.ReadWriteLock.readLock acquisition.

[11542.txt](https://github.com/ow2-proactive/scheduling/files/634287/11542.txt)

Accordingly, ReentrantReadWriteLock implementation seems not adapted to the cleaning cache mechanism, where readLock acquisition should have higher priority than writeLock acqu..

fviale:
The issue did not appear again after the fix, closing it.

PR #2755, opened 2017-01-03
fix for #2743 : When a Task is hanging, it can prevent other tasks from running on the same JVM

fviale:
- cache cleaning is now performed using a trylock call. Without this change the cache cleaning thread could block the execution of tasks on the node while its trying to acquire the node. As the cache cleaning is supposed to be performed only when a node is idle, trylock fits well.
- added a regression test which reproduces the problem.


rm/rm-server/src/test/java/functionaltests/dataspace/DataSpaceNodeConfigurationAgentTest.java
```diff
@@ -0,0 +1,133 @@
...
+                .newActive(DataSpaceNodeConfigurationAgent.class.getName(), null);
+        boolean result = nodeConfigurationAgent.configureNode();
+        PAActiveObject.terminateActiveObject(nodeConfigurationAgent, false);
+
+
+        // submit a main thread which will hold the cache read lock for a long time (40s)
+        Future future = executor.submit(new Runnable() {
+            @Override
+            public void run() {
+                lockAndWait(40000);
```
marcocast: divided by 2 :) ?

Issue #2394, opened 2015-11-30
ProcessTreeKiller does not work on Windows


fviale:
It used to work, but currently ProcessTreeKiller is not able to kill any process on windows.

Running the TestProcessTreeKiller, the test is finished leaving eight processes TestSleep.exe.
Each of them contain 3 PTKCookies:
PROCESS_KILLER_COOKIE    10000_08235f2b-8cd3-415d-95b5-3054eb2c268a
PROCESS_KILLER_COOKIE_node    d63e54e8-1587-487c-8772-987b61dfcb4c
PROCESS_KILLER_COOKIE_TEST    7596808f-50d6-4587-8792-3ecc8354d895

The first one is put by the TaskLauncher, the second one, by the RMNodeStarter, and the third one by the test suite.

None of the PTK triggered by these three components have been able to kill a single process.

We can call that a failure ...

fviale:
Latest release of winp (1.24) seems to fix the issue.

https://mvnrepository.com/artifact/org.jvnet.winp/winp/1.24

fviale:
When a node is started on windows, it displays the following error:

```
Connected to the resource manager at pnp://192.168.1.149:64738/
janv. 03, 2017 12:15:59 PM org.ow2.proactive.process_tree_killer.ProcessTree get
AVERTISSEMENT: Failed to load winp. Reverting to the default
java.lang.UnsatisfiedLinkError: C:\GitHubFork\scheduling\dist\lib\winp.x64.33B272BA81F7889846732466467FA204.dll: %1 nÔÇÖest pas une application Win32 valide
        at java.lang.ClassLoader$NativeLibrary.load(Native Method)
        at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1937)
        at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1822)
        at java.lang.Runtime.load0(Runtime.java:809)
        at java.lang.System.load(System.java:1086)
        at org.jvnet.winp.Native.loadDll(Native.java:196)
        at org.jvnet.winp.Native.load(Native.java:128)
        at org.jvnet.winp.Native.<clinit>(Native.java:57)
        at org.jvnet.winp.WinProcess.enableDebugPrivilege(WinProcess.java:212)
        at org.ow2.proactive.process_tree_killer.ProcessTree$Windows.<clinit>(ProcessTree.java:432)
[...]

PR #2756, opened 2017-01-04
work on #2394 : ProcessTreeKiller does not work on Windows

fviale:
- upgrade winp version to 1.24


mbenguig:
Can we close https://github.com/ow2-proactive/scheduling/issues/2394 ?

PR #2764, opened 2017-01-11
useBashScriptEngineWhichSendsSigtermInNonForkedMode

tobwiens:
Use new version of bash script engine

Problem:
- Bash script engine does not send a SIGTERM when in non forked mode and a task is killed

Solution:
- Use new version of script engine which has that fixed


tobwiens:
Retest this please.

tobwiens:
Retest this please.

PR #2768, opened 2017-01-20
Allow RM nodes to be locked whatever their state is

### LINKED ISSUE #2748: Only free nodes can be locked
There are use cases for a customer where it should be possible to lock a node when it is busy (i.e. running a Task). This would allow waiting for the completion of the Task but prevent others executing.

Checking node state at client side is not an acceptable solution since compare and set will not be atomic, thus leading to race conditions.

As a workaround, it is possible to play with the state of the Scheduler: setting the Scheduler in frozen state will prevent Tasks and Jobs to be started but running ones to complete. However, this solution affects all Jobs and thus nodes. It is not possible for instance to prevent execution on a specific host for a maintenance that is planned.

In the code, there is a specific check performed for the state of the node before locking:

https://github.com/ow2-proactive/scheduling/blob/7.19.X/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java#L1663

It seems that it is not a bug but at least a choice. I don't know if this choice was constrained by something. Deeper investigations would be required.

Personally, I think that locking should be allowed whatever the state of the node is.

lpellegr:
The purpose of this PR is to offer the possibility to lock ProActive Nodes managed by the Resource Manager whatever their state is. Before this PR, a Node can be locked if it is in the _FREE_ state only. 

Besides, the LOCK state was the only state which is users managed but mixed with other states managed by the Scheduler internally. This PR removes the LOCK state from the list of existing Node states and introduces a dedicated field along with a few others to keep tracker of who has locked a Node and at which time.

This PR resolves #2748.


sonarqube-activeeon:
SonarQube analysis reported 3 issues

1. MAJOR [RMStateCaching.java#L131](https://github.com/ow2-proactive/scheduling/blob/bd8175c35ff78271b3c2179a58cc29df5abfbc5c/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMStateCaching.java#L131): Reduce this anonymous class number of lines from 28 to at most 20, or make it a named class. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS1188)
1. MINOR [RMDeployingNode.java#L46](https://github.com/ow2-proactive/scheduling/blob/bd8175c35ff78271b3c2179a58cc29df5abfbc5c/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/rmnode/RMDeployingNode.java#L46): Remove this unused import 'org.ow2.proactive.resourcemanager.common.event.RMNodeDescriptor'. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AUselessImportCheck)
1. MINOR [StringUtility.java#L313](https://github.com/ow2-proactive/scheduling/blob/bd8175c35ff78271b3c2179a58cc29df5abfbc5c/rest/rest-cli/src/main/java/org/ow2/proactive_grid_cloud_portal/cli/utils/StringUtility.java#L313): Remove those useless parentheses. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AUselessParenthesesCheck)

lpellegr:
@ow2-proactive/developers Could 2 people please review the changes?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/rmnode/RMNode.java
```diff
@@ -232,10 +232,30 @@
...
-     * @return true if the node is locked, false otherwise.
+     * Returns a boolean saying whether the node is locked or not.
+     *
+     * @return {@code true} if the node is locked, {@code false} otherwise.
      */
     boolean isLocked();
 
+    /**
+     * Returns the timestamp at which the node has been locked.
+     * <p>
```
paraita: Is that valid comment format ? (no `</p>` ?)

sonarqube-activeeon:

1. MAJOR [RMStateCaching.java#L131](https://github.com/ow2-proactive/scheduling/blob/82c41829184c8cff1219b58247d63384f0650eb9/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMStateCaching.java#L131): Reduce this anonymous class number of lines from 28 to at most 20, or make it a named class. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS1188)

rm/rm-client/src/test/java/org/ow2/proactive/resourcemanager/common/NodeStateTest.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testParseInvalidInput() throws Exception {
+        NodeState.parse("Unknown Input Value");
+    }
+
+    @Test
+    public void testToString() {
+        assertThat(NodeState.FREE.toString()).isNotEqualTo(NodeState.FREE.name());
```
fviale: this test is a bit weird

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -441,33 +468,47 @@ private RMNode getNodebyUrl(String url) {
...
+
         Client client = rmNode.getOwner();
         if (client == null) {
             // node has been just configured, so the user initiated this action is the node provider
             client = rmNode.getProvider();
         }
-        // reseting owner here
+
+        // resetting owner here
         rmNode.setFree();
```
fviale: When the node was previously locked, the node will be set in free state but not put in the freeNodes map. After looking at the code of other methods (lock, unlock). The way the node is handled is obscure to me. Would not a lockedNode collection be clearer to understand ?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -1645,77 +1687,101 @@ private boolean checkNodeAdminPermission(RMNode rmnode, Client client) {
...
     public BooleanWrapper unlockNodes(Set<String> urls) {
+        return map(urls, new Predicate<RMNode>() {
+            @Override
+            public boolean apply(RMNode node) {
+                return internalUnlockNode(node);
+            }
+        }, "unlock");
+    }
+
+    public BooleanWrapper map(Set<String> nodeUrls, Predicate<RMNode> operation, String operationName) {
```
fviale: could you name your method with a less generic name, such as mapOnUrlSet ?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -1645,77 +1687,101 @@ private boolean checkNodeAdminPermission(RMNode rmnode, Client client) {
...
         boolean result = true;
-        for (String url : urls) {
+
+        for (String url : nodeUrls) {
             RMNode rmnode = getNodebyUrl(url);
 
             if (rmnode == null) {
-                logger.warn("Cannot lock unknown node with the following url " + url);
-                result = false;
+                logger.warn("Cannot " + operationName + " unknown node: " + url);
```
fviale: I would add a comma : ", unknown node: "

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/rmnode/RMNode.java
```diff
@@ -267,10 +287,18 @@
...
+     */
+    void lock(Client client);
+
+    /**
+     * Unlock the node.
+     *
+     * @param client the client that unlocks the node.
      */
-    void lock(Client owner);
+    void unlock(Client client);
```
fviale: there was no unlock method ?!?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/rmnode/RMNode.java
lpellegr: Yes, it is valid (Javadoc is based on an very old version of HTML). Look at the seventh bullet of the following web page:

http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/rmnode/RMNode.java
lpellegr: Indeed, there was no unlock method :hole:

rm/rm-client/src/test/java/org/ow2/proactive/resourcemanager/common/NodeStateTest.java
lpellegr: This test is important since `toString` is overridden and not expected to return the same value as the enum name. That's especially why the `NodeState` has a `desc` field.

Regarding the missing tests you mention they are already implemented as system tests: 

https://bitbucket.org/activeeon/scheduling-system-tests/pull-requests/29/update-and-add-system-tests-related-to-the/diff

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
lpellegr: Done.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
lpellegr: Done.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
lpellegr: The main issue is with the naming of the `freeNodes` instance variable. As commented on its declaration, this variable is a list that contains nodes which are _free and not locked_, not nodes which are _free_ only. 

I have renamed _freeNodes_ to _eligibleNodes_ along with a comment that repeats the conditions to satisfy when an `add` or `remove` is performed.

I prefer not to introduce a new collection since it will increase the complexity in order to find nodes that satisfy the eligibility condition.

At the end, the best would be to rename `*FreeNodes*` to `*EligibleNodes*` everywhere but it will break the API since the change will be reflected on the REST service output due to Dozer bean mapping.

sonarqube-activeeon:

1. MAJOR [RMStateCaching.java#L131](https://github.com/ow2-proactive/scheduling/blob/26bce7d4fcd897ee7111dc49885aef8986c2ceae/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMStateCaching.java#L131): Reduce this anonymous class number of lines from 28 to at most 20, or make it a named class. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS1188)

PR #2772, opened 2017-01-26
Update remote visualisation Workflow 

MaelAudren:
- Instead of waiting 1000 seconds to end the task, the remote visualization will stop after 120 seconds.
- The remote visualization and the task can be stopped manually by pressing the "q" button on the keyboard.

### DIFF
--- a/config/workflows/templates/9/job.xml
+++ b/config/workflows/templates/9/job.xml
@@ -1,18 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <job
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xmlns="urn:proactive:jobdescriptor:3.5"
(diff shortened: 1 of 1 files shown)


lpellegr:
@MaelAudren Is there a system test to ensure this Workflow works as expected?

MaelAudren:
@lpellegr It seems that the only test related to the task is about the name and the order of the task among others tasks and I didn't change those values.

lpellegr:
@MaelAudren yes sure, that's why it's maybe the right time to add a system test for this Workflow if none are present?

MaelAudren:
@lpellegr The workflow is related to remote visualisation. I think that adding test that ensure that the workflow it working well would be out off scope or too much work concerning the card.

lpellegr:
OK, as you want. As consequence, it will not be possible to detect regressions on such example.

PR #2779, opened 2017-02-06
useBashScriptEngineWithShutdownHook

tobwiens:
Use bash script engine with shutdown hook

Problem:
- The bash script engine terminates immediately in run as me forked mode, when a task is killed.

Solution:
- Use bash script engine in which that problem is fixed


tobwiens:
retest this please

tobwiens:
retest this please

PR #2787, opened 2017-02-10
Apply code formatting rules

yinan-liu:


yinan-liu:
@lpellegr i didn't add the gitattribute file. Not all projects have it now, only two projects which had encountered windows build issue have this file. In addition, this is not a part of code formatting task.

PR #2796, opened 2017-02-23
Improve Node Pinging Mechanism

fviale:
- when the task launcher ping fails, verify if the node is accessible. If the task launcher is not reachable but the node is still alive, it means that the task has simply initiated its ending sequence (no real node failure). If the node is not reachable, we increase the number of ping attempts.
- added a dedicated threadpool for task pinging to improve regularity of task ping period, not making it dependant of other time consuming operations
- improved the documentation of task ping attempts.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
@@ -35,6 +35,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 
 import org.apache.log4j.Logger;
+import org.objectweb.proactive.core.node.Node;
 import org.ow2.proactive.scheduler.common.NotificationData;
 import org.ow2.proactive.scheduler.common.SchedulerEvent;
 import org.ow2.proactive.scheduler.common.SchedulerStatus;
@@ -56,7 +57,6 @@ import org.ow2.proactive.scheduler.policy.Policy;
 import org.ow2.proactive.scheduler.task.TaskInfoImpl;
 import org.ow2.proactive.scheduler.task.TaskLauncher;
 import org.ow2.proactive.scheduler.task.TaskResultImpl;
-import org.ow2.proactive.scheduler.task.exceptions.ForkedJvmProcessException;
 import org.ow2.proactive.scheduler.task.internal.InternalTask;
 import org.ow2.proactive.scheduler.util.JobLogger;
 import org.ow2.proactive.scheduler.util.TaskLogger;
@@ -939,7 +939,7 @@ public class SchedulingService {
         }
     }
 
-    void pingTaskNode(RunningTaskData taskData) {
+    void getProgressAndPingTaskNode(RunningTaskData taskData) {
         if (!jobs.canPingTask(taskData)) {
             return;
         }
@@ -955,34 +955,36 @@ public class SchedulingService {
                                           new NotificationData<TaskInfo>(SchedulerEvent.TASK_PROGRESS,
                                                                          new TaskInfoImpl((TaskInfoImpl) task.getTaskInfo())));
             }
-        } catch (NullPointerException e) {
-            //should not happened, but avoid restart if execInfo or launcher is null
-            //nothing to do
-            if (tlogger.isDebugEnabled()) {
-                tlogger.debug(task.getId(), "getProgress failed", e);
-            }
-        } catch (IllegalArgumentException e) {
-            //thrown by (1)
-            //avoid setting bad value, no event if bad
-            if (tlogger.isDebugEnabled()) {
-                tlogger.debug(task.getId(), "getProgress failed", e);
-            }
-        } catch (ForkedJvmProcessException e) {
-            //thrown by when user has overridden getProgress method and the method throws an exception
-            // * if forked JVM process is dead
-            //nothing to do in any case
-            if (tlogger.isTraceEnabled()) {
-                tlogger.trace(task.getId(), "getProgress failed", e);
-            }
         } catch (Throwable t) {
-            RunningTaskData runningTask = jobs.getRunningTask(task.getId());
-            if (runningTask != null) {
+            tlogger.debug(task.getId(), "TaskLauncher is not accessible, checking if the node can be reached.", t);
+            pingTaskNodeAndInitiateRestart(task);
+        }
+    }
+
+    private void pingTaskNodeAndInitiateRestart(InternalTask task) {
+
+        RunningTaskData runningTask = jobs.getRunningTask(task.getId());
+        if (runningTask != null) {
+            // We try to ping the node where the task is running to make sure the exception raised is due to a node failure.
+            // We don't consider here other nodes reserved for the task,
+            // as it is the responsibility of the task itself to manage extra nodes lifecycle
+            // in case of complex multinodes task deployment.
+            Node nodeUsedToExecuteTask = runningTask.getNodeExecutor();
+
+            try {
+                nodeUsedToExecuteTask.getNumberOfActiveObjects();
+
+            } catch (Exception e) {
                 int attempts = runningTask.increaseAndGetPingAttempts();
+                String nodeUrl = nodeUsedToExecuteTask.getNodeInformation().getURL();
                 if (attempts > PASchedulerProperties.SCHEDULER_NODE_PING_ATTEMPTS.getValueAsInt()) {
-                    tlogger.info(task.getId(), "node failed", t);
+                    tlogger.error(task.getId(), "node failed " + nodeUrl + ", initiate task restart.", e);
                     restartTaskOnNodeFailure(task);
                 } else {
-                    tlogger.debug(task.getId(), "node failed - waiting while it comes back");
+                    tlogger.warn(task.getId(),
+                                 "cannot contact node " + nodeUrl + " - waiting while it comes back, attempt " +
+                                               attempts,
+                                 e);
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -977,13 +978,18 @@ void pingTaskNode(RunningTaskData taskData) {
         } catch (Throwable t) {
             RunningTaskData runningTask = jobs.getRunningTask(task.getId());
             if (runningTask != null) {
-                int attempts = runningTask.increaseAndGetPingAttempts();
-                if (attempts > PASchedulerProperties.SCHEDULER_NODE_PING_ATTEMPTS.getValueAsInt()) {
-                    tlogger.info(task.getId(), "node failed", t);
-                    restartTaskOnNodeFailure(task);
+                if (NodesPinger.pingNode(runningTask.getNodes().get(0)) == -1) {
```
lpellegr: Why reusing `NodesPinger#pingNode`? I mean, this last makes up to 10 remote calls if there is no connection issue (and this value is hardcoded :fire:). Is this really needed?

lpellegr: Why are you considering the first Node only? if multi-node feature is used, and some Nodes are booked on different machines, don't you think that connection issue with those Nodes should also be considered, even if no ProActive Task is running on them?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -977,13 +978,18 @@ void pingTaskNode(RunningTaskData taskData) {
...
+                    int attempts = runningTask.increaseAndGetPingAttempts();
+                    if (attempts > PASchedulerProperties.SCHEDULER_NODE_PING_ATTEMPTS.getValueAsInt()) {
```
lpellegr: The property `PASchedulerProperties.SCHEDULER_NODE_PING_ATTEMPTS` is bypassed by your new ping attempts. If you don't try to restart immediately, is the restart behavior not changing? I mean is the failure counter (e.g. the one displayed in the Scheduler Web portal) still increasing? what happens if the node never reconnects? is it not waiting/retrying forever?

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/frontend/topology/pinging/NodesPinger.java
```diff
@@ -70,7 +70,7 @@
      * @param node to ping
      * @return ping value to this node from the one it is executing on
      */
-    private Long pingNode(Node node) {
+    public static Long pingNode(Node node) {
 
         final int ATTEMPS = 10;
```
lpellegr: :fire_engine:

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
fviale: You are right, I don't understand why the NodesPinger do these multiple atttemps in the first place. I could simply write another method in the SchedulingService.

fviale: I could add indeed a ping of all nodes, but in that case should we ping all nodes each time we do a call to getProgress ? Meaning that we will reschedule the task even if the TaskLauncher is still responding, and only an extra node fails to respond ?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
fviale: With my change, the property PASchedulerProperties.SCHEDULER_NODE_PING_ATTEMPTS is taken into consideration only if the node is failed. Which seems to me in adequation with the property name. If the node is failed, then the number of ping attempts will increase until it reaches the maximum number of ping attemps, meaning it will not wait forever. And then the task will initiate a restart and increase the failure counter.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
fviale: Actually I know, it's because it needs to compute the average "distance". So this method is inappropriate in our case.

sonarqube-activeeon:

1. MAJOR [SchedulingService.java#L1006](https://github.com/ow2-proactive/scheduling/blob/d61cad258a0851df1861039fe2e4868cf2008cb8/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java#L1006): Either log or rethrow this exception. [rule](http://master.ci.activeeon.com:9000/coding_rules#rule_key=squid%3AS1166)

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -977,17 +978,37 @@ void pingTaskNode(RunningTaskData taskData) {
...
             }
         }
     }
 
+    private boolean pingNode(Node node) {
+        try {
+            node.getNumberOfActiveObjects();
+            return true;
+        } catch (Exception e) {
+            // cannot reach the node
```
marcocast: Should we log the exception?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -977,17 +978,37 @@ void pingTaskNode(RunningTaskData taskData) {
...
             if (runningTask != null) {
-                int attempts = runningTask.increaseAndGetPingAttempts();
-                if (attempts > PASchedulerProperties.SCHEDULER_NODE_PING_ATTEMPTS.getValueAsInt()) {
-                    tlogger.info(task.getId(), "node failed", t);
-                    restartTaskOnNodeFailure(task);
+                // We try to ping the node where the task is running.
+                // We don't consider here other nodes reserved for the task, 
+                // as it is the responsibility of the task itself to manage extra nodes lifecycle 
+                // in case of complex multinodes task deployment.
+                if (!pingNode(runningTask.getNodes().get(0))) {
```
marcocast: One question, if a task reserves 2 nodes, are we sure that the actual node that runs the task is the runningTask.getNodes().get(0) and not runningTask.getNodes().get(1) ?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
lpellegr: OK, I understand but it should be clear in the documentation that nodes booked by the multi-node feature (with the exception of the first running the ProActive Task) are not monitored. As a consequence, a failure on one of these nodes does not cause a redeployment of the Task on other nodes.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
lpellegr: Could you please explain why it is required to ping the running Task in this catch block? If we are in this catch block, it seems that we already have detected a connection issue with the Node. Why do you need to ping again?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
fviale: I need to improve the exception raised to the user indeed.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
fviale: The first node in the list is the node used to execute the task. This is how the NodeSet is built. Maybe there's a way to make it more explicit, I will have a look but it may be a bit longer to implement.

fviale: It is the purpose of this pull request. The current behavior is to detect a node failure when there is an exception calling the getProgress method. However, this occurs regularly simply at the end of the task (the TaskLauncher autokills itself), without any real node failure. You can observe the difference by comparing the number of "node failed" message in this pull request tests:
http://jenkins.activeeon.com/view/Pull%20Requests/job/scheduling-pull-requests/1348/jdk=JDK8,label=Linux/artifact/logs/Scheduler-tests.log  => 1 "node failed" due to a dedicated test.

compared to the scheduling project logs:
http://jenkins.activeeon.com/job/scheduling/jdk=JDK8,label=Linux/lastSuccessfulBuild/artifact/logs/Scheduler-tests.log  => 4 "node failed"

lpellegr: Yes, for sure, with your change failure detection will appear less frequently, but is it not possible to do the same thing at the protocol level (i.e. increase timeout value)? If you ping 50 times, the issue may even appear less frequently if you are lucky (e.g. churn scenario)?

fviale: Increasing timeouts does not solve the problem of false positives when the TaskLauncher kills itself at the end of the task. As in this case, the exception raised will be that the TaskLauncher Active Object does not exist (it's not a timeout issue).

Basically here, there are two responsibilities. The first responsibility of this method is to get the progress of the task and store it. If there is an error while getting this progress (which can happen for several reasons, such as node failure or task ending), then we switch to the second responsibility which is node failure detection. I will create more explicit submethods to make things clearer.

PR #2827, opened 2017-03-23
Add START_AT job into templates

laurianed:
- Add a "Start At" Job, in Template Studio Menu, just below 2 Min CRON.
- Put in the START_AT Variable: 2017-04-01T12:00:00+01:00
- Use the Tasks from the existing Distributed Computing Job to define this "Start At" Job

### DIFF
--- a/config/workflows/templates/17/job.xml
+++ b/config/workflows/templates/17/job.xml
@@ -11,7 +11,7 @@
     <![CDATA[ START_AT generic information is defined for Job Computing PI according to MonteCarlo method. ]]>
   </description>
   <genericInformation>
-    <info name="START_AT" value="2017-04-01T12:00:00+01:00"/>
+    <info name="START_AT" value="2199-01-01T12:00:00+01:00"/>
   </genericInformation>
   <taskFlow>
     <task name="Computation1">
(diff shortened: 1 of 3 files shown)


config/workflows/templates/17/job.xml
```diff
@@ -0,0 +1,116 @@
...
+        xsi:schemaLocation="urn:proactive:jobdescriptor:3.8 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.8/schedulerjob.xsd"
+        name="Start At" projectName="1. Basic Workflows"
+        priority="normal"
+        onTaskError="continueJobExecution">
+
+  <description>
+    <![CDATA[ START_AT generic information is defined for Job Computing PI according to MonteCarlo method. ]]>
+  </description>
+  <genericInformation>
+    <info name="START_AT" value="2017-04-01T12:00:00+01:00"/>
```
MaelAudren: If the aim is to show that the workflow is still pending before the START_AT date then we should increase the default value so it will not be outdated in 7 days.

tobwiens: :+1:

tobwiens:
Just change the START_AT date to 2199 or so and it's okay

laurianed:
The date was given in the specifications, I asked @brianamedro to clarify it

PR #2829, opened 2017-03-27
Change assertion for testSSHInfrastructureV2WithRestartDownNodes

ShatalovYaroslav:
after restart nodes - check not amount of free nodes, but all nodes that are not down

### DIFF
--- a/rm/rm-server/src/test/java/functionaltests/nodesource/TestSSHInfrastructureV2.java
+++ b/rm/rm-server/src/test/java/functionaltests/nodesource/TestSSHInfrastructureV2.java
@@ -179,7 +179,7 @@ public class TestSSHInfrastructureV2 extends RMFunctionalTest {
         RMTHelper.log("Final checks on the scheduler state");
         s = resourceManager.getState();
         assertEquals(NB_NODES, s.getTotalNodesNumber());
-        assertEquals(NB_NODES, s.getFreeNodesNumber());
+        assertEquals(NB_NODES, s.getTotalAliveNodesNumber()); // check amount of all nodes that are not down
     }
 
     @After


rm/rm-server/src/test/java/functionaltests/nodesource/TestSSHInfrastructureV2.java
```diff
@@ -179,7 +179,7 @@ public void testSSHInfrastructureV2WithRestartDownNodes() throws Exception {
         RMTHelper.log("Final checks on the scheduler state");
         s = resourceManager.getState();
         assertEquals(NB_NODES, s.getTotalNodesNumber());
-        assertEquals(NB_NODES, s.getFreeNodesNumber());
+        assertEquals(NB_NODES, s.getTotalAliveNodesNumber()); // check amount of all nodes that are not down
```
marcocast: are alive nodes the sum of pending + busy + free?

lpellegr: The state _Pending_ does not exist for Nodes.

The method `getTotalAliveNodesNumber` returns the number of Nodes which are not _DOWN_:

https://github.com/ow2-proactive/scheduling/blob/master/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java#L699-L699

ShatalovYaroslav: In fact we can not from resourceManager.getState() retrieve: pending, busy nodes amount.
But according to method RMCore::listAliveNodeUrls() it returns all nodes that are not down. I suppose that we expect to check exactly this behavior in this test - the amount of nodes that are not down.

PR #2836, opened 2017-04-10
fix RestSmartProxyTest.testInErrorEventsReception

fviale:
- SchedulerFrontendState was not properly forwarding some notifications
- SmartProxy was not registering to all scheduler events
- changed the test scenario to use finishInErrorTask


fviale:
Retest this please

PR #2842, opened 2017-04-12
Separate functional test by modules

MaelAudren:


MaelAudren:
The change are unecessary because trigger the functionalTest on a specific modules can be done directly through gradle with DIRECTORY:MODULE:functionalTest

Issue #2884, opened 2017-06-05
MemoryLeak: RM DB Node History is never cleaned


fviale:
Node History is updated each time a node state change occurs.

Unfortunately this history is never cleaned and keeps growing indefinitely ...

PR #2887, opened 2017-06-06
#2884: MemoryLeak: RM DB History is never cleaned

fviale:
Added a cron-based mechanism to automatically delete old history events


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/db/RMDBManager.java
```diff
@@ -66,6 +68,8 @@
 
     private final TransactionHelper transactionHelper;
 
+    private Scheduler houseKeepingScheduler;
```
paraita: can we rename it to houseKeepingRm ?

fviale: Is it not called houseKeepingScheduler because it's an instance of Scheduler ?

Issue #2927, opened 2017-07-21
RSA decryption randomly fails because of padding issues


fviale:
Random exceptions appear when decrypting RSA keys. This is a critical issue as it makes jobs randomly fail.

A few examples:
```
[Scheduler output]: org.ow2.proactive.scheduler.task.exceptions.ForkedJvmProcessException: Failed to execute task in a forked JVM
[Scheduler output]: 	at org.ow2.proactive.scheduler.task.executors.ForkedTaskExecutor.createTaskResult(ForkedTaskExecutor.java:130)
[Scheduler output]: 	at org.ow2.proactive.scheduler.task.executors.ForkedTaskExecutor.execute(ForkedTaskExecutor.java:113)
[Scheduler output]: 	at org.ow2.proactive.scheduler.task.TaskLauncher.doTask(TaskLauncher.java:206)
[Scheduler output]: 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[Scheduler output]: 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[Scheduler output]: 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[Scheduler output]: 	at java.lang.reflect.Method.invoke(Method.java:498)
[Scheduler output]: 	at org.objectweb.proactive.core.mop.MethodCall.execute(MethodCall.java:243)
[Scheduler output]: 	at org.objectweb.proactive.core.body.request.RequestImpl.serveInternal(RequestImpl.java:207)
[Scheduler output]: 	at org.objectweb.proactive.core.body.request.RequestImpl.serve(RequestImpl.java:153)
[Scheduler output]: 	at org.objectweb.proactive.core.body.BodyImpl$ActiveLocalBodyStrategy.serveInternal(BodyImpl.java:561)
[Scheduler output]: 	at org.objectweb.proactive.core.body.BodyImpl$ActiveLocalBodyStrategy.serve(BodyImpl.java:486)
[Scheduler output]: 	at org.objectweb.proactive.core.body.AbstractBody.serve(AbstractBody.java:417)
[Scheduler output]: 	at org.objectweb.proactive.Service.serve(Service.java:119)
[Scheduler output]: 	at org.objectweb.proactive.Service.blockingServeOldest(Service.java:203)
[...]

PR #2935, opened 2017-08-02
fix for #2927 RSA decryption randomly fails because of padding issues

fviale:
Issue was not due to cryptography but to the fact that all replicated tasks shared the same executable container. As replicated task are actually an internal task deeply cloned several times, the test which checks that the container is null always return false, meaning that all replicated tasks share the same container.


fviale:
Retest this please

PR #2949, opened 2017-09-08
Downgrade jython to version 2.7.0

fviale:
Recent upgrade of jython 2.7.1 introduced regressions when executing some jython scripts.

Switching back to version 2.7.0 remove those issues.


paraita:
@tobwiens the update of the library was upon a client's request to prevent possible issues. We noticed it instead introduced more problems (I'll create an issue on the jython bugtracker). We can talk about it if you want more details :)

PR #2950, opened 2017-09-12
Downgrade jruby to version 9.0.5.0

fviale:
Downgrade of jython 2.7.1 to 2.7.0 has created a strange regression on the jruby scripts. Try with previous jruby version to see if the issue still occurs.


paraita:
We cannot test locally for any regression before merging ?

PR #2951, opened 2017-09-12
load-examples.groovy relies on an ordered bucket list file for cloud …

mbenguig:
…automation compatibility

### DIFF
--- a/tools/load-examples.groovy
+++ b/tools/load-examples.groovy
@@ -56,11 +56,11 @@ class Main {
 
 		// Deduced variables
 		def catalog_url = scheduler_rest_url.substring(0,scheduler_rest_url.length()-4) + "catalog"
-		def example_dir_path = examples_zip_path.substring(0,examples_zip_path.lastIndexOf("."))
+		def examples_dir_path = examples_zip_path.substring(0,examples_zip_path.lastIndexOf("."))
 
 		println "[" + scriptFileName + "] Variables : "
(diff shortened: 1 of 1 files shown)


mbenguig:
Retest this please.

tools/load-examples.groovy
```diff
@@ -112,168 +112,177 @@ class Main {
...
-		// For each example directory
-		example_dir.eachDir() { dir ->
+		// Retrieve the ordered busket list
+		def ordered_bucket_list_file = new File(examples_dir, "ordered_bucket_list")
+		ordered_bucket_list_file.text.split(",").each { bucket_name ->
 
-			def metadata_file = new File(dir.absolutePath, "METADATA.json")
-			if (metadata_file.exists())
+			// If the bucket directory exists
+			def bucket_dir = new File(examples_dir, bucket_name)
```
jrochas: Don't you want to use a file extension for the new File? bucket_name + ".json" maybe?

MaelAudren:
@mbenguig Do we have tests that check those changes ?

tools/load-examples.groovy
jrochas: Or other extension in fact

mbenguig: I had in mind that , but this workaround (provide an ordered bucket list to follow for the deployment) must be removed as soon as possible, since anyone can create any bucket at anytime.

PR #2965, opened 2017-10-04
secureCatalogLoadExamples

tobwiens:


tools/load-examples.groovy
```diff
@@ -25,6 +59,23 @@ class Main {
 		}
 	}
 
+	def writeToOutput(String output) {
+		println "[" + this.SCRIPT_NAME + "] "+output
```
fviale: try to use a log4j logger instead of printing to stdout

tobwiens: log4j added. I am testing it locally.

tobwiens: Looks good output example is: [2017-10-04 18:42:52,700       main INFO                 o.o.p.scheduler] [load-examples.groovy]  email found? true

PR #2988, opened 2017-11-02
fix inherited variables in pre, post and clean scripts

luispinedamx:
- Also corrected the assignment of job variables to jobInfo.
- Includes functional test.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/TerminationData.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/TerminationData.java
@@ -46,6 +46,7 @@ import org.ow2.proactive.scheduler.common.job.JobId;
 import org.ow2.proactive.scheduler.common.job.JobVariable;
 import org.ow2.proactive.scheduler.common.task.TaskId;
 import org.ow2.proactive.scheduler.common.task.TaskResult;
+import org.ow2.proactive.scheduler.common.task.TaskVariable;
 import org.ow2.proactive.scheduler.common.task.util.SerializationUtil;
 import org.ow2.proactive.scheduler.core.rmproxies.RMProxiesManager;
 import org.ow2.proactive.scheduler.job.InternalJob;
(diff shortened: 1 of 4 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/job/InternalJob.java
```diff
@@ -1282,4 +1282,10 @@ public void setScheduledTimeForRemoval(long scheduledTimeForRemoval) {
         this.jobInfo.setScheduledTimeForRemoval(scheduledTimeForRemoval);
     }
 
+    @Override
+    public void setVariables(Map<String, JobVariable> variables) {
+        super.setVariables(variables);
+        this.jobInfo.setVariables(this.getVariablesAsReplacementMap());
+    }
```
fviale: why not. I think you should do the same with genericInformation. I tested that the jobInfo contains a null generic info when it should not.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/TerminationData.java
```diff
@@ -309,9 +310,23 @@ public VariablesMap getStringSerializableMap(SchedulingService service, TaskTerm
         } else {
             variablesMap.setInheritedMap(fillMapWithTaskResult(taskResult, true));
         }
+        variablesMap.setScopeMap(removeInheritedVarsFromScope(variablesMap, taskData.getTask().getVariables()));
```
fviale: it looks like the following line, appearing earlier in the code: variablesMap.setScopeMap(taskData.getTask().getScopeVariables());
is overwritten by your code, so it should be removed

luispinedamx: fixed @fviale !

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/job/InternalJob.java
luispinedamx: addressed @fviale

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/job/InternalJob.java
```diff
@@ -1282,4 +1282,16 @@ public void setScheduledTimeForRemoval(long scheduledTimeForRemoval) {
...
+
+    @Override
+    public void setGenericInformation(Map<String, String> genericInformation) {
+        super.setGenericInformation(genericInformation);
+        this.jobInfo.setGenericInformation(genericInformation);
+    }
+
```
luispinedamx: Now, both variables and generic information are propagated to all tasks and available through getTaskInfo().getJobInfo().getVariables() (or getGenericInformation()).

PR #2990, opened 2017-11-03
Fix assignment of null to genericInformation

luispinedamx:
This is a fix detected from test:

TestJobAttributes.testJobAttribute():
job.setGenericInformation(null);

genericInformation should not be allowed to be null (we assign instead an empty Map).


luispinedamx:
Retest this please

PR #2995, opened 2017-11-10
Support package installation in multiple buckets and rename tools/Loa…

Aminelouati:
…dExamples.groovy -> tools/LoadPackages.groovy

In this PR: 

I enhanced the current representation of the METADATA.json to support the package installation in multiple bucket. The client can either assign a single bucket value as a String e.g "bucket" : "Machine_Learning" or a multiple bucket value as a list of String e.g "bucket" : ["Machine_Learning", "Analysis", "Bucket_1"]. Both representation are possible.

### DIFF
--- a/tools/LoadPackage.groovy
+++ b/tools/LoadPackage.groovy
@@ -96,7 +96,7 @@ class LoadPackage {
 
     def run(File package_dir) {
         def target_dir_path = ""
-        def bucket = ""
+        def package_buckets_list
         String sessionId = loginAdminUserCredToSchedulerAndGetSessionId()
         //If the package dir is a zip file, create a temporary directory that contains the unzipped package dir
         if (package_dir.getPath().endsWith(".zip")) {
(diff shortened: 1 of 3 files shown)


tools/LoadPackage.groovy
```diff
@@ -96,7 +96,7 @@ class LoadPackage {
 
     def run(File package_dir) {
         def target_dir_path = ""
-        def bucket = ""
+        def package_buckets_list
```
mbenguig: a little bit confusing.. why not buckets_list ?

Aminelouati: yes

tools/LoadPackage.groovy
```diff
@@ -96,7 +96,7 @@ class LoadPackage {
 
     def run(File package_dir) {
```
mbenguig: prefer "void run" since it does not return any object .. and check the other signatures ..(http://groovy-lang.org/style-guide.html -> "When using def in Groovy, the actual type holder is Object")

tools/LoadPackage.groovy
mbenguig: To improve readability.. it could be better to split the run function into specialized smallest ones.
Code sections could being boxed 

from l101 to l114  ( unzipPAckage(xx) )
from l124 to l140 ( populateDataspace(xxx))
from l153 to l287 ( populateBucket(xxxx) )
from l191 to l286 (pushObject(xxx) )
from l231 to l248 (populateTemplateDir(xxx))

marcocast:
retest this please

PR #3005, opened 2017-11-14
Add RM REST endpoint to set a writable JMX attribute.

vinseon:
# New RM REST endpoint to write JMX attributes

## Features

- REST call rely on a new method `setMBeanInfo` (inspired from existing `getMBeanInfo`)
- POST request to `/rest/rm/info/{name}` (inspired from existing GET)
- 1 path parameter `name` to specify the object name of the targeted MBEAN
- 3 query parameters: `type`, `attr`, `value` to update the desired attribute
- a valid session ID is required

## Limitations

- only 'integer' and 'string' attributes are currently supported
- write one attribute per request
- do not call JMX operations

## Usage example

In the following examples, we assume that a valid session ID has been retrieved like this:

```
SESSION_ID=$(curl -X POST -d "username=admin&password=secret" http://localhost:8080/rest/rm/login)
```

####  Update the number of nodes of the NodeSource `remoteNodes` (using a dynamic JMX policy)

```
curl -X POST -H "sessionid:$SESSION_ID" http://localhost:8080/rest/rm/info/org.ow2.proactive.resourcemanager.nodesource.policy:type=remoteNodes?type=integer&attr=Nodes&value=5
```

####  Update the  maximal number of nodes of the NodeSource `azureInfra`

```
curl -X POST -H "sessionid:$SESSION_ID" http://localhost:8080/rest/rm/info/org.ow2.proactive.resourcemanager.nodesource.policy:type=azureInfra?type=integer&attr=MaxNodes&value=100
```

#### Retrieve the attribute `Nodes` of the NS `remoteNodes`

```
curl -X GET -H "sessionid:$SESSION_ID" http://localhost:8080/rest/rm/info/org.ow2.proactive.resourcemanager.nodesource.policy:type=remoteNodes?attr=Nodes
```

Result:
`[{"name":"Nodes","value":5}]`

#### Retrieve the attribute `MaxNodes` of the NS `azureInfra`

```
curl -X GET -H "sessionid:$SESSION_ID" http://localhost:8080/rest/rm/info/org.ow2.proactive.resourcemanager.nodesource.policy:type=azureInfra?attr=MaxNodes
```

Result:
`[{"name":"MaxNodes","value":100}]`

### DIFF
--- a/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
+++ b/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
@@ -30,9 +30,13 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 
+import javax.management.Attribute;
 import javax.management.AttributeList;
+import javax.management.AttributeNotFoundException;
 import javax.management.InstanceNotFoundException;
 import javax.management.IntrospectionException;
+import javax.management.InvalidAttributeValueException;
+import javax.management.MBeanException;
 import javax.management.MBeanInfo;
 import javax.management.ObjectName;
 import javax.management.ReflectionException;
@@ -289,6 +293,40 @@ public class RMProxyUserInterface extends RMListenerProxy implements ResourceMan
     public AttributeList getMBeanAttributes(ObjectName name, String[] attributes)
             throws InstanceNotFoundException, IntrospectionException, ReflectionException, IOException {
         return this.jmxClient.getConnector().getMBeanServerConnection().getAttributes(name, attributes);
+    }
+
+    /**
+     * Set a single JMX attribute of the MBean <code>objectName</code>.
+     * Only integer and string attributes are currently supported, see <code>type</code>.
+     *
+     * @param objectName    the object name of the MBean
+     * @param type          the type of the attribute to set ('int' and 'string' are currently supported)
+     * @param name          the name of the attribute to set
+     * @param value         the new value of the attribute (defined as a String, it is automatically converted according to <code>type</code>)
+     * @throws InstanceNotFoundException
+     * @throws IntrospectionException
+     * @throws ReflectionException
+     * @throws IOException
+     * @throws MBeanException
+     * @throws InvalidAttributeValueException
+     * @throws AttributeNotFoundException
+     */
+    public void setMBeanAttribute(ObjectName objectName, String type, String name, String value)
+            throws InstanceNotFoundException, IntrospectionException, ReflectionException, IOException, MBeanException,
+            InvalidAttributeValueException, AttributeNotFoundException {
+        Object attributeValue;
+        switch (type.toLowerCase()) {
+            case "int":
+                attributeValue = Integer.valueOf(value);
+                break;
+            case "string":
+                attributeValue = value;
+                break;
+            default:
+                return;
+        }
+        this.jmxClient.getConnector().getMBeanServerConnection().setAttribute(objectName,
+                                                                              new Attribute(name, attributeValue));
     }
 
     public BooleanWrapper isNodeAdmin(String nodeUrl) {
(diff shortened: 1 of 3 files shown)


marcocast:
retest this please

vinseon:
retest this please

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
```diff
@@ -291,6 +295,40 @@ public AttributeList getMBeanAttributes(ObjectName name, String[] attributes)
         return this.jmxClient.getConnector().getMBeanServerConnection().getAttributes(name, attributes);
     }
 
+    /**
+     * Set a single JMX attribute of the MBean <code>objectName</code>.
+     * Only integer and string attributes are currently supported, see <code>type</code>.
+     *
+     * @param objectName    the object name of the MBean
+     * @param type          the type of the attribute to set ('int' and 'string' are currently supported)
```
jrochas: Why don't we use an enum instead? How do the other REST endpoints manage types?

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
```diff
@@ -291,6 +295,40 @@ public AttributeList getMBeanAttributes(ObjectName name, String[] attributes)
...
+            case "int":
+                attributeValue = Integer.valueOf(value);
+                break;
+            case "string":
+                attributeValue = value;
+                break;
+            default:
+                return;
+        }
+        this.jmxClient.getConnector().getMBeanServerConnection().setAttribute(objectName,
```
paraita: will potentials `AttributeNotFoundException` be translated into 404 status code from the REST interface ?

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
vinseon: Actually this is the first/only endpoint that requires types, is enum a better approach?

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
vinseon: Well I'm not even sure as most of the code was taken from the exiting `getNodeMBeansInfo` method and therefore I hope that all `InstanceNotFoundException`, `IntrospectionException`, `ReflectionException`, `IOException`, `NotConnectedException`, `MalformedObjectNameException`, and `NullPointerException` are mapped to HTTP response codes.
I'll test it anyway, do you know where REST errors are handled in general?

paraita: I believe RESTEASY maps exceptions to status code 500. I don't know how to define custom error handlers with RESTEASY but in Spring you usually catch exceptions (`InstanceNotFoundException`, `IntrospectionException`, `ReflectionException`, `IOException`, `NotConnectedException`, `MalformedObjectNameException`) manually in the exposed service and throw your custom exception handler instead so you don't expose your internals. Don't bother changing it if that's the case, we don't support it in the other projects anyway. We can always improve that later, there's no hurry.

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
jrochas: Yes, I believe so. Enums only leave a set of possible values to check (and of possible values to use, from the user point of view), unlike the String approach. Enums converts to String easily, and they are much faster to compare in the switch. And all future REST endpoints that need types will be able to use those in a consistent way.

vinseon: Alright that was quite fun to implement, done in 70aee36b6b33a1819156a0ac078cb09189fd6376.
@paraita yet another exception to the method signature :-\

vinseon:
retest this please

PR #3012, opened 2017-11-17
Load proactive package in the catalog from an url

Aminelouati:
In this PR, we covered several Url types and managed smoothly all kind of errors with convenient exception messages. The client is able to load a proactive package directory in the catalog from: 
1. Direct-download zip urls
2. Exposed web directories (supports only tomcat directory listings for the moment)
3. Entire Github repositories given by their http links or downloadable archive format
4. Sub-directories within a github repository
5. Supports forwarded URLS as in shortened bitly urls, etc


rest/rest-cli/src/main/java/org/ow2/proactive_grid_cloud_portal/cli/cmd/sched/InstallPackageCommand.java
```diff
@@ -71,53 +77,61 @@ public void execute(ApplicationContext currentContext) throws CLIException {
...
-    private void validatePackagePath() {
-        File file = new File(PACKAGE_PATH_NAME);
-        if (file.exists()) {
-            if (!(file.isDirectory() || file.getPath().endsWith(".zip"))) {
-                logger.warn(PACKAGE_PATH_NAME + " must be a directory or a zip file.");
-                throw new CLIException(REASON_INVALID_ARGUMENTS,
-                                       String.format("'%s' must be a directory or a zip file.", PACKAGE_PATH_NAME));
+    private String retrievePackagePath() {
+        File file = new File(SOURCE_PACKAGE);
+        if (!SOURCE_PACKAGE.matches(REGEX_URL)) {
```
sophiesongge: can we simplify these 3 "if"?

PR #3015, opened 2017-11-21
limitNumberOfNodes

tobwiens:
Make maximumNumberOfNodes not static

This variable is not static because under some circumstances the compiler put all static initializations into the cinit bytecode block and decided to set this variable to -1 (probably knowing that -1 has the same code path as null). But that made it impossible to overwrite it later on.


tobwiens:
Retest this please

tobwiens:
Retest this please

tobwiens:
retest this please

PR #3016, opened 2017-11-22
Add license policy + license policy test + improve buckets order management

mbenguig:


mbenguig:
Retest this please.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -0,0 +1,142 @@
...
+
+        // If it requires software licenses
+        if (requiredLicenses != null) {
+
+            // Initialize eligibleTasksDescriptorsLicenses
+            if (eligibleTasksDescriptorsLicenses == null)
+                eligibleTasksDescriptorsLicenses = new HashMap<String, FixedSizeArrayList>();
+
+            // Retrieve the maximum licenses numbers per software from the configuration file
+            final Properties properties = LicenseConfiguration.getConfiguration().getProperties();
```
fviale: configuration should not be read for each task. If you have 10000 tasks to evaluate, it would be a huge overhead.

Configuration should be done only once. We can accept for now that the scheduler must be restarted if the configuration changes.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -0,0 +1,142 @@
...
+                if (!canGetLicense(task, currentRequiredLicense, numberOfLicenses)) {
+                    logger.debug("License for " + currentRequiredLicense + " not available, returning false");
+                    return false;
+                }
+            }
+            // Here we manage to give a license for all required software licenses
+            for (int i = 0; i < requiredLicensesArray.length; i++) {
+                eligibleTasksDescriptorsLicenses.get(requiredLicensesArray[i]).add(task);
+            }
+            logger.debug("All licenses are available, returning true");
```
fviale: instead of "returning true/false", better "executing task", "keep task pending"

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/FixedSizeArrayList.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * If needed, contact us to obtain a release under GPL Version 2 or 3
+ * or a different license than the AGPL.
+ */
+package org.ow2.proactive.scheduler.policy.license;
+
+import java.util.ArrayList;
+
+
+public class FixedSizeArrayList extends ArrayList<Object> {
```
fviale: consider using LinkedBlockingQueue which can be used in a un-blocking manner

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -0,0 +1,142 @@
...
+
+        FixedSizeArrayList eligibleTasksDescriptorsLicense = eligibleTasksDescriptorsLicenses.get(currentRequiredLicense);
+
+        // if the maximum number of license is not reached, return true
+        if (!eligibleTasksDescriptorsLicense.isLimitReached()) {
+            return true;
+        }
+        // otherwise remove all terminated tasks and check again if the limit is reached
+        else {
+            Iterator<Object> iter = eligibleTasksDescriptorsLicense.iterator();
```
fviale: put the logic to clean the list in a dedicated method

scheduler/scheduler-server/src/test/java/functionaltests/policy/license/TestLicensePolicy.java
```diff
@@ -0,0 +1,60 @@
...
+
+
+public class TestLicensePolicy extends SchedulerFunctionalTestLicensePolicy {
+
+    private static URL jobDescriptor = TestLicensePolicy.class.getResource("/functionaltests/descriptors/Job_simple_license_policy.xml");
+
+    @Test
+    public void testLicensePolicy() throws Throwable {
+
+        JobId jobId = schedulerHelper.submitJob(new File(jobDescriptor.toURI()).getAbsolutePath());
```
fviale: this test unfortunately does not test that the tasks are executed sequentially instead of in-parallel. 
I would rather : 
 .  configure the policy with single licences
 .  define a workflow containing simply two tasks using the same licences (a replication is not needed)
 . use waitForEventTaskFinished on the two tasks and compare tasks start time / finished time

mbenguig:
Retest this please.

PR #3017, opened 2017-11-22
Ensure Database Compatibilty

fviale:
- RM database: set serialversionuid of 1L to Pincipal classes

 - Scheduler database: create a JobDataVariable table to replace the existing Map stored as LOB with serialversionuid. As the JobVariable class may evolve over time, it is not safe to simply set the serialuid of this class to 1L


fviale:
Retest this

fviale:
retest this please

fviale:
retest this please

PR #3018, opened 2017-11-23
Only set maxNumberOfNodes if property is set

tobwiens:
The maxNumberOfNodes property is most likely set to -1, during every gradle build. That probably hinders us from setting the variable later on.


tobwiens:
Retest this please

PR #3022, opened 2017-11-27
Remove previously persisted node urls when redeploying nodes on host

jrochas:
- remove all the node URLs for a given host when it is redeployed in HostsFileBaseInfrastructureManager, because these URLs are not contactable anymore and all the nodes on this host will be redeployed. We need to do this to avoid having more nodes than required handled per host.
- retrieve the node source information from database if the node source acannot be found in memory before updating and persisting it.


marcocast:
This need to be tested in LNG env as well

jrochas:
retest this please

jrochas:
retest this please

jrochas:
retest this please

PR #3026, opened 2017-11-28
Adding submitFromCatalog logic in the schedulerapi

marcocast:
With this changes in any jvm tasks it will be possible to submit workflows that are stored in the catalog in a user friendly way : 

schedulerapi.submitFromCatalog("bucketID","WorkflowName")

or 
Map<String, String> variables = new HashMap<String, String>();
variables.put("KEY","VALUE");
schedulerapi.submitFromCatalog("bucketID","WorkflowName", variables)

A full set of specific system tests will cover this feature (too complex to cover it with unit tests)


config/scheduler/settings.ini
```diff
@@ -10,6 +10,9 @@ pa.scheduler.home=.
 # When the server has a public endpoint, different from the hostname available on the machine, this property should be used to correctly set the url
 #pa.scheduler.rest.url=
 
+#Catalog rest endpoint
+pa.catalog.rest.url=http://localhost:8080/catalog
```
fviale: Maybe there is some work to do in the class SchedulerStarter to makes configuration easier. Currently the rest url is inferred dynamically (if not set explicitly) the catalog could also be.

rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/ISchedulerClient.java
```diff
@@ -178,6 +178,48 @@ JobId submit(File job, Map<String, String> variables)
...
+     * @param bucketId
+     * @param workflowName
+     * @param variables
+     * @return
+     * @throws NotConnectedException
+     * @throws PermissionException
+     * @throws SubmissionClosedException
+     * @throws JobCreationException
+     */
+    JobId submitFromCatalog(String bucketId, String workflowName, Map<String, String> variables)
```
fviale: I think it's important to include as well an option revision parameter.

marcocast: In the catalog client project the revision are not supported. I can implement it but in a separate moment as it will be more complex

PR #3031, opened 2017-11-29
Handling of outdated node source DB updates

jrochas:
Node source updates that happened after a node source is removed was causing a lot of error traces in the logs.
- refactor data structure used in RMDBManagerBuffer to have a dedicated NodeOperation class
- remove blocking wait when node source is shut down (to avoid blocking the RMCore)
- put info level for logging of node source add, node source removal, node add, node removal


jrochas:
Retest this please

PR #3034, opened 2017-11-30
Add the PAMR over SSH new parameter

paraita:


paraita:
@vinseon thanks for your comment, I've modified the description to explain in what use-cases this new parameter can be relevant. The _why_ of this parameter will be explained in the documentation. I reuse your input for that ! Can you have a look at https://github.com/ow2-proactive/documentation/pull/365 when you have a moment please ?

PR #3037, opened 2017-12-01
Lazy fetch mode for scripts defined by urls 

fviale:
When a script is defined by an URL, prior mechanism was to fetch the script immediately at Script instance creation. This change allows to configure the script instance for early or late binding. Late binding means that the url will be resolved at execution time.

Inside the scheduler all scripts will be in lazy mode, except for selection scripts. In case of selection scripts, lazy mode cannot be used as the scheduler server needs the script content during the scheduling loop, prior to the script execution on the nodes.

Also this change adds the possibility to define a script by url and language (when for example, it is not possible to infer the script language from the url value). A new version of the job schema is defined.

 - Refactored HttpClientBuilder > CommonHttpClientBuilder to allow its integration in common-api project.
 - Created CommonHttpResourceDownloader, which makes use of the above class to download http/https resources.
 - Modified FileUtils to extract filename and extension from an url
 - Added a fetchScript method to Script class, this method either return an already defined inline script or try to
 download the defined url, without modifying the internal script object structure.
 - Changed the schemas, xml parser, xml serializer to integrate the <file url="" language="" /> syntax
 - Added in the ScriptData and SelectionScriptData the script url definition
 - When script authorization is enabled, use the fetchScript method to get the script content before authorizing the script.


common/common-api/build.gradle
```diff
@@ -1,5 +1,6 @@
 dependencies {
     compile 'commons-codec:commons-codec:1.10'
+    compile 'org.apache.httpcomponents:httpclient:4.5.2'
```
marcocast: is this lib going to create some conflict with other projects?

common/common-api/src/main/java/org/ow2/proactive/http/CommonHttpResourceDownloader.java
```diff
@@ -0,0 +1,155 @@
...
+                throw new IOException(String.format("Cannot access resource %s: code %d", url, status.getStatusCode()));
+            }
+
+            return readContent(response.getEntity().getContent());
+        }
+    }
+
+    public UrlContent getResourceContent(String sessionId, String url, boolean insecure) throws IOException {
+
+        CommonHttpClientBuilder builder = new CommonHttpClientBuilder().maxConnections(CONNECTION_POOL_SIZE)
```
marcocast: this first part is the same as the other method getResource(.... and could be extracted in a private method

common/common-api/src/main/java/org/ow2/proactive/http/CommonHttpResourceDownloader.java
```diff
@@ -0,0 +1,155 @@
...
+            while ((tmp = buf.readLine()) != null) {
+                builder.append(tmp).append("\n");
+            }
+
+            return builder.toString();
+        }
+    }
+
+    public static class UrlContent {
+        private String content;
```
marcocast: final

common/common-api/src/main/java/org/ow2/proactive/http/CommonHttpResourceDownloader.java
```diff
@@ -0,0 +1,155 @@
...
+
+        private String fileName;
```
marcocast: final

common/common-api/src/main/java/org/ow2/proactive/scripting/Script.java
```diff
@@ -505,4 +622,29 @@ public void overrideDefaultScriptName(String defaultScriptName) {
             scriptName = defaultScriptName;
         }
     }
+
+    private static class ScriptContentAndEngineName {
```
marcocast: can you make tis class immutable and pass those 2 params in the constructor?

common/common-api/src/main/java/org/ow2/proactive/utils/FileUtils.java
```diff
@@ -122,11 +124,49 @@ public static File createTempDirectory(String prefix, String suffix, File direct
...
+     * protocol://host:port/leading_path/filename.extension/trailing_path?query
+     *
+     * If there is no dot character inside the URL path, the file name returned is based on the last occurrence of
+     * the slash character:
+     * protocol://host:port/leading_path/filename?query
+     *
+     * @param url input url
+     * @return a filename with its extension.
+     */
+    public static String getFileNameWithExtension(URL url) {
```
marcocast: All this method can be replaced by the URL utility methods if i m not wrong

common/common-api/build.gradle
fviale: I double checked that, it's basically the lib which is used in project common-http.

PR #3048, opened 2017-12-13
Add the possibility to create/modify third-party credentials for standard users

fviale:
Standard users should be able to use their own credentials.


fviale:
@marcocast guests should not be allowed, and providers neither. I think there is an error with the current guests permission, they should not be able to submit jobs.

marcocast:
retest this please

PR #3058, opened 2017-12-20
fix NullPointerException in byte array conversion

luispinedamx:
- This fix prevents a NullPointerException when receiving a null input instead of a Map in ObjectByteConverter.
- Other methods of the class already had this `null` validation.
- The exception was identified when attempting to convert a null value for propagatedVariables:
```
Caused by: java.lang.NullPointerException
                at org.ow2.proactive.utils.ObjectByteConverter.mapOfBase64StringToByteArray(ObjectByteConverter.java:223)
                at org.ow2.proactive.scheduler.rest.data.TaskResultImpl.<init>(TaskResultImpl.java:70)
```
... I also corrected a typo ;)

### DIFF
--- a/common/common-api/src/main/java/org/ow2/proactive/utils/ObjectByteConverter.java
+++ b/common/common-api/src/main/java/org/ow2/proactive/utils/ObjectByteConverter.java
@@ -42,7 +42,7 @@ import org.apache.commons.codec.binary.Base64;
 
 /**
  * Utility functions for converting object to a byte array,
- * and vis versa.
+ * and vice versa.
  * <p>
  * This class can also compress stream
  *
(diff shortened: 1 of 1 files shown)


common/common-api/src/main/java/org/ow2/proactive/utils/ObjectByteConverter.java
```diff
@@ -220,6 +220,9 @@ public static Serializable base64StringToSerializable(String input) throws IOExc
     }
 
     public static Map<String, byte[]> mapOfBase64StringToByteArray(Map<String, String> input) {
+        if (input == null) {
+            return null;
```
jrochas: I don't know what the others think, but what about returning an empty map to prevent further NPE?

luispinedamx: @jrochas I thought the same, I also thought of initializing the propagatedVariables map of TaskResultImpl (further down in my stack) to an empty map from the beginning as it's the only usage of the method. Then I saw that other methods in the class also had this `return null` (maybe for a purpose, I didn't check them all tbh).

luispinedamx: @jrochas in your opinion should we return empty objects instead of null in all cases?

PR #3060, opened 2017-12-21
Return empty map instead of null in byte conversion

luispinedamx:
correction of previous PR


luispinedamx:
Made a mistake here.

PR #3069, opened 2018-01-03
Add fix for Hibernate issue with TextBlob.

gheon:
See https://github.com/jhipster/generator-jhipster/issues/5995.


gheon:
See also: https://hibernate.atlassian.net/browse/HHH-11614

tobwiens:
Looks good.

Just to note it here: This issue was fixed in hibernate 5.2.11 and JHipster 4.8.2. So this fix can be safely removed when we upgrade the dependencies in the future.

vinseon:
@tobwiens true, I tried to update it but that's not easy, seems we have too much old custom data type :-\

PR #3078, opened 2018-01-09
Performance evaluation improvements

Mykhailenko:
Change timeout behavior.
There is two types of timeouts in JUnit. 
Timeout as a Rule and as an attribute of Test annotation.
In the first case, timeout sets whole time for Before, Test, and After methods. In case we break this timeout in Test, After will not be called.
In second case, when timeout is set as an attribute of Test annotation, breaking timeout will cause After been called.

Thus, I have increased timeout set by Rule twice. and put timeouts for test method. Moreover, I moved Before method in Test method itself, because the longest execution was in Before method. So now, even if one test fails because of timeout, other will be executed anyway. Before, failing by timeout will not call After method, which would not kill RM or Scheduler, which prevented other test cases form execution.


scheduler/scheduler-server/src/test/java/performancetests/recovery/BaseRecoveryTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Rule;
+import org.junit.rules.Timeout;
+import org.objectweb.proactive.core.config.CentralPAPropertyRepository;
+
+import functionaltests.utils.SchedulerFunctionalTestWithCustomConfigAndRestart;
+
+
+public class BaseRecoveryTest extends SchedulerFunctionalTestWithCustomConfigAndRestart {
```
sonarqube-activeeon: BLOCKER Add some tests to this class. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2187)

sonarqube-activeeon:
* ![BLOCKER][BLOCKER] 1 blocker

[BLOCKER]: https://sonarsource.github.io/sonar-github/severity-blocker.png 'Severity: BLOCKER'

PR #3084, opened 2018-01-11
Nodes & Task Recovery: Aggregate results in csv file

Mykhailenko:
Change timeout behavior.
There is two types of timeouts in JUnit.
Timeout as a Rule and as an attribute of Test annotation.
In the first case, timeout sets whole time for Before, Test, and After methods. In case we break this timeout in Test, After will not be called.
In second case, when timeout is set as an attribute of Test annotation, breaking timeout will cause After been called.

Thus, I have increased timeout set by Rule twice. and put timeouts for test method. Moreover, I moved Before method in Test method itself, because the longest execution was in Before method. So now, even if one test fails because of timeout, other will be executed anyway. Before, failing by timeout will not call After method, which would not kill RM or Scheduler, which prevented other test cases form execution.


scheduler/scheduler-server/src/test/java/performancetests/recovery/BaseRecoveryTest.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Rule;
+import org.junit.rules.Timeout;
+import org.objectweb.proactive.core.config.CentralPAPropertyRepository;
+
+import functionaltests.utils.SchedulerFunctionalTestWithCustomConfigAndRestart;
+
+
+public class BaseRecoveryTest extends SchedulerFunctionalTestWithCustomConfigAndRestart {
```
sonarqube-activeeon: BLOCKER Add some tests to this class. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2187)

sonarqube-activeeon:
* ![BLOCKER][BLOCKER] 1 blocker

[BLOCKER]: https://sonarsource.github.io/sonar-github/severity-blocker.png 'Severity: BLOCKER'

Mykhailenko:
retest this please

Mykhailenko:
retest this please

scheduler/scheduler-server/src/test/java/performancetests/recovery/BaseRecoveryTest.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Rule;
+import org.junit.rules.Timeout;
+import org.objectweb.proactive.core.config.CentralPAPropertyRepository;
+
+import functionaltests.utils.SchedulerFunctionalTestWithCustomConfigAndRestart;
+
+
+@SuppressWarnings("squid:S2187")
```
jrochas: Could you comment the suppress warning?

Mykhailenko: Withour this supperss it complains that there are no test which covers this class. I need this class just to have a bigger timeout for performance tests

scheduler/scheduler-server/build.gradle
```diff
@@ -64,8 +64,12 @@ task('functionalTest', type: Test).configure schedulingFunctionalTestConfigurati
 task('performanceTest', type: Test).configure schedulingPerformanceTestConfiguration
 
 task verifyPerformanceChange(type: Test){
-    systemProperty 'pathToStorage', System.getProperty('pathToStorage')
-    systemProperty 'threshold', '0.3'
+    systemProperties << ['threshold': System.getProperty('threshold')]
```
jrochas: Maybe we could give a more precise name for the threshold, like performanceTestThreshold and which could also indicate how the threshold is computed (is it a multiple of a previous run value, an arbitrary number, etc...)

Mykhailenko: You are right

scheduler/scheduler-server/src/test/java/performancetests/verification/VerifyChangeTest.java
```diff
@@ -111,14 +113,15 @@ public void changeShouldBeLessThanThreshold() throws IOException {
     }
 
     private List<File> getTwoLastFiles() {
-        final String pathToStorage = System.getProperty("pathToStorage");
-        File folder = new File(pathToStorage);
+        File folder = new File(System.getProperty("pa.rm.home"));
```
jrochas: So the performance files will be at the root of `pa.rm.home`?

Mykhailenko: Yes. I use Copy to Slave plugin of Jenkins. And it copies files directly to the workspace of slave == pa.rm.home

PR #3085, opened 2018-01-12
Refactor functional tests to use junit test suites

fviale:
- Created two test suites on each project, a StandardTestSuite used in pull requests and a RegressionTestSuite
 - StandardTestSuite tries to group tests which runs in the same environment
 - minor changes in scheduler tests to use a non-restart mode whenever possible
 - minor reaffectations to regression tests


build.gradle
```diff
@@ -186,11 +186,10 @@ task wrapper(type: Wrapper) {
 
 ext.schedulingFunctionalTestConfiguration = {
 
-    include 'functionaltests/**'
+    include 'functionaltests/StandardTestSuite*'
 
-    if(!project.hasProperty('regression')){
-        def regression_tests = new File(project.rootDir, "/regression_tests.txt")
```
jrochas: The `regression_tests.txt` file should also be removed as part of this PR.

rm/rm-server/src/test/java/functionaltests/RegressionTestSuite.java
```diff
@@ -0,0 +1,49 @@
...
+import functionaltests.nodesrecovery.RecoverSSHInfrastructureV2Test;
+import functionaltests.nonblockingcore.NonBlockingCoreTest;
+import functionaltests.permissions.TestNSNodesPermissions;
+import functionaltests.selectionscript.SelectionScriptTimeOutTest;
+
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({ NonBlockingCoreTest.class, TestNSNodesPermissions.class, SelectionScriptTimeOutTest.class,
+                      RecoverLocalInfrastructureTest.class, RecoverSSHInfrastructureV2Test.class,
+                      TestSSHInfrastructureV2RestartDownNodesPolicy.class })
```
jrochas: I would add here the `TestLocalInfrastructureRestartDownNodesPolicy.class` as well.

rm/rm-server/src/test/java/functionaltests/StandardTestSuite.java
```diff
@@ -0,0 +1,94 @@
...
+
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({ AuthenticationTest.class, ConnectionTest.class, ConnectionTest2.class,
+                      DataSpaceNodeConfigurationAgentTest.class, NodeHistoryTest.class, NodeSourcesTest.class,
+                      RMDBManagerBufferTest.class, RMDBManagerTest.class, TestExecRemote.class,
+                      AddGetDownRemoveTest.class, AddGetReleaseRemoveTest.class, AddGetRemoveTest.class,
+                      AddGetTest.class, ResourceManagerJMXTest.class, RMProxyUserInterfaceTest.class,
+                      TestRMMonitoring.class, TestRMNodeMonitoring.class,
+                      TestLocalInfrastructureRestartDownNodesPolicy.class, TestLocalInfrastructureStaticPolicy.class,
```
jrochas: As per the previous comment, I would remove here the `TestLocalInfrastructureRestartDownNodesPolicy.class`

scheduler/scheduler-server/src/test/java/functionaltests/RegressionTestSuite.java
```diff
@@ -0,0 +1,96 @@
...
+                      // Tests with scheduler restart
+                      TestProcessTreeKiller.class,
+
+                      // Tests custom configuration
+                      TestJobKilled.class,
+
+                      TestDataspaceConcurrentTransfer.class, TestDataspaceConcurrentKilling.class,
+                      TaskReconnectionWithForkedTaskExecutorTest.class,
+                      TaskReconnectionWithInProcessTaskExecutorTest.class,
+                      TaskRecoveryWhenNodesAreReservedInBatchTest.class })
```
jrochas: `TaskReconnectionWithForkedTaskExecutorTest.class`, `TaskReconnectionWithInProcessTaskExecutorTest.class`, and `TaskRecoveryWhenNodesAreReservedInBatchTest.class` are tests with scheduler restart. We can move them to the `// Tests with scheduler restart` category above.

scheduler/scheduler-server/src/test/java/functionaltests/RegressionTestSuite.java
```diff
@@ -0,0 +1,96 @@
...
+                      TestWorkflowReplicateJobs.class, TestWorkflowReplicateJobs2.class,
+                      TestWorkflowReplicateJobs3.class,
+
+                      // Tests with scheduler restart
+                      TestProcessTreeKiller.class,
+
+                      // Tests custom configuration
+                      TestJobKilled.class,
+
+                      TestDataspaceConcurrentTransfer.class, TestDataspaceConcurrentKilling.class,
```
jrochas: In my opinion, both `TestDataspaceConcurrentTransfer.class` and `TestDataspaceConcurrentKilling.class` tests should go to the standard test suite, even though they often fail, because (1) their load actually do emphasize mistakes (and we want to catch them as soon as possible) and (2) they are quite fast to execute.

scheduler/scheduler-server/src/test/java/functionaltests/RegressionTestSuite.java
```diff
@@ -0,0 +1,96 @@
...
+import functionaltests.workflow.complex.TestWorkflowReplicateJobs;
+import functionaltests.workflow.complex.TestWorkflowReplicateJobs2;
+import functionaltests.workflow.complex.TestWorkflowReplicateJobs3;
+
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+                      // DB Tests
+                      SchedulerDbManagerConcurrencyTest.class, TestJobAttributes.class, TestJobRemove.class,
+                      TestMultipleTasks.class,
```
jrochas: I think in general DB tests should not be moved to regression tests, as we modify the tables or impact on the persisted objects quite often.

build.gradle
fviale: agreed

rm/rm-server/src/test/java/functionaltests/RegressionTestSuite.java
fviale: It could be, but it is not so unstable, and leaving it as a standard test allows at least to test the restartDownNode policy outside of regression

scheduler/scheduler-server/src/test/java/functionaltests/RegressionTestSuite.java
fviale: I put them in the category "with custom configuration"

scheduler/scheduler-server/src/test/java/functionaltests/RegressionTestSuite.java
fviale: it's true that they are not so slow, but for me they are the definition of regression tests (i.e. sanity tests)

scheduler/scheduler-server/src/test/java/functionaltests/RegressionTestSuite.java
fviale: I agree in general, but TestJobRemove is the slowest test of the suite, so I would prefer to keep it in regression, SchedulerDbManagerConcurrencyTest is a sanity test and should be left in regression. TestJobAttributes & TestMultipleTasks can be moved to standard,

fviale:
It is true that SchedulerFunctionalTestWithRestart was used when sometimes it was not needed. But this was chosen at a time when the test suite was really unstable (never finishing, etc)

fviale:
Retest this please

fviale:
Retest this please

fviale:
Retest this please

fviale:
Retest this please

fviale:
Retest this pleas

fviale:
Retest this please

PR #3091, opened 2018-01-16
Increase performance test limits

Mykhailenko:
It seems my local machine is much faster than jenkins slaves. 
So I increased limits for performance tests.


Mykhailenko:
retest this please

PR #3126, opened 2018-02-08
Fix of LogsCheckingAfterRestartTest

Mykhailenko:
Workaround to fix LogsCheckingAfterRestartTest test. In this test we start, stop, and start server again.
When we perform second start, during initializing process RMCore retrieves LocalNodes nodesource from db. And after initialization we try to add another nodesource with the same name. So that is why I added isNodeSourceAlreadyExist method to RMCore, thus during server start-up we can check first if there is already nodesource with this name and if it exists we do not add new nodesource.

### DIFF
--- a/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
+++ b/rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
@@ -367,6 +367,11 @@ public class RMProxyUserInterface extends RMListenerProxy implements ResourceMan
     }
 
     @Override
+    public boolean isNodeSourceAlreadyExist(String nodeSourceName) {
+        return this.target.isNodeSourceAlreadyExist(nodeSourceName);
+    }
+
+    @Override
     public List<ScriptResult<Object>> executeScript(String script, String scriptEngine, String targetType,
             Set<String> targets) {
         return this.target.executeScript(script, scriptEngine, targetType, targets);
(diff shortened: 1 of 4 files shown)


rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
```diff
@@ -366,6 +366,11 @@ public NodeSet getNodes(Criteria criteria) {
         return target.getNodes(criteria);
     }
 
+    @Override
+    public boolean isNodeSourceAlreadyExist(String nodeSourceName) {
```
marcocast: isNodeSourceAlreadyExisting

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMProxyUserInterface.java
```diff
@@ -366,6 +366,11 @@ public NodeSet getNodes(Criteria criteria) {
...
+        return this.target.isNodeSourceAlreadyExist(nodeSourceName);
```
marcocast: isNodeSourceAlreadyExist

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/frontend/ResourceManager.java
```diff
@@ -350,6 +350,13 @@ NodeSet getNodes(int number, TopologyDescriptor descriptor, List<SelectionScript
      */
     NodeSet getNodes(Criteria criteria);
 
+    /**
+     *
+     * @param nodeSourceName
+     * @return true if RM already has node source with given name
+     */
+    boolean isNodeSourceAlreadyExisting(String nodeSourceName);
```
fviale: Better to use BooleanWrapper to let the caller decide to handle it synchronously or not.

Mykhailenko: yes right, make sense

PR #3135, opened 2018-02-14
Adding total number of nodes in the logs

marcocast:


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -859,7 +859,7 @@ private boolean isNumberOfNodesLimited() {
      *
      * @return total number of alive nodes
      */
-    private int getTotalAliveNodesNumber() {
+    public int getTotalAliveNodesNumber() {
```
fviale: Would it make sense to add this method to ResourceManager interface ?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
```diff
@@ -382,11 +382,11 @@ private NodeSet doSelectNodes(Criteria criteria, Client client) {
                                                                                                                     " extra nodes"
                                                                                                                   : "";
             logger.info(client + " requested " + criteria.getSize() + " nodes with " + criteria.getTopology() +
-                        " and will get " + selectedNodes.size() + " nodes " + extraNodes + " [freeNodes:" +
-                        freeNodes.size() + ";filteredNodes:" + filteredNodes.size() + ";reordered after policy:" +
-                        afterPolicyNodes.size() + ";selection script present:" + hasScripts +
-                        ";nodes filtered by selection script:" + matchedNodes.size() + ";selectedNodes:" +
-                        selectedNodes.size() + "]");
+                        " and will get " + selectedNodes.size() + " nodes " + extraNodes + " [totalNodes:" +
+                        rmcore.getTotalAliveNodesNumber() + ";freeNodes:" + freeNodes.size() + ";filteredNodes:" +
```
fviale: I would really prefer that you add the call to the RMCore in the doSelectNodes method, next to List<RMNode> freeNodes = rmcore.getFreeNodes();

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
marcocast: I will use listAliveNodeUrls() that already exists in the ResourceManager interface

marcocast:
retest this please

PR #3139, opened 2018-02-15
Avoid retrieve node source before persisting node

jrochas:
- prefer loading the node source in the database transaction to persist a node instead to retrieve it at the application level
- set a maximum number of bufferized database request so that the buffer does not grow too big in case there is continuous income of database requests.


jrochas:
Remember to test regression test as well for this PR, as it impacts mostly recovery.

jrochas:
retest this please

Issue #3136, opened 2018-02-14
Temporary workflow xml files are not deleted


fviale:
The issue occurs on Windows (to double check if it does not occur on linux also). But it is probably related to Windows file-locking semantics.

On a up-an-running scheduler, if a workflow is open in the studio, xml files are appearing in the TEMP folder, such as:
valid-job992783000741727014d
valid-job993147509328946495d

These files keep on appearing but are never deleted.

PR #3143, opened 2018-02-19
fix for #3136 Temporary workflow xml files are not deleted

fviale:
- rationalize all usages of FileInputStream/OutputStream/Reader/Writer with try-with-resources
 - ServerJobAndTaskLogs: delete temporary html file produced by WorkflowSubmitter when a job is deleted

### DIFF
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/utils/Zipper.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/utils/Zipper.java
@@ -55,7 +55,9 @@ public class Zipper {
         public static void zip(File file, OutputStream os) throws IOException {
             checkNotNull(file);
             checkNotNull(os);
-            GZIP.zip(new FileInputStream(file), os);
+            try (InputStream inputStream = new FileInputStream(file)) {
+                GZIP.zip(inputStream, os);
+            }
         }
 
         public static void zip(InputStream is, OutputStream os) throws IOException {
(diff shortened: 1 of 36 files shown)


fviale:
Retest this please

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/utils/Zipper.java
```diff
@@ -160,7 +170,9 @@ public static void unzip(InputStream is, File outFile) throws IOException {
                         entryContainer.mkdirs();
                     }
                     if (!entryFile.isDirectory()) {
-                        Zipper.ZIP.unzipEntry(zis, new FileOutputStream(entryFile));
+                        FileOutputStream outputStream = new FileOutputStream(entryFile);
```
marcocast: don't we need to close this outputStream ?

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/utils/Zipper.java
```diff
@@ -119,10 +125,14 @@ private static void zipFiles(List<File> files, String basepath, OutputStream os)
                 closer.register(zos);
                 if (files.size() == 1) {
                     File file = files.get(0);
-                    writeZipEntry(new ZipEntry(file.getName()), new FileInputStream(file), zos);
+                    FileInputStream inputStream = new FileInputStream(file);
```
marcocast: don't we need to close this inputStream ?

fviale: this is done by the "closer"

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/utils/Zipper.java
fviale: Same remark

PR #3145, opened 2018-02-20
Remove synchronization when task is started

jrochas:
- instead of waiting for the task to be started, we check at recovery time whether the doTask method of the task laucnher has already been called.

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
@@ -33,6 +33,7 @@ import java.security.PublicKey;
 import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.log4j.Logger;
 import org.objectweb.proactive.Body;
@@ -103,6 +104,8 @@ public class TaskLauncher implements InitActive {
 
     private TaskLauncherRebinder taskLauncherRebinder;
 
+    private AtomicBoolean doTaskInvoked = new AtomicBoolean(false);
+
     /**
      * Needed for ProActive but should never be used manually to create an instance of the object.
      */
@@ -142,6 +145,11 @@ public class TaskLauncher implements InitActive {
         return true;
     }
 
+    @ImmediateService
+    public boolean getDoTaskInvoked() {
+        return doTaskInvoked.get();
+    }
+
     void doTask(ExecutableContainer executableContainer, TaskResult[] previousTasksResults,
             TaskTerminateNotification terminateNotification) {
         doTask(executableContainer, previousTasksResults, terminateNotification, null, false);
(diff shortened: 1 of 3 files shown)


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
```diff
@@ -142,6 +145,11 @@ public boolean isActivated() {
         return true;
     }
 
+    @ImmediateService
+    public boolean getDoTaskInvoked() {
+        return doTaskInvoked.get();
```
fviale: isTaskStarted

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerStateRecoverHelper.java
```diff
@@ -192,9 +192,17 @@ private boolean runningTaskMustBeResetToPending(InternalTask task, RMProxy rmPro
...
+                logger.debug("Checking whether task launcher has called its doTask method: " +
+                             launcher.getDoTaskInvoked());
+                if (launcher.getDoTaskInvoked()) {
+                    logger.info("Recover running task " + task.getId() + " (" + task.getName() +
+                                ") successfully with task launcher " + launcher);
+                    resetToPending = false;
+                } else {
+                    logger.info(FAIL_TO_RECOVER_RUNNING_TASK_STRING + task.getId() + " (" + task.getName() +
+                                ") because its task launcher has not started to execute the task");
+                    resetToPending = true;
```
fviale: I suppose we need to do something with the task launcher. Unless the rm takes care of (kill) it when we release the node.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
```diff
@@ -103,6 +104,8 @@
 
     private TaskLauncherRebinder taskLauncherRebinder;
 
+    private AtomicBoolean doTaskInvoked = new AtomicBoolean(false);
```
marcocast: isTaskStarted

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerStateRecoverHelper.java
jrochas: Ok, makes sense to initiate a kill here. But after looking at it, it seems that TaskLauncher#kill is not what we want. I would instead make TaskLauncher#terminate public and call it to stop the task launcher activity. How does it sound?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
jrochas: done

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
jrochas: done

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerStateRecoverHelper.java
fviale: I think the nodes must be released to the RM, they will be busy in that case (and stay so if nothing is done)

jrochas: ok, after the recovery, there is a mechanism that releases the busy nodes that have no running task. What you describe falls in this specific case, so there is no need to add a killing mechanism at this point.

PR #3152, opened 2018-02-28
Add diagnostic messages to unstable TestJobServerLogs

Mykhailenko:
TestJobServerLogs fails because there are logs dedicated for the job in the logs folder after we removed job. It does not happen all the time. I guess it happens because after we remove files, some task writes to to file. I do not want to guess what is wrong right now. I added diagnostic messages, which will print first 5 and last 5 lines of EVERY files in the /logs folder. So we will see who prints what.


Mykhailenko:
retest this please

Mykhailenko:
retest this please

Mykhailenko:
retest this please

PR #3165, opened 2018-03-14
Split TestJobLegacySchemas test 

Mykhailenko:
and add diagnostic messages in case if job is not finished after 700s timeout.

### DIFF
--- a/scheduler/scheduler-server/src/test/java/functionaltests/monitor/SchedulerMonitorsHandler.java
+++ b/scheduler/scheduler-server/src/test/java/functionaltests/monitor/SchedulerMonitorsHandler.java
@@ -29,7 +29,10 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Vector;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.objectweb.proactive.core.ProActiveTimeoutException;
 import org.objectweb.proactive.utils.TimeoutAccounter;
@@ -300,6 +303,15 @@ public class SchedulerMonitorsHandler {
             return jobsEvents.get(id).remove((jobsEvents.get(id).indexOf(tmp)));
         } else
             return null;
+    }
+
+    public List<JobEventMonitor> getJobEvents(JobId id) {
+        return jobsEvents.entrySet()
+                         .stream()
+                         .filter(entry -> entry.getKey().equals(id))
+                         .map(entry -> entry.getValue())
+                         .findFirst()
+                         .orElse(new ArrayList<>());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


scheduler/scheduler-server/src/test/java/functionaltests/monitor/SchedulerMonitorsHandler.java
```diff
@@ -298,13 +297,21 @@ private void addTaskEvent(SchedulerEvent event, TaskInfo taskInfo) {
      * @param event type to look for
      * @return removed JobEventMonitor, or null if not exists.
      */
-    public Optional<JobEventMonitor> removeJobEvent(JobId id, SchedulerEvent event) {
+    private JobEventMonitor removeJobEvent(JobId id, SchedulerEvent event) {
```
MaelAudren: Why removing optional ? I think it is more clear that the method could return null if in the prototype there is an optional.

scheduler/scheduler-server/src/test/java/functionaltests/utils/SchedulerTHelper.java
```diff
@@ -752,46 +751,41 @@ public JobInfo waitForEventJobFinished(JobId id) throws Exception {
...
-                    return opJobMonitor.get().getJobInfo();
-                }
-
-                monitorRunningToFinished = (JobEventMonitor) getSchedulerMonitorsHandler().getMonitor(new JobEventMonitor(SchedulerEvent.JOB_RUNNING_TO_FINISHED,
-                                                                                                                          id));
-                monitorPendingToFinished = (JobEventMonitor) getSchedulerMonitorsHandler().getMonitor(new JobEventMonitor(SchedulerEvent.JOB_PENDING_TO_FINISHED,
-                                                                                                                          id));
+            return waitForJobEvent(userInterface,
+                                   id,
+                                   700000, // 700s to wait, when overall junit timeout is 800s
```
fviale: could you reuse CentralPAPropertyRepository.PA_TEST_TIMEOUT.getValue() instead of a hardcoded value ?

scheduler/scheduler-server/src/test/java/functionaltests/monitor/SchedulerMonitorsHandler.java
Mykhailenko: Actually I reverted back my commit where I also added Optional...

scheduler/scheduler-server/src/test/java/functionaltests/utils/SchedulerTHelper.java
```diff
@@ -752,46 +752,41 @@ public JobInfo waitForEventJobFinished(JobId id) throws Exception {
...
-                    return opJobMonitor.get().getJobInfo();
-                }
-
-                monitorRunningToFinished = (JobEventMonitor) getSchedulerMonitorsHandler().getMonitor(new JobEventMonitor(SchedulerEvent.JOB_RUNNING_TO_FINISHED,
-                                                                                                                          id));
-                monitorPendingToFinished = (JobEventMonitor) getSchedulerMonitorsHandler().getMonitor(new JobEventMonitor(SchedulerEvent.JOB_PENDING_TO_FINISHED,
-                                                                                                                          id));
+            return waitForJobEvent(userInterface,
+                                   id,
+                                   CentralPAPropertyRepository.PA_TEST_TIMEOUT.getValue() - 100000, // the idea is to put smaller timeout than existing timeout for tests
```
fviale: 85 % ?

Mykhailenko: ok

PR #3175, opened 2018-03-27
Add SFTP-Connector to proactive examples

amouhoub:
In this PR, we add the SFTP connector java class to scheduler-examples in order to be used for the sftp-connector task in data-connectors bucket (proactive-examples).

### DIFF
--- a/scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
+++ b/scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
@@ -62,6 +62,11 @@ public class SFTPConnector extends JavaExecutable {
 
     private static final int DEFAULT_SFTP_PORT = 22;
 
+    /**
+     * The SFTP_URL_KEY = "sftp://<username>@<host>" is used as a key to store sftp passwords in 3rd party credentials
+     */
+    private static final String SFTP_URL_KEY = "sftp://<username>@<host>";
+
     private static final String GET = "GET";
 
     private static final String PUT = "PUT";
@@ -133,13 +138,13 @@ public class SFTPConnector extends JavaExecutable {
         }
 
         // This key is used for logs and for getting the password from 3rd party credentials.
-        sftpUrlKey = "sftp://" + sftpUsername + "@" + sftpHostname + ":" + sftpPort;
+        sftpUrlKey = "sftp://" + sftpUsername + "@" + sftpHostname;
 
         sftpPassword = getThirdPartyCredential(sftpUrlKey);
 
         if (sftpPassword == null) {
             throw new IllegalArgumentException("Please add your sftp password to 3rd-party credentials under the key :\"" +
-                                               sftpUrlKey + "\"");
+                                               SFTP_URL_KEY + "\"");
         }
     }
 
@@ -243,7 +248,7 @@ public class SFTPConnector extends JavaExecutable {
 
             File[] files = sourceFile.listFiles();
 
-            if (files != null && !sourceFile.isHidden()) {
+            if (!sourceFile.isHidden()) {
 
                 // create a directory if it does not exist
                 if (!remoteDirExists(channelSftp, remoteFilePath)) {
@@ -294,7 +299,7 @@ public class SFTPConnector extends JavaExecutable {
 
     /**
      * This file downloads recursively a folder and all its contents over SFTP. <br>
-     * By default, this method <b>OVERWRITES </b>existing files and <b>APPENDS</b> existing folders.
+     * By default, this method <b>OVERWRITES</b> existing files and <b>APPENDS</b> existing folders.
      *
      * @param channelSftp
      * @param sftpRemoteRelativePath
@@ -307,7 +312,16 @@ public class SFTPConnector extends JavaExecutable {
             String sftpLocalRelativePath) throws SftpException, IOException {
 
         List<String> filesRelativePathName = new ArrayList<>();
-        List<ChannelSftp.LsEntry> fileAndFolderList = channelSftp.ls(sftpRemoteRelativePath);
+
+        // Do an ls on the remote path (sftpRemoteRelativePath)
+        List<ChannelSftp.LsEntry> fileAndFolderList = null;
+        try {
+            fileAndFolderList = channelSftp.ls(sftpRemoteRelativePath);
+        } catch (SftpException e) {
+            throw new SftpException(e.id,
+                                    "File not found or permission denied at " + sftpRemoteRelativePath,
+                                    e.getCause());
+        }
 
         // Check if the LS results point to a file not a directory
         if (isRemoteFile(fileAndFolderList)) {
@@ -317,7 +331,7 @@ public class SFTPConnector extends JavaExecutable {
             // Download single file
             filesRelativePathName.add(downloadSingleFile(channelSftp, sftpRemoteRelativePath, localFilePath));
 
-            // if the file is a zip and sftpextractarchive == true then extract the archive
+            // if the file is a zip and sftpExtractArchive == true then extract the archive
             if (sftpExtractArchive && sftpRemoteRelativePath.endsWith(".zip")) {
                 extractArchive(localFilePath, sftpLocalRelativePath);
             }
@@ -325,7 +339,7 @@ public class SFTPConnector extends JavaExecutable {
             return filesRelativePathName;
         }
 
-        //Iterate through list of folder content
+        // Iterate through list of folder content
         for (ChannelSftp.LsEntry item : fileAndFolderList) {
 
             String remoteFilePath = Paths.get(sftpRemoteRelativePath, item.getFilename()).toString();
@@ -379,7 +393,7 @@ public class SFTPConnector extends JavaExecutable {
      *
      * @param channelSftp
      * @param path
-     * @throws SftpException
+     * @throws SftpException if the remote folder(s) cannot be created
      */
     private void makeRemoteDirectories(ChannelSftp channelSftp, Path path) throws SftpException {
         if (!remoteDirExists(channelSftp, path.toString())) {
@@ -392,6 +406,11 @@ public class SFTPConnector extends JavaExecutable {
         }
     }
 
+    /**
+     * This method extracts a zip file located at localFilePath and saves the unpacked files to sftpLocalRelativePath.
+     * @param localFilePath
+     * @param sftpLocalRelativePath
+     */
     private void extractArchive(String localFilePath, String sftpLocalRelativePath) {
         getOut().println("Decompressing archive: " + localFilePath);
         ZipUtil.unpack(new File(localFilePath), new File(sftpLocalRelativePath));
--- a/scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
+++ b/scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
@@ -64,6 +64,16 @@ public class SFTPConnector extends JavaExecutable {
 
     private int sftpPort = DEFAULT_SFTP_PORT;
 
+    private static final String GET = "GET";
+
+    private static final String PUT = "PUT";
+
+    private static final String STRICT_HOST_KEY_CHEKING = "StrictHostKeyChecking";
+
+    private static final String CURRENT_FOLDER = ".";
+
+    private static final String PARENT_FOLDER = "..";
+
     private boolean sftpExtractArchive;
 
     private String sftpLocalRelativePath;
@@ -99,7 +109,7 @@ public class SFTPConnector extends JavaExecutable {
             sftpLocalRelativePath = args.get(SFTP_LOCAL_RELATIVE_PATH_ARG).toString();
         } else {
             //we throw an exception tp prevent transferring all the contents of the global space.
-            if (sftpMode.equals("PUT")) {
+            if (sftpMode.equals(PUT)) {
                 throw new IllegalArgumentException("Please specify a local relative path. Empty value is not allowed.");
             }
             //default value is getlocalspace() because it will always be writable and moreover can be used to transfer files to another data space(global, user)
@@ -142,7 +152,7 @@ public class SFTPConnector extends JavaExecutable {
             session = jsch.getSession(sftpUsername, sftpHostname, sftpPort);
             session.setPassword(sftpPassword);
             Properties config = new Properties();
-            config.put("StrictHostKeyChecking", "no");
+            config.put(STRICT_HOST_KEY_CHEKING, "no");
             session.setConfig(config);
             // create sftp Session
             getOut().println("Connecting to " + sftpUrlKey);
@@ -154,7 +164,7 @@ public class SFTPConnector extends JavaExecutable {
 
             switch (sftpMode) {
                 //ftp mode is get
-                case "GET":
+                case GET:
                     getOut().println("BEGIN DOWNLOAD FROM SFTP SERVER: " + sftpUrlKey);
                     // recursive folder content download from sftp server
                     filesRelativePathName = recursiveFolderDownload(channelsftp,
@@ -164,7 +174,7 @@ public class SFTPConnector extends JavaExecutable {
                     break;
 
                 //ftp mode is put
-                case "PUT":
+                case PUT:
                     getOut().println("BEGIN UPLOAD TO SFTP SERVER: " + sftpUrlKey);
                     // Create remote directory(ies) if it(they) does(do) not exist
                     if (!remoteDirExists(channelsftp, sftpRemoteRelativePath)) {
@@ -178,7 +188,7 @@ public class SFTPConnector extends JavaExecutable {
                     break;
 
                 default:
-                    throw new IllegalArgumentException("SFTP mode can only be PUT or GET.");
+                    throw new IllegalArgumentException("SFTP mode can only be " + PUT + " or " + GET);
             }
         } catch (JSchException e) {
             throw new JSchException("An error occured while trying to connect to SFTP server: ", e);
@@ -301,9 +311,7 @@ public class SFTPConnector extends JavaExecutable {
 
             // if the file is a zip and sftpextractarchive == true then extract the archive
             if (sftpExtractArchive && sftpRemoteRelativePath.endsWith(".zip")) {
-                getOut().println("Decompressing archive: " + localFilePath);
-                ZipUtil.unpack(new File(localFilePath), new File(sftpLocalRelativePath));
-                getOut().println("Archive decompressed successfully at: " + sftpLocalRelativePath);
+                extractArchive(localFilePath, sftpLocalRelativePath);
             }
 
             return filesRelativePathName;
@@ -321,7 +329,7 @@ public class SFTPConnector extends JavaExecutable {
                 filesRelativePathName.add(downloadSingleFile(channelSftp, remoteFilePath, localFilePath));
 
             } else // skip '.' and '..'
-            if (!(".".equals(item.getFilename()) || "..".equals(item.getFilename()))) {
+            if (!(CURRENT_FOLDER.equals(item.getFilename()) || PARENT_FOLDER.equals(item.getFilename()))) {
                 // Copy empty folder.
                 localFile.mkdirs();
                 // Enter found folder on server to read its contents and create locally.
@@ -374,5 +382,11 @@ public class SFTPConnector extends JavaExecutable {
                 channelSftp.mkdir(path.toString());
             }
         }
+    }
+
+    private void extractArchive(String localFilePath, String sftpLocalRelativePath) {
+        getOut().println("Decompressing archive: " + localFilePath);
+        ZipUtil.unpack(new File(localFilePath), new File(sftpLocalRelativePath));
+        getOut().println("Archive decompressed successfully at: " + sftpLocalRelativePath);
     }
 }
--- a/scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
+++ b/scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
@@ -96,7 +96,11 @@ public class SFTPConnector extends JavaExecutable {
             sftpHostname = args.get("sftpHostname").toString();
         }
         if (args.containsKey("sftpPort")) {
-            sftpPort = Integer.parseInt(args.get("sftpPort").toString());
+            try {
+                sftpPort = Integer.parseInt(args.get("sftpPort").toString());
+            } catch (NumberFormatException e) {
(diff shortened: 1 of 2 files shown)


scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,366 @@
...
+            getOut().println("File uploaded successfully to " +
+                             Paths.get(destinationPath, sourceFile.getName()).toString());
+            filesRelativePathName.add(Paths.get(destinationPath, sourceFile.getName()).toString());
+
+        } else {
+
+            String remoteFilePath = Paths.get(destinationPath, sourceFile.getName()).toString();
+            File[] files = sourceFile.listFiles();
+
+            if (files != null && !sourceFile.getName().startsWith(".")) {
```
fviale: maybe consider isHidden() ?

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,366 @@
...
+
+                channelSftp.cd(destinationPath);
```
fviale: this looks very strange, and looks like a bug, unless destinationPath is always absolute ! As the channelSftp is passed recursively, it has a memory of the remote current folder.

What happens when you want to transfer:
a/b/file_c
d/file_e
?

To me, it will do:
mkdir a, cd a, mkdir b, cd b, transfer file_c
mkdir d, cd d,  transfer file_e
And this will result in the structure:
a/b/file_c
a/b/d/file_e

instead of what is needed.

Maybe this code works, it is hard to tell, because the distinction relative/absolute is not clear enough

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,366 @@
...
+            }
+            //default value is getlocalspace() because it will always be writable and moreover can be used to transfer files to another data space(global, user)
+            sftpLocalRelativePath = getLocalSpace();
+        }
+        if (args.containsKey("sftpRemoteRelativePath")) {
+            sftpRemoteRelativePath = args.get("sftpRemoteRelativePath").toString();
+        }
+
+        sftpUsername = getThirdPartyCredential("SFTP_USERNAME");
+        sftpPassword = getThirdPartyCredential("SFTP_PASSWORD");
```
fviale: I am thinking that maybe there is a better way to store username/password combination. Currently, it is only possible to interact with one sftp server for a given scheduler user.

The user name could be passed as a task argument, and the password could be stored as:
key:
sftp://sftpHostname:sftpPort 
value:
password

The same comment applies for the FTPConnector

fviale: sorry, I meant:
key:
username@sftp://sftpHostname:sftpPort

amouhoub: This is a good point. In fact, we have the same problem for all data connectors (you can have only one connection per workflow unless you manually add prefixes to credentials variables as well as to Hostname, etc...) This would be a great [temporary?] fix for all data-connectors.

Btw, this is not related, but we kind of (KIND OF) serve the same purpose with the URL-connector task wich in turn calls the sftp workflow from the catalog after extracting all the stuff from the url. However, we don't properly deal yet with the password, but at least you can have as many sftp/ftp connections you want with URL tasks without risking any collision.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: Yes, besides, that would also work for Windows-hidden files that are not prefixed with '.' (in case of sftp server on windows using OpenSSH). 
Thank you @fviale

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: In fact, it seems a bit confusing, but it works perfectly either with relative or absolute paths (see the attached screenshot).

folder-structure

amouhoub:
@fviale can you have a second look/approve this PR please ? thank you in advance

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+
+    private static final String SFTP_LOCAL_RELATIVE_PATH_ARG = "sftpLocalRelativePath";
+
+    @Override
+    public void init(Map<String, Serializable> args) throws Exception {
+        if (args.containsKey("sftpHostname")) {
+            sftpHostname = args.get("sftpHostname").toString();
+        }
+        if (args.containsKey("sftpPort")) {
+            sftpPort = Integer.parseInt(args.get("sftpPort").toString());
```
jrochas: You could catch a possible `NumberFormatException` here and throw a more explicit `IllegalArgumentException` that chains the `NumberFormatException`.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+
+        ChannelSftp channelsftp = null;
+        Session session = null;
+        Channel channel = null;
+
+        try {
+            JSch jsch = new JSch();
+            session = jsch.getSession(sftpUsername, sftpHostname, sftpPort);
+            session.setPassword(sftpPassword);
+            Properties config = new Properties();
```
jrochas: You could extract the three lines of SSH configuration in a dedicated private method (in case we need to add more config for exmaple).

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+            config.put("StrictHostKeyChecking", "no");
+            session.setConfig(config);
+            // create sftp Session
+            getOut().println("Connecting to " + sftpUrlKey);
+            session.connect();
+            // open sftp Channel
+            channel = session.openChannel("sftp");
```
jrochas: Can you create a constant for the "sftp" String?

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+        }
+        if (args.containsKey("sftpExtractArchive")) {
+            sftpExtractArchive = Boolean.parseBoolean(args.get("sftpExtractArchive").toString());
+        }
+        if (args.containsKey(SFTP_LOCAL_RELATIVE_PATH_ARG) &&
+            !args.get(SFTP_LOCAL_RELATIVE_PATH_ARG).toString().isEmpty()) {
+            sftpLocalRelativePath = args.get(SFTP_LOCAL_RELATIVE_PATH_ARG).toString();
+        } else {
+            //we throw an exception tp prevent transferring all the contents of the global space.
+            if (sftpMode.equals("PUT")) {
```
jrochas: Can you create a constant for the "PUT" String?

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+                    getOut().println("BEGIN DOWNLOAD FROM SFTP SERVER: " + sftpUrlKey);
+                    // recursive folder content download from sftp server
+                    filesRelativePathName = recursiveFolderDownload(channelsftp,
+                                                                    sftpRemoteRelativePath,
+                                                                    sftpLocalRelativePath);
+                    getOut().println("END DOWNLOAD FROM SFTP");
+                    break;
+
+                //ftp mode is put
+                case "PUT":
```
jrochas: We can use the constant here instead of "PUT".

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+            getOut().println("Connecting to " + sftpUrlKey);
+            session.connect();
+            // open sftp Channel
+            channel = session.openChannel("sftp");
+            channel.connect();
+            channelsftp = (ChannelSftp) channel;
+
+            switch (sftpMode) {
+                //ftp mode is get
+                case "GET":
```
jrochas: Can you create a constant for the "GET" String?

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+                    }
+                    // recursive folder content upload to sftp server
+                    filesRelativePathName = recursiveFolderUpload(channelsftp,
+                                                                  sftpLocalRelativePath,
+                                                                  sftpRemoteRelativePath);
+                    getOut().println("END UPLOAD TO SFTP");
+                    break;
+
+                default:
+                    throw new IllegalArgumentException("SFTP mode can only be PUT or GET.");
```
jrochas: You can replace SFTP, PUT and GET in the exception message by the constants.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+        if (sourceFile.isFile()) {
+            // copy if it is a file
+            getOut().println("Uploading " + sourcePath);
+            channelSftp.put(new FileInputStream(sourceFile), remoteFilePath, ChannelSftp.OVERWRITE);
+            getOut().println("File uploaded successfully to " + remoteFilePath);
+            filesRelativePathName.add(remoteFilePath);
+
+        } else {
+
+            File[] files = sourceFile.listFiles();
```
jrochas: If you'd like, you can rewrite the following in Java 8 flavor, e.g. `Files.list(Paths.get(sourceFile)).forEach(...);`

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+        if (isRemoteFile(fileAndFolderList)) {
+
+            String localFilePath = Paths.get(sftpLocalRelativePath, fileAndFolderList.get(0).getFilename()).toString();
+
+            // Download single file
+            filesRelativePathName.add(downloadSingleFile(channelSftp, sftpRemoteRelativePath, localFilePath));
+
+            // if the file is a zip and sftpextractarchive == true then extract the archive
+            if (sftpExtractArchive && sftpRemoteRelativePath.endsWith(".zip")) {
+                getOut().println("Decompressing archive: " + localFilePath);
```
jrochas: You can move the three lines in a private method `extractArchive`.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+     * This method downloads a single file over SFTP.<br>
+     * By default, this method overwrites existing files.
+     *
+     * @param channelSftp
+     * @param remoteFilePath
+     * @param localFilePath
+     * @return
+     * @throws SftpException
+     */
+    private String downloadSingleFile(ChannelSftp channelSftp, String remoteFilePath, String localFilePath)
```
jrochas: To reduce the scope of this class (that could be to only handle the SFTP connection), I suggest you could create two utility classes next to `SFTPConnection`. For example `SFTPConnectorDownloader` and `SFTPConnectorUploader`.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+            String remoteFilePath = Paths.get(sftpRemoteRelativePath, item.getFilename()).toString();
+            String localFilePath = Paths.get(sftpLocalRelativePath, item.getFilename()).toString();
+            File localFile = new File(localFilePath);
+
+            // Check if it is a file (not a directory).
+            if (!item.getAttrs().isDir()) {
+                filesRelativePathName.add(downloadSingleFile(channelSftp, remoteFilePath, localFilePath));
+
+            } else // skip '.' and '..'
+            if (!(".".equals(item.getFilename()) || "..".equals(item.getFilename()))) {
```
jrochas: You could have constants `CURRENT_FOLDER` and `PARENT_FOLDER` for these.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+
+    /**
+     * This method recursively creates nonexisting nested folders
+     * Example /folder/sub1/sub2/sub3/sub...
+     *
+     * @param channelSftp
+     * @param path
+     * @throws SftpException
+     */
+    private void makeRemoteDirectories(ChannelSftp channelSftp, Path path) throws SftpException {
```
jrochas: Maybe change name to `createRemoteDirectories`?

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: Normally this should be prevented in the workflow by specifying the PA:Integer model for the port variable. But yes, that can't be always enough

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: This one is used only once, and it cannot take another value (exec, shell, etc) , otherwise the code will not work, so we hardcode this on purpose.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: Agree

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: Agree

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: Agree as well

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,378 @@
...
+            throws SftpException, FileNotFoundException {
+
+        List<String> filesRelativePathName = new ArrayList<>();
+        File sourceFile = new File(sourcePath);
+        String remoteFilePath = Paths.get(destinationPath, sourceFile.getName()).toString();
+
+        if (sourceFile.isFile()) {
+            // copy if it is a file
+            getOut().println("Uploading " + sourcePath);
+            channelSftp.put(new FileInputStream(sourceFile), remoteFilePath, ChannelSftp.OVERWRITE);
```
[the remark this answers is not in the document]
amouhoub: Well, we thought of this, but then we decided to do this later because we have other connectors and we want to homogenize some common parameters including this one.
The idea is to have a boolean `Overwrite` variables in all connectors that fit.
So it's a do-it-later one.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: In fact, we prefer "make" than "create" because it is closer to mkdir (make dir).

amouhoub:
thank you @jrochas for your in-depth reviews. We took some in consideration and left some for another iteration to synchronize with other data-connectors.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,396 @@
...
+        if (args.containsKey(SFTP_LOCAL_RELATIVE_PATH_ARG) &&
+            !args.get(SFTP_LOCAL_RELATIVE_PATH_ARG).toString().isEmpty()) {
+            sftpLocalRelativePath = args.get(SFTP_LOCAL_RELATIVE_PATH_ARG).toString();
+        } else {
+            //we throw an exception tp prevent transferring all the contents of the global space.
+            if (sftpMode.equals(PUT)) {
+                throw new IllegalArgumentException("Please specify a local relative path. Empty value is not allowed.");
+            }
+            //default value is getlocalspace() because it will always be writable and moreover can be used to transfer files to another data space(global, user)
+            sftpLocalRelativePath = getLocalSpace();
```
fviale: getLocalSpace() is an absolute path.

instead of using this as the default sftpLocalRelativePath, you should define a sftpLocalAbsolutePath:

- if sftpLocalRelativePath is empty, then sftpLocalAbsolutePath = getLocalSpace()
- if sftpLocalRelativePath is not empty, then sftpLocalAbsolutePath = new File(getLocalSpace(), sftpLocalRelativePath)

This is the only way to guaranty coherence between fork or non-fork mode. In the latter case, current directory (.) is not writeable, but getLocalSpace() is always writeable.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,396 @@
...
+            throw new SftpException(e.id,
+                                    "File not found, permission denied or operation not supported by SFTP server.",
+                                    e);
+        } finally {
+            if (channelsftp != null)
+                channelsftp.disconnect();
+            if (channel != null)
+                channel.disconnect();
+            if (session != null)
+                session.disconnect();
```
Mykhailenko: [Java Code Convention](http://www.oracle.com/technetwork/java/codeconventions-150003.pdf) says: `if statements always use braces {}`

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,396 @@
...
+                                                                  sftpLocalRelativePath,
+                                                                  sftpRemoteRelativePath);
+                    getOut().println("END UPLOAD TO SFTP");
+                    break;
+
+                default:
+                    throw new IllegalArgumentException("SFTP mode can only be " + PUT + " or " + GET);
+            }
+        } catch (JSchException e) {
+            throw new JSchException("An error occured while trying to connect to SFTP server: ", e);
```
Mykhailenko: Why do you wrap JSchException in JSchException?

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,396 @@
...
+
+    private String sftpMode;
+
+    private String sftpUsername;
+
+    private String sftpUrlKey;
+
+    private String sftpPassword;
+
+    private static final String SFTP_LOCAL_RELATIVE_PATH_ARG = "sftpLocalRelativePath";
```
Mykhailenko: Please, order these fields, e.g. first static ones, than instance ones, thanks

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: Just to emit a custom and understandable error message, and forward the stacktrace up...

Mykhailenko: then maybe: `throw new JSchException("An error occured while trying to connect to SFTP server: ", e.getCause());`

amouhoub:
thank you @Mykhailenko. Your suggestions have been included...

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
amouhoub: Makes sense. Done

amouhoub:
@Mykhailenko can you approve it plz ?

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SFTPConnector.java
```diff
@@ -0,0 +1,400 @@
...
+        if (args.containsKey("sftpUsername")) {
+            sftpUsername = args.get("sftpUsername").toString();
+        }
+
+        if (sftpUsername == null) {
+            throw new IllegalArgumentException("Username is required to access an SFTP server");
+        }
+
+        // This key is used for logs and for getting the password from 3rd party credentials.
+        sftpUrlKey = "sftp://" + sftpUsername + "@" + sftpHostname + ":" + sftpPort;
```
fviale: I really think the port is not necessary for the credentials key. You can create a different variable to store the url

amouhoub:
retest this please

PR #3185, opened 2018-04-12
prevent the submission of nameless job

Aminelouati:


paraita:
Good. One question though, how does it behave with the job name " " ? Is that a valid job name ?

amouhoub:
@paraita it takes a freaky mind to think about freaky situations !!!
What about if the name is "THE BEST CRAZY WORKFLOW EVER" ?

Aminelouati:
@paraita it is a bug :D

paraita:
the `" "` case don't necessarily need to be included in this PR, we can put it as an inprovement in the issues tab anyway :)

PR #3213, opened 2018-05-09
Fix deadlock in LocalInfrastrucure by writing protected variables in separate threads only.

vinseon:
This also makes the infrastructure more clean and robust:
 - Do not acquire more than maxNodes nodes
 - Do not manually check the status of deployment processes
 - Delete processes when removing latest node
 - Use concurrent hashmap to safely manipulate 'nodes to processes' mapping
 - Decrement all appropriate counters when deleting a node
 - Do not mess with down nodes detection: do not touch acquiredNode counter!
 - Reorder, simplify and add some wrapper methods
 - Remove unused methods and dead code

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
@@ -365,26 +365,26 @@ public class LocalInfrastructure extends InfrastructureManager {
     }
 
     private void increaseNumberOfHandledNodesWithLockAndPersist(final int additionalNumberOfNodes) {
-        setPersistedInfraVariable((PersistedInfraVariablesHandler<Void>) () -> {
+        setPersistedInfraVariable(() -> {
             int updated = (int) this.persistedInfraVariables.get(NB_HANDLED_NODES_KEY) + additionalNumberOfNodes;
             this.persistedInfraVariables.put(NB_HANDLED_NODES_KEY, updated);
-            return null;
+            return updated;
         });
     }
 
     private void incrementNumberOfAcquiredNodesWithLockAndPersist() {
-        setPersistedInfraVariable((PersistedInfraVariablesHandler<Void>) () -> {
+        setPersistedInfraVariable(() -> {
             int updated = (int) this.persistedInfraVariables.get(NB_ACQUIRED_NODES_KEY) + 1;
             this.persistedInfraVariables.put(NB_ACQUIRED_NODES_KEY, updated);
-            return null;
+            return updated;
         });
     }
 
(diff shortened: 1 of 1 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
```diff
@@ -389,22 +327,49 @@ protected void initializePersistedInfraVariables() {
...
         return getPersistedInfraVariable(() -> (int) this.persistedInfraVariables.get(NB_ACQUIRED_NODES_KEY));
     }
 
-    private int getNumberOfLostNodesWithLock() {
-        return getPersistedInfraVariable(() -> (int) this.persistedInfraVariables.get(NB_LOST_NODES_KEY));
+    private void increaseNumberOfHandledNodesWithLockAndPersist(final int additionalNumberOfNodes) {
+        setPersistedInfraVariable((PersistedInfraVariablesHandler<Void>) () -> {
+            int updated = (int) this.persistedInfraVariables.get(NB_HANDLED_NODES_KEY) + additionalNumberOfNodes;
+            this.persistedInfraVariables.put(NB_HANDLED_NODES_KEY, updated);
+            return null;
```
Mykhailenko: is it fine that you return null?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
```diff
@@ -119,19 +120,23 @@ public void acquireAllNodes() {
...
-        startNodes(1);
+        if (maxNodes - getNumberOfAcquiredNodesWithLock() > 0) {
+            startNodes(1);
+        }
     }
 
     @Override
     public void acquireNodes(int n, Map<String, ?> nodeConfiguration) {
-        if (n > 0) {
+        if (n > 0 && (maxNodes - getNumberOfAcquiredNodesWithLock() - n) >= 0) {
```
Mykhailenko: this is barely readable, what is n?

vinseon: OK changed in 38b2ed2f5df5328736c051dcc52116bc112cf27d, BTW this is the actual signature of the overrided NodeSource method.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
vinseon: Yes as we don't use the return value, see `PersistedInfraVariablesHandler<Void>`. Changed in fe1041da4cb996c6841786fc000dfb120c836238 as it is confusing.

Issue #3218, opened 2018-05-15
Regression in some SSH related tests


paraita:
Due to recent [changes to the JDK 8](http://www.oracle.com/technetwork/java/javase/8u151-relnotes-3850493.html), our SSH related tests (e.g TestSSHInfrastructureV2) are failing on the following error:

```
java.security.InvalidKeyException: The security strength of SHA-1 digest algorithm is not sufficient for this key size
	at sun.security.provider.DSA.checkKey(DSA.java:111)
	at sun.security.provider.DSA.engineInitSign(DSA.java:143)
	at java.security.Signature$Delegate.engineInitSign(Signature.java:1177)
```

As a result, this introduce instability to our CI pipeline.

PR #3219, opened 2018-05-15
Use RSA instead of DSA to prevent errors on recent JDK updates

paraita:
Will stabilize SSH functional tests described in #3218

### DIFF
--- a/rm/rm-server/src/test/java/functionaltests/nodesource/TestSSHInfrastructureV2.java
+++ b/rm/rm-server/src/test/java/functionaltests/nodesource/TestSSHInfrastructureV2.java
@@ -122,7 +122,9 @@ public class TestSSHInfrastructureV2 extends RMFunctionalTest {
 
         sshd = SshServer.setUpDefaultServer();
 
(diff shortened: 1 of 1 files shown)


rm/rm-server/src/test/java/functionaltests/nodesource/TestSSHInfrastructureV2.java
```diff
@@ -122,7 +122,9 @@ public static void startSSHServer() throws Exception {
 
         sshd = SshServer.setUpDefaultServer();
 
-        sshd.setKeyPairProvider(new SimpleGeneratorHostKeyProvider());
+        SimpleGeneratorHostKeyProvider keyProvider = new SimpleGeneratorHostKeyProvider();
+        keyProvider.setAlgorithm("RSA");
```
gheon: RSA-??? . I'm just curious to know whether or not you can specify the number of bits.

paraita: yep it's possible !

PR #3260, opened 2018-07-17
SynchronizationWrapper: lookup synchronization service only when needed

fviale:
This is to prevent the start of a ProActive Runtime on each task execution, even though the synchronization API is not used inside the task.

 With this change, the synchronization service stub is transient. It will be looked up on the worker node only when any method of the api is called.

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/synchronization/SynchronizationWrapper.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/synchronization/SynchronizationWrapper.java
@@ -27,11 +27,15 @@ package org.ow2.proactive.scheduler.synchronization;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeoutException;
 
+import org.apache.log4j.Logger;
+import org.objectweb.proactive.api.PAActiveObject;
+import org.objectweb.proactive.core.mop.MOP;
 import org.ow2.proactive.scheduler.common.task.TaskId;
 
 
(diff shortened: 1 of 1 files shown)


fviale:
Retest this please

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/synchronization/SynchronizationWrapper.java
```diff
@@ -27,11 +27,15 @@
...
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeoutException;
 
+import org.apache.log4j.Logger;
+import org.objectweb.proactive.ActiveObjectCreationException;
```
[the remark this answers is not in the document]
marcocast: as sonar said

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/synchronization/SynchronizationWrapper.java
```diff
@@ -53,170 +58,221 @@
...
         this.taskId = taskId;
         this.internalAPI = internalAPI;
+        if (internalAPI != null && MOP.isReifiedObject(internalAPI)) {
+            internalAPIUrls = PAActiveObject.getUrls(internalAPI);
+        }
+    }
+
+    private void initInternalAPI() {
+        if (this.internalAPI == null && this.internalAPIUrls != null) {
+            for (String lookupURL : this.internalAPIUrls) {
```
marcocast: should this block be synchronised to avoid 2 calls creating 2 times the object?

fviale: good point !

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/synchronization/SynchronizationWrapper.java
```diff
@@ -53,170 +58,221 @@
 
     TaskId taskId;
 
-    SynchronizationInternal internalAPI;
+    transient SynchronizationInternal internalAPI;
+
+    String[] internalAPIUrls;
```
Mykhailenko: Should not all fields be private?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/synchronization/SynchronizationWrapper.java
```diff
@@ -53,170 +58,221 @@
...
+        }
+    }
+
+    private void initInternalAPI() {
+        if (this.internalAPI == null && this.internalAPIUrls != null) {
+            for (String lookupURL : this.internalAPIUrls) {
+                try {
+                    this.internalAPI = PAActiveObject.lookupActive(SynchronizationInternal.class, lookupURL);
+                    return;
+                } catch (Exception e) {
```
Mykhailenko: Maybe we can distinguish case when we cannot lookup because it is not alive and other cases.

sonarqube-activeeon:
SonarQube analysis reported 14 issues
* ![MINOR][MINOR] 14 minor

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

PR #3272, opened 2018-07-27
Fix unpause mechanism

Mykhailenko:


fviale:
if you could add a scheduler db test to reproduce the problem, that would be great.

Mykhailenko:
Retest this please

PR #3273, opened 2018-07-27
Lite refactoring

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -1604,17 +1605,15 @@ public ExecutableContainer loadExecutableContainer(final InternalTask task) {
...
-                                                nameAndCount[0].toString(),
-                                                0,
-                                                Long.parseLong(nameAndCount[2].toString()),
-                                                Integer.parseInt(nameAndCount[1].toString())));
-            }
-            return users;
+            List<Object[]> list = query.list();
+
+            return list.stream()
+                       .map(nameAndCount -> new SchedulerUserInfo(null,
```
MaelAudren: Maybe you could use :
Arrays.steam(query.list()).map(nameAndCount ->...

Mykhailenko: I will keep my version, cause Arrays.stream need array not list

PR #3284, opened 2018-08-02
Externalize IAM configuration and update authentication code

nebilbenmabrouk:
Three main changes are introduced:

i) A new folder 'config/iam' is added. It contains all the config files required by IAM. This allows users to change IAM configuration easily (instead of looking at the config files inside 'iam.jar' as previously done).

ii) The code of the 'boot-microservices' module is updated to start IAM with a specific SpringBoot profile called 'proactive'. This profile uses the config files introduced above.

iii) The authentication code in the sub-module 'common:common-client' is updated because IAM is upgraded to use a new version of CAS (CAS 5.3.2). The returned JWT have no longer the same format. The code is changed accordingly.


boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/IAMStarter.java
```diff
@@ -61,20 +61,25 @@ private IAMStarter() {
     /**
      *  Start IAM microservice
      */
-    public static Process start(String paHome, String bootMicroservicesPath, String bootConfigurationPath)
+    public static Process start(String paHome, String iamMicroservicesPath, String iamConfigurationPath)
```
Mykhailenko: Could you please make a javadoc comment, where you explain iamMicroservicesPath and iamConfigurationPath

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

common/common-client/src/main/java/org/ow2/proactive/authentication/iam/IAMRestClient.java
```diff
@@ -67,17 +68,17 @@ public String getSSOTicket(String url, String username, String password, String
                 throw new IAMException("Failed to acquire SSO ticket for the user" + username);
             }
 
-            Header[] headers = httpResponse.getHeaders(ticketHeader);
+            String htmlResponse = EntityUtils.toString(httpResponse.getEntity());
+            Document doc = Jsoup.parse(htmlResponse);
+            String tgtAttribute = doc.getElementsByAttribute(ticketMarker).first().attributes().get(ticketMarker);
```
MaelAudren: tgtAttribute ? It is not clear for me maybe you could add some comment.

config/iam/cas/activeeon-private.key
```diff
@@ -0,0 +1,15 @@
+-----BEGIN RSA PRIVATE KEY-----
```
MaelAudren: Why are we publishing a private key ?

nebilbenmabrouk: The user can configure IAM to sign and encrypt some information (e.g., JWT Json Web Token, credentials)  that will be sent to IAM clients. We use a private and a public key to sign and encrypt such information.

common/common-client/src/main/java/org/ow2/proactive/authentication/iam/IAMRestClient.java
nebilbenmabrouk: OK, I'll change it to 'htmlSSOTicket'. When we send a REST request to IAM to get a SSO ticket, it responds in a html format. So we parse the reponse to get a html SSO ticket, then we extract the final ticket from that 'htmlSSOTicket'.

config/iam/cas/activeeon-private.key
MaelAudren: I think that a private key published on github is not really a private anymore.

nebilbenmabrouk: I agree, and this is why the user has to change the keys in production. The ones in github are just for demo purposes.

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

PR #3305, opened 2018-08-28
IAM microservice:: Support signed and encrypted JWT

nebilbenmabrouk:
JAAS authentication using signed and encrypted JWT is henceforth supported. 
- The common::common-client module is changed. The classes JWTUtils and JWTValidator are modified to deal with signed and encrypted JWT.
- config/iam/application-proactive.properties is changed to add JWT signature and encryption configuration (so IAM issues signed and encrypted JWTs)
- config/authentication/jaas.config is modified to consider JWT signature and encryption configuration (so JAAS can verify and decode signed and encrypted JWTs)


common/common-client/src/main/java/org/ow2/proactive/authentication/iam/JWTUtils.java
```diff
@@ -42,38 +49,141 @@ private JWTUtils() {
...
+    private static JwtClaims parsePlainJWT(String jwt, String issuer, String service) {
+
+        try {
+
+            JsonWebSignature jws = new JsonWebSignature();
+            jws.setAlgorithmHeaderValue(AlgorithmIdentifiers.NONE);
+            jws.setAlgorithmConstraints(AlgorithmConstraints.ALLOW_ONLY_NONE);
+            jws.setPayload(jwt);
+
+            JwtConsumer jwtConsumer = new JwtConsumerBuilder().setRequireExpirationTime() // the JWT must have an expiration time
```
fviale: is the expiration time configurable ?

fviale: Also, is it sensible to machine clock differences ?

common/common-client/src/main/java/org/ow2/proactive/authentication/jaas.config
```diff
@@ -22,10 +22,10 @@ SchedulerIAMLoginMethod {
 	roleAttributeName="role"
 	roleSeparator=","
 	ssoTicketMarker="action"
-	jsonWebTokenSigned="false"
-	jsonWebTokenEncrypted="false"
-	jsonWebTokenSignatureKey=""
-	jsonWebTokenEncryptionKey="";
+	jsonWebTokenSigned="true"
+	jsonWebTokenEncrypted="true"
+	jsonWebTokenSignatureKey="_pyp8dL5rHqLXiS6aBzvbJ_LZdBN3-oQqbJVbC5mveTdsZ1r3WM_j3SUENB65XS7EAu7Kop3rfQx81GS_mu1Yw"
```
sophiesongge: this key should be hard written here?

common/common-client/src/main/java/org/ow2/proactive/authentication/iam/JWTUtils.java
fviale: I see in https://bitbucket.org/b_c/jose4j/wiki/JWT%20Examples things like :  
jwtConsumer .setAllowedClockSkewInSeconds(30)  // allow some leeway in validating time based claims to account for clock skew

Isn't it very difficult to handle ?

fviale: A related dicussion:
https://github.com/mozilla/addons-server/issues/1072

This looks like a serious design flaw, must be investigated

nebilbenmabrouk: - The expiration time is configurable at the server side (i.e., in IAM microservice). 
- As IAM makes part of PA (both are executed in the same machine), it is supposed that there is no clock differences.
- Why do you consider the clock skew is difficult to handle?

common/common-client/src/main/java/org/ow2/proactive/authentication/jaas.config
nebilbenmabrouk: Yes, PA is considered as a client of IAM (and it uses JAAS for authentication), so the configuration of PA should be written in jaas.config

common/common-client/src/main/java/org/ow2/proactive/authentication/iam/JWTUtils.java
fviale: Ok I understand, thanks

PR #3313, opened 2018-09-06
Improve robustness of nodes housekeeping

jrochas:
- validate further properties and make sure the resource manager can start in case a failure happen in the starting of the nodes housekeeping service.

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/housekeeping/NodesHouseKeepingService.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/housekeeping/NodesHouseKeepingService.java
@@ -50,19 +50,30 @@ public class NodesHouseKeepingService {
     }
 
     public void start() {
-        if (PAResourceManagerProperties.RM_UNAVAILABLE_NODES_REMOVAL_FREQUENCY.isSet() &&
-            PAResourceManagerProperties.RM_UNAVAILABLE_NODES_MAX_PERIOD.isSet() &&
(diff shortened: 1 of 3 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/housekeeping/NodesHouseKeepingService.java
```diff
@@ -50,19 +50,31 @@ public NodesHouseKeepingService(RMCore rmCoreStub) {
...
+    private boolean validProperties() {
+        boolean valid = false;
+        try {
+            valid = PAResourceManagerProperties.RM_UNAVAILABLE_NODES_REMOVAL_FREQUENCY.isSet() &&
+                    PAResourceManagerProperties.RM_UNAVAILABLE_NODES_MAX_PERIOD.isSet() &&
+                    PAResourceManagerProperties.RM_UNAVAILABLE_NODES_MAX_PERIOD.getValueAsInt() > 0;
+        } catch (RuntimeException e) {
+            logger.error("Nodes housekeeping properties are not valid. Nodes housekeeping will be deactivated.", e);
+        }
+        return valid;
```
MaelAudren: Why not : 
```java
try {
            return PAResourceManagerProperties.RM_UNAVAILABLE_NODES_REMOVAL_FREQUENCY.isSet() &&
                    PAResourceManagerProperties.RM_UNAVAILABLE_NODES_MAX_PERIOD.isSet() &&
                    PAResourceManagerProperties.RM_UNAVAILABLE_NODES_MAX_PERIOD.getValueAsInt() > 0;
        } catch (RuntimeException e) {
            logger.error("Nodes housekeeping properties are not valid. Nodes housekeeping will be deactivated.", e);
            return false;
        }

PR #3316, opened 2018-09-07
IAM Microservice:: Fix PA and IAM startup (with errors) issues

nebilbenmabrouk:
The 'boot-microservices' module is slightly changed to deal with startup issues if IAM and PA.

1) IAMStarter reads the output of IAM (execution), detects problems, and throws errors accordingly. Hence PA does not start when IAM is not correctly started.

2) A shutdown hook is added to IAMStarter, hence when PA is not started correctly, IAM also is shutdown.


boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/util/IAMConfiguration.java
```diff
@@ -48,6 +48,8 @@
 
     public static final String READY_MARKER = "iam.ready.marker";
 
+    public static final String ERROR_MARKER = "iam.error.markercle";
```
jrochas: Is it normal that the string ends with "cle"? I would have expected "iam.error.markercle".

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/IAMStarter.java
```diff
@@ -113,14 +121,19 @@ private static String streamOutput(InputStream inputStream) throws ExecutionExce
...
         Future<String> future = executor.submit((Callable) () -> {
 
             try (BufferedReader br = new BufferedReader(new InputStreamReader(inputStream))) {
                 String line;
 
                 while ((line = br.readLine()) != null) {
-                    if (line.contains(config.getString(IAMConfiguration.READY_MARKER))) {
+                    if (line.contains(readyMarker)) {
                         break;
+                    } else if (line.contains(errorMarker)) {
```
jrochas: Maybe you could also log the error before throwing the exception?

nebilbenmabrouk: I can only log the read line. I tried it and it gives no information to the user. I deem that it is cleaner to do that way. The user can see the error details in IAM logs.

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

PR #3340, opened 2018-09-24
IAM microservice :: Prepare web client integration 

nebilbenmabrouk:
This PR aims at preparing the proper ground for configuring PA web microservices as IAM clients. Two main changes are introduced (two operations are executed after IAM startup, i.e., post-startup operations):

1) SSL certificate located in $PA_Home/config/web/keystore is added dynamically to the JVM instance (trustStore) used at the client side (i.e., used by jetty hosting PA web microservices). A class called SSLUtils is added to the module 'boot-microservices'.

Henceforth, IAM and PA will use the same keystore  located under $PA_Home/config/web/keystore (IAM previously uses $PA_Home/config/iam/cas/keystore). 

2) configuration parameters that will be used by PA micoservices authentication filters are passed as system parameters. These parameters mainly include IAM URL and jetty base URL.


boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/util/SSLUtils.java
```diff
@@ -0,0 +1,95 @@
...
+import com.oneandone.compositejks.KeyStoreLoader;
+import com.oneandone.compositejks.SslContextUtils;
+
+
+/**
+ * Utility methods for SSL context.
+ */
+public final class SSLUtils {
+
+    private static final String X509_ALGORITHM = "SunX509";
```
Mykhailenko: Could you please put some comment, What is and what for we use SunX509 algorithm. Thanks

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/IAMStarter.java
```diff
@@ -26,6 +26,7 @@
 package org.ow2.proactive.boot.microservices;
 
 import java.io.*;
```
jrochas: Ideally you could remove the star imports.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/IAMStarter.java
```diff
@@ -107,6 +116,17 @@ public static Process start(String paHome, String iamMicroservicesPath, String i
 
             LOGGER.debug(streamOutput(process.getInputStream()));
 
+            /*
```
jrochas: The format of this comment looks a bit weird.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/IAMStarter.java
nebilbenmabrouk: The star import is automatically generated (by idea) as I import all the classes under 'java.io' and 'java.util.concurrent'. The star allows for reducing the size of imports in such a case.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/IAMStarter.java
nebilbenmabrouk: I agree. I just wanted to mark the post-startup operations with a specific comment. I'll try to find a better solution.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/util/SSLUtils.java
jrochas: Also, did you consider moving this to IAMConfiguration ?

rm/rm-server/src/test/java/functionaltests/utils/IAMTHelper.java
```diff
@@ -26,6 +26,7 @@
 package functionaltests.utils;
 
 import java.io.*;
```
jrochas: Same comment as before for the imports.

nebilbenmabrouk: fixed

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/util/SSLUtils.java
nebilbenmabrouk: SSL params are moved to '$PA_Home/config/iam/application-proactive.properties'. Comments are added to explain these SSL params.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/IAMStarter.java
nebilbenmabrouk: After investigating. This seems the right way to add a multi-line comment inside a method. We use '/*' with only one star ('/**' is a documentation comment that can be added only outside methods).

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

nebilbenmabrouk:
re-test this please

Issue #3330, opened 2018-09-17
Variables values are removed the Data Base when the Job is finished


marcocast:
When a job is finished, the variables values are lost in the scheduler portal.

Mykhailenko:
Resolved in 8.3.1

PR #3347, opened 2018-09-27
Do not remove job vars after job is finished

Mykhailenko:
https://github.com/ow2-proactive/scheduling/issues/3330


fviale:
Can you double-check that the variables are removed from the DB when the job is removed ?

Mykhailenko:
@fviale you are right. With this PR vars will remain forever (until housekeeping launches).

So I will remove vars when LiveJobs::removeJob is called

fviale:
@Mykhailenko isn't housekeeping doing the same as removeJob ?
If removeJob does not remove entries, I double housekeeping would.

Mykhailenko:
@fviale could you look again. I move single line, before we were removing vars when job become finished, but now it will be removed only when whole job is removed. Are you ok with it?

PR #3363, opened 2018-10-05
Popup DME so it does not restart transaction

Mykhailenko:
TransactionHelper helper considers all exceptions as database errors. So it retries each transaction 5 times (also there are delays between transactions). However, in SchedulerDBManager we throw some exceptions,  these exceptions are business exceptions, so there is nothing to retry. Thus if TransactionHelper catches DatabaseManagerError or IllegalArgumentException, it re-throws them.

### DIFF
--- a/common/common-db/src/main/java/org/ow2/proactive/db/TransactionHelper.java
+++ b/common/common-db/src/main/java/org/ow2/proactive/db/TransactionHelper.java
@@ -94,7 +94,7 @@ public class TransactionHelper {
         for (int i = 0; i <= maximumNumberOfRetries; i++) {
             try {
                 return tryExecuteTransaction(sessionWork, readWriteTransaction, readOnlyEntities);
-            } catch (DatabaseManagerException e) {
+            } catch (DatabaseManagerException | IllegalArgumentException e) {
                 throw e;
             } catch (Throwable exception) {
                 lastException = exception;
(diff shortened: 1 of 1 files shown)


common/common-db/src/main/java/org/ow2/proactive/db/TransactionHelper.java
```diff
@@ -94,6 +94,8 @@ public TransactionHelper(SessionFactory sessionFactory) {
         for (int i = 0; i <= maximumNumberOfRetries; i++) {
             try {
                 return tryExecuteTransaction(sessionWork, readWriteTransaction, readOnlyEntities);
+            } catch (DatabaseManagerException e) {
```
fviale: I think there are IllegalArgumentException thrown as well, can you check if there are other expected exception ?

common/common-db/src/main/java/org/ow2/proactive/db/TransactionHelper.java
```diff
@@ -124,6 +126,8 @@ public TransactionHelper(SessionFactory sessionFactory) {
             session.getTransaction().commit();
 
             return result;
+        } catch (DatabaseManagerException | IllegalArgumentException e) {
+            throw e;
```
Mykhailenko: @fviale I believe here I have to rollback as well. DME and IAE are both thrown by us, but transaction still has to be rollbacked. Isn't it?

PR #3366, opened 2018-10-09
Add Scheduler Vars replacement in job parsing

amouhoub:
In this PR, we apply variable substitution for SchedulerVars : `PA_SCHEDULER_REST_URL` and `PA_CATALOG_REST_URL` during job parsing.

The main goal behind this is to allow referencing Catalog objects with relative Scheduler Catalog URL links.

Therefore, from now on, a task implementation that uses a catalog script by reference will have the following URL:
`${PA_CATALOG_REST_URL}`/buckets/scripts/my_script`


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/StaxJobFactory.java
```diff
@@ -102,6 +103,10 @@
 
     private static final String FILE_ENCODING = PASchedulerProperties.FILE_ENCODING.getValueAsString();
 
+    private static final String PA_CATALOG_REST_URL = PASchedulerProperties.CATALOG_REST_URL.getValueAsString();
```
fviale: I think this static initialization will create issues as the scheduler & catalog rest are not initialized on startup.

Can you do a lazy initialization instead ?

PR #3369, opened 2018-10-11
 Source $HOME/.bashrc from the impersonated process

fviale:
Without this the impersonated process does not load automatically the user .bashrc file if defined (su and ssh does not load automatically the environment in this context).


tobwiens:
Retest this please

fviale:
Retest this please

PR #3370, opened 2018-10-11
Fix node source thread exhaustion prevents from detecting down nodes

jrochas:


jrochas:
retest this please

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
```diff
@@ -112,6 +111,10 @@
 
     public static final int EXTERNAL_POOL = 1;
 
+    public static final int PINGER_POOL = 2;
```
Mykhailenko: in regular day i would ask for java doc and maybe load "2" from configuration.

jrochas:
retest this please

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
jrochas: Actually the "2" is the index of the thread pools array. But you are right, then the name of the constant should be PINGER_POOL_INDEX.

PR #3375, opened 2018-10-12
Add a new Variable Model Type CATALOG_OBJECT

Aminelouati:
In this PR, we add a new Variable Model Type CATALOG_OBJECT and some unit tests

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CatalogObjectValidator.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CatalogObjectValidator.java
@@ -1,0 +1,61 @@
+/*
+ * ProActive Parallel Suite(TM):
+ * The Open Source library for parallel and distributed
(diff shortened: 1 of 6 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CatalogObjectValidator.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @author ActiveEon Team
+ * @since 11/10/2018
+ */
+public class CatalogObjectValidator implements Validator<String> {
+
+    //catalogObjectModelRegexp = bucketName/objectName[/revision]. The revision is a h-code number represented by 13 digit.
+
+    public static final String CATALOG_OBJECT_MODEL_REGEXP = "^[^/]+/[^/]+(/[^/][0-9]{12})?$";
```
Mykhailenko: I think examples of the good and bad strings could be useful.

Aminelouati:
re-test please

PR #3395, opened 2018-10-23
Allow timeout on Authentication.isActivated

fviale:
The previous implementation used a ProActive synchronous call (primitive boolean return value) which can block indefinitely. This could cause random freeze of a windows agent trying to contact the RM (the precise reason why the method never return is still uncertain).

This modification allow to not wait more than the expected timeout, allowing retries on windows agents.

An immediate service call is also added to try to avoid the root cause of the issue.

### DIFF
--- a/common/common-client/src/main/java/org/ow2/proactive/authentication/Connection.java
+++ b/common/common-client/src/main/java/org/ow2/proactive/authentication/Connection.java
@@ -35,11 +35,14 @@ import java.net.UnknownHostException;
 
 import org.apache.log4j.Logger;
 import org.objectweb.proactive.api.PAActiveObject;
+import org.objectweb.proactive.api.PAFuture;
+import org.objectweb.proactive.core.ProActiveTimeoutException;
 import org.objectweb.proactive.core.config.CentralPAPropertyRepository;
 import org.objectweb.proactive.core.exceptions.NotBoundException;
 import org.objectweb.proactive.core.remoteobject.AbstractRemoteObjectFactory;
 import org.objectweb.proactive.core.remoteobject.RemoteObjectFactory;
 import org.objectweb.proactive.core.util.ProActiveInet;
+import org.objectweb.proactive.core.util.wrapper.BooleanWrapper;
 
 
 /**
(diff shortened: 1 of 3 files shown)


common/common-client/src/main/java/org/ow2/proactive/authentication/Connection.java
```diff
@@ -162,8 +165,16 @@ public T waitAndConnect(String url, long timeout) throws Exception {
             // waiting until scheduling is initialized
             while (leftTime > 0) {
```
vinseon: I'm not very familiar with all this stuff but don't you think that we can get rid of the whole `while` loop, decrementing `leftTime` variable, `sleepOrThrow` method (thanks to the new `waitFor` call) and just throw `ERROR_NOT_ACTIVATED` if the boolean is false?

fviale: There are two wait mechanism actually.

1) The PAFuture.waitFor waits for the server to respond to the request (normally it should be fast, but in case there is some problem we add here a timeout).
2) after the server responded we check the status of the server (is it activated or not). If the server is not yet activated, we loop until it is activated (case of server startup).

common/common-client/src/main/java/org/ow2/proactive/authentication/Connection.java
```diff
@@ -162,8 +165,16 @@ public T waitAndConnect(String url, long timeout) throws Exception {
...
                 long startTime = System.currentTimeMillis();
+                BooleanWrapper future = authentication.isActivated();
```
sonarqube-activeeon: MAJOR A "NullPointerException" could be thrown; "authentication" is nullable here. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS2259)

common/common-client/src/main/java/org/ow2/proactive/authentication/Connection.java
```diff
@@ -162,8 +165,16 @@ public T waitAndConnect(String url, long timeout) throws Exception {
...
 
-                if (authentication.isActivated()) {
+                try {
```
sonarqube-activeeon: MAJOR Extract this nested try block into a separate method. [rule](https://sonarcloud.io/coding_rules#rule_key=squid%3AS1141)

sonarqube-activeeon:
SonarQube analysis reported 2 issues
* ![MAJOR][MAJOR] 2 major

[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

PR #3398, opened 2018-10-24
add dockerfile.properties

cdiop:


config/scriptengines/dockerfile.properties
```diff
@@ -0,0 +1,5 @@
+# General configuration
+docker.file.command=/usr/local/bin/docker
+docker.file.sudo.command=/usr/bin/sudo
+docker.file.use.sudo=false
+docker.host=
```
tobwiens: Can we can remove docker host ?

cdiop: For sur 
Done

PR #3400, opened 2018-10-25
Store unresolved job generic information and variables in Job

jrochas:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/StaxJobFactory.java
```diff
@@ -1821,7 +1860,14 @@ private static String replace(String str, Map<String, String> variables) throws
 
         // Include given variables if any
         if (variables != null) {
-            replacements.putAll(variables);
+            for (Map.Entry<String, String> variable : variables.entrySet()) {
```
fviale: why this code change ? could you add comments to explain ?

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/task/CommonAttribute.java
```diff
@@ -169,6 +171,20 @@ public void setMaxNumberOfExecution(int numberOfExecution) {
...
     }
 
+    /**
+     * Returns the generic information without variable replacements.
+     *
+     * @return unresolved generic information.
+     */
+    public Map<String, String> getUnresolvedGenericInformation() {
+        Set<Entry<String, String>> entries = this.unresolvedGenericInformation.entrySet();
+        Map<String, String> result = new HashMap<>(entries.size());
```
fviale: let's use LinkedHashMap

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/task/CommonAttribute.java
```diff
@@ -71,6 +71,8 @@
     /** Common user informations */
     protected Map<String, String> genericInformation = new HashMap<String, String>();
 
+    protected Map<String, String> unresolvedGenericInformation = new HashMap<>();
```
fviale: let's use LinkedHashMap to avoid order changes in the catalog

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/StaxJobFactory.java
jrochas: Yes. If a system variable is overriden by a variable, then the system variable value should apply, correct?
This is what the following code says. Otherwise, we must decide which value preempts the other.

PR #3401, opened 2018-10-26
Check for hsqldb

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/HsqldbServer.java
```diff
@@ -125,21 +125,25 @@ public void addCatalog(Path defaultLocation, Path hibernateConfiguration) throws
...
-        String catalogUsername = hibernateProperties.getProperty(PROP_HIBERNATE_CONNECTION_USERNAME);
-
-        Path catalogPath;
-        if (catalogLocation == null) {
-            // Catalog path references a file.
-            // Double resolve is used to get a dedicated folder with the catalog name.
-            catalogPath = defaultLocation.resolve(catalogName).resolve(catalogName);
-        } else {
-            catalogPath = Paths.get(catalogLocation);
+        if (connectionUrl.toLowerCase().contains("hsqldb")) {
```
jrochas: So, this would prevent error messages in the logs when using another DB than HSQL, correct?

fviale: we hope so

Mykhailenko:
Retest this please

PR #3402, opened 2018-10-26
IAM microservice :: REST API as an IAM client

nebilbenmabrouk:
The goal of this PR is to yield PA sessions using the new IAM microservice.

 - It introduces IAMSession that extends the legacy Session in ProActive.

- IAMSession is used to to perform REST operations (e.g., submit a job). When the session is not used for a period of time (set to 1 hour), it expires.  Otherwise it is automatically renewed (i.e.,  the elapsed time is reset to 0).

 - The sessionId (of IAMSession) is generated as a cyphered JWT (Json Web Token), or a ST (Service Token). This can be configured via 'config/iam/application-proactive.properties'

 - The JWT includes the user credentials (login & cyphered password), and an IAM SSO ticket.

- REST disconnect operations remove  sessions from IAM.

- IAMLoginModule (used by ProActive for login) can henceforth perform authentication against IAM using three methods: (i) generate a ST, (ii) generate a JWT, or (iii) simply validate user creds without creating any sessions (i.e., it simply checks user credentilas against IAM). The latter method is the one used by default.


boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMSessionUtil.java
```diff
@@ -0,0 +1,111 @@
...
+
+    public boolean deleteToken(String ssoTicket) {
+        return new IAMRestClient().deleteSSOTicket(iamURL + IAMConfiguration.IAM_TICKET_REQUEST + "/" + ssoTicket);
+    }
+
+    public boolean tokenIsValid(String ssoTicket) {
+        return new IAMRestClient().isSSOTicketValid(iamURL + IAMConfiguration.IAM_TICKET_REQUEST + "/" + ssoTicket);
+    }
+
+    public AbstractMap.SimpleEntry<String, JwtClaims> createNewSessionToken(String username, char[] password) {
```
Mykhailenko: According google code convention: each public method/field should have javadoc.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMRestClient.java
```diff
@@ -0,0 +1,222 @@
...
+
+    public String validateUserCredentials(String url, String username, char[] password) {
+
+        try (CloseableHttpClient httpClient = new CommonHttpClientBuilder().allowAnyCertificate(true).build()) {
+
+            HttpPost httpPost = new HttpPost(url);
+
+            List<NameValuePair> params = new ArrayList<>();
+            params.add(new BasicNameValuePair(USER_KEY, username));
+            params.add(new BasicNameValuePair(SECRET_KEY, String.valueOf(password)));
```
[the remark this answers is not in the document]
nebilbenmabrouk: List<NameValuePair> params is cleared just after HTTP request is sent.
The REST API of IAM accepts only String (and not char array)

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMSessionUtil.java
nebilbenmabrouk: Java doc is added to all public methods.

nebilbenmabrouk:
re-test this please

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMJsonUtils.java
```diff
@@ -0,0 +1,100 @@
...
+import org.jose4j.json.internal.json_simple.JSONObject;
+import org.jose4j.json.internal.json_simple.parser.JSONParser;
+import org.jose4j.json.internal.json_simple.parser.ParseException;
+
+
+public class IAMJsonUtils {
+
+    /**
+     * Attributes used to parse a (json) credentials validation answer from IAM
+     * Example of json answer:
```
jrochas: Maybe you can put a version number for this example.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMRestClient.java
```diff
@@ -0,0 +1,266 @@
...
+
+            if (asJWT) {
+                params.add(new BasicNameValuePair(TOKEN_KEY, TOKEN_VALUE));
+            }
+
+            httpPost.setEntity(new UrlEncodedFormEntity(params));
+
+            HttpResponse httpResponse = httpClient.execute(httpPost);
+
+            params = new ArrayList<>();
```
jrochas: Why is `param` reset here? It is not used afterwards.

nebilbenmabrouk: I use char[] array for representing passwords instead of String (since String is immutable). Nevertheless, the REST API (of CAS) accepts only String (and not char[] array). So, I send the password as String then I reset the sent params to avoid accessing the value of the password.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMRestClient.java
```diff
@@ -0,0 +1,266 @@
...
+            HttpPost httpPost = new HttpPost(url);
+
+            List<NameValuePair> params = new ArrayList<>();
+            params.add(new BasicNameValuePair(SERVICE_KEY, service));
+            httpPost.setEntity(new UrlEncodedFormEntity(params));
+
+            HttpResponse httpResponse = httpClient.execute(httpPost);
+
+            if ((httpResponse.getStatusLine().getStatusCode() != HttpStatus.SC_OK) ||
+                (null == httpResponse.getEntity()))
```
jrochas: Having an if with no brackets is not within our code practice generally, don't you think?

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMSessionUtil.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * indicates whether the IAM session is still valid.
+     *
+     * @param ssoTicket SSO ticket of IAMSession to be validated
+     * @return boolean indicating whether IAM session is valid.
+     * @since version 8.3.0
+     */
+    public boolean sessionIsValid(String ssoTicket) {
+        return new IAMRestClient().isSSOTicketValid(iamURL + IAMConfiguration.IAM_TICKET_REQUEST + "/" + ssoTicket);
```
jrochas: Can't you instantiate once the IAMRestClient and save it?

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMSessionUtil.java
```diff
@@ -0,0 +1,150 @@
...
+                                                               isJWTSession);
+
+        JwtClaims claims = isJWTSession ? JWTUtils.parseJWT(sessionToken,
+                                                            tokenSignatureEnabled,
+                                                            tokenSignatureKey,
+                                                            tokenEncryptionEnabled,
+                                                            tokenEncryptionKey,
+                                                            iamURL,
+                                                            iamURL)
+                                        : null;
```
jrochas: Is there any other value that we could store than null? Because then, it will be the value of a map record, and when searching by key afterwards we won't know whether there is a key and its value is null or if there is not such key in the map.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/JWTUtils.java
```diff
@@ -40,6 +40,7 @@
 import org.jose4j.jwt.consumer.JwtConsumerBuilder;
 import org.jose4j.keys.AesKey;
 import org.jose4j.lang.JoseException;
+import org.ow2.proactive.boot.microservices.iam.exceptions.IAMException;
```
jrochas: I don't see any other change using this import

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/JWTValidator.java
```diff
@@ -132,15 +109,15 @@ public Assertion validate(String jsonWebToken, String service) throws TicketVali
...
+            // Create authentication assertion
             Assertion assertion = new AssertionImpl(new AttributePrincipalImpl(principal.getName(),
                                                                                claims.getClaimsMap()));
 
             logger.debug("Assertion generated for principal " + principal.getName());
 
             return assertion;
         } catch (MalformedClaimException e) {
-            throw new IAMException("");
+            throw new IAMException("Malformed Json Web Token acquired from IAM [" + jsonWebToken + "]");
```
jrochas: Can you link the exception cause `e` here?

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/OperatingSystem.java
```diff
@@ -0,0 +1,93 @@
...
+    IRIX("Irix", OperatingSystemFamily.UNIX),
+    DIGITAL_UNIX("Digital Unix", OperatingSystemFamily.UNIX),
+    NETWARE("NetWare", OperatingSystemFamily.UNIX),
+    OSF1("OSF1", OperatingSystemFamily.UNIX),
+    OPENVMS("OpenVMS", OperatingSystemFamily.DEC_OS),
+    WINDOWS_GENERIC("Windows", OperatingSystemFamily.WINDOWS),
+    LINUX_OS("Linux", OperatingSystemFamily.LINUX),
+    MAC_OS("Mac OS", OperatingSystemFamily.MAC);
+
+    final private String label;
```
jrochas: Sonar will tell you that keywords are in wrong order :-) 'private final'

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -147,10 +162,23 @@ private RMProxyUserInterface checkAccess(String sessionId) throws NotConnectedEx
     public String rmConnect(@FormParam("username") String username, @FormParam("password") String password)
             throws KeyException, LoginException, RMException, ActiveObjectCreationException, NodeException {
 
-        Session session = sessionStore.create(username);
+        if ((username == null || username.trim().isEmpty()) || (password == null || password.trim().isEmpty())) {
```
jrochas: You could use here instead the Apache Commons `StringUtils#isBlank`.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -178,18 +211,43 @@ public String loginWithCredential(@MultipartForm LoginForm multipart) throws Act
             NodeException, KeyException, IOException, LoginException, RMException {
 
         Session session;
+        CredData credData;
+
         if (multipart.getCredential() != null) {
-            session = sessionStore.createUnnamedSession();
             Credentials credentials = Credentials.getCredentials(multipart.getCredential());
+            String schedulerPrivateKey = PASchedulerProperties.getAbsolutePath(PASchedulerProperties.SCHEDULER_AUTH_PRIVKEY_PATH.getValueAsString());
+            credData = credentials.decrypt(schedulerPrivateKey);
```
jrochas: What about if credentials cannot be decrypted. Don't we need a try/catch or check here?

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -143,6 +139,13 @@
 
     private static final String PATH_TASKS = "/tasks/";
 
+    private static final String IAM_LOGIN_METHOD = "IAMLoginMethod";
+
+    public static final Boolean IAM_IS_USED = PASchedulerProperties.SCHEDULER_LOGIN_METHOD.getValueAsString()
```
jrochas: I believe those constants appear several times. Can we factorize?

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -2925,11 +2938,24 @@ public boolean isConnected(@HeaderParam("sessionid")
     public String login(@FormParam("username") String username, @FormParam("password") String password)
             throws LoginException, SchedulerRestException {
         try {
-            if ((username == null) || (password == null)) {
+
+            if ((username == null || username.trim().isEmpty()) || (password == null || password.trim().isEmpty())) {
```
jrochas: Same remark for the use of Apache Commons here.

rm/rm-server/src/test/java/functionaltests/authentication/ConnectionTest2.java
```diff
@@ -63,19 +63,13 @@ public void action() throws Exception {
...
-                }
-            }
-        };
-        t.start();
+        try {
+            Thread.sleep(1000);
+            log("Running resource manager");
+            rmHelper.getRMAuth();
+        } catch (Exception e) {
+            log("Failed: unexpected error " + e.getMessage());
```
jrochas: The stack trace could also be useful here.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMJsonUtils.java
nebilbenmabrouk: I prefer not adding a version for this, to avoid updating it (in the scheduling project) each time the iam project is updated. If the json produced by IAM changes, this will create issues in the scheduling project. The code and related javadoc will be then updated, accordingly.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMRestClient.java
nebilbenmabrouk: I agree. Fixed.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/IAMSessionUtil.java
nebilbenmabrouk: When the value of the map is null, no search can be performed on it, and a NullPointerException will be thrown. I did it in purpose as a 'fail-fast' mode to avoid using this map (JwtClaims) when the session id is not a JWT (json web token).

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/JWTUtils.java
nebilbenmabrouk: The name of the package changed.

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/JWTValidator.java
nebilbenmabrouk: done

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/OperatingSystem.java
nebilbenmabrouk: fixed

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
nebilbenmabrouk: If credentials cannot be decrypted, an exception will be thrown, and it is better to not catch it. All its details/trace will appear in the sent response.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
nebilbenmabrouk: I agree, it is better to use StringUtils. Done :)

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
nebilbenmabrouk: done

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/OperatingSystem.java
```diff
@@ -0,0 +1,93 @@
...
+ * or a different license than the AGPL.
+ */
+package org.ow2.proactive.boot.microservices.iam.util;
+
+/**
+ * OperatingSystem
+ *
+ * @author The ProActive Team
+ */
+public enum OperatingSystem {
```
fviale: Instead of rewriting this class in your project, I think it would be cleaner to move the OS & OSF classes from common-client to common-api.

I suppose your project could inherit from common-api without much trouble, isn't it ?

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
nebilbenmabrouk: Yes I agree, factorized

boot-microservices/src/main/java/org/ow2/proactive/boot/microservices/iam/util/OperatingSystem.java
nebilbenmabrouk: OS & OSF centralized in 'common-api'. Thanks

PR #3403, opened 2018-10-27
Add GIs in non iteractive submit cli command + Make JobKeyValueTransf…

mbenguig:
…ormer static + Fix issue in GIs submission + tests


mbenguig:
Replaced by another one.

PR #3414, opened 2018-11-12
Fix job-without-vars submission from wf automation bug

amouhoub:
In this PR, we fix the bug found in automation-dashboard that calls validateurl endpoint before submission and mistakenly inserts a `;` at the end of the url. This insertion is misinterpreted by the scheduler as a variables map with an empty sting as a key and a null as a value.

Please note that the same happens when calling job submission API but it doesn't seem affected by this.

### DIFF
--- a/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/util/WorkflowVariablesTransformer.java
+++ b/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/util/WorkflowVariablesTransformer.java
@@ -38,9 +38,9 @@ public class WorkflowVariablesTransformer {
     public Map<String, String> getWorkflowVariablesFromPathSegment(PathSegment pathSegment) {
         Map<String, String> variables = null;
         MultivaluedMap<String, String> matrixParams = pathSegment.getMatrixParameters();
-        // Remove any empty keys that might be mistakenly sent to the scheduler to prevent bad behaviour
-        matrixParams.remove("");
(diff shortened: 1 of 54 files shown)


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/util/WorkflowVariablesTransformer.java
```diff
@@ -38,6 +38,8 @@
     public Map<String, String> getWorkflowVariablesFromPathSegment(PathSegment pathSegment) {
         Map<String, String> variables = null;
         MultivaluedMap<String, String> matrixParams = pathSegment.getMatrixParameters();
+        // Remove any empty keys that might be mistakenly sent to the scheduler to prevent bad behaviour
+        matrixParams.remove("");
```
fviale: please move this statement inside the if, as matricParams can be null

amouhoub: @fviale In fact this can never be null because the pathSegment.getMatrixParameters() returns always returns a map, even an empty one.
btw sonar insisted that I remove the !=null test but I didn't want to touch that

fviale: sure, but even if it cannot be null, you need to put this remove inside the check, otherwise the code  is not readable, removing makes no sense on a null or empty map.

PR #3415, opened 2018-11-13
 Update scheduler shutdown action action to freeze and kill the scheduler + update CLI

mbenguig:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -255,28 +255,7 @@ public boolean resume() {
...
-                if (jobs.getRunningTasks().isEmpty()) {
-                    listener.schedulerStateUpdated(SchedulerEvent.SHUTDOWN);
-                } else {
-                    infrastructure.schedule(this, 5000);
-                }
-            }
-        }, 5000);
-
-        return true;
+        return freeze() && kill();
```
fviale: I think this will not be enough, because you must wait until there are no currently running tasks

fviale: part of the previous code was actually ok, except we should not unpauseAll I think, and we should call kill when there are no running tasks. We must see which statuses become obsolete (I think the status SHUTING_DOWN makes sense, but I don't know if we use the status SHUTDOWN or KILLED during the final step)

mbenguig: Thx for this comment.. but not sure I completely understood. Freezing the scheduler already do the job: it makes the scheduler wait for running tasks to be terminated. Moreover we said that shutdown <=> freeze + kill.  For your second comment, as i can see for the other admin actions (freeze, pause), the status is updated before executing the procedure to freeze/pause.. In our case, we would set the scheduler status to SHUTDOWN before shuting down it.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/Scheduler.java
```diff
@@ -521,10 +521,7 @@ void changeJobPriority(JobId jobId, JobPriority priority)
     boolean resume() throws NotConnectedException, PermissionException;
 
     /**
-     * For administrator only, Shutdown the scheduler.<br>
-     * It will terminate every submitted jobs but won't accept new submit.<br>
-     * Use {@link #kill()} if you want to stop the scheduling and exit the
-     * scheduler.
+     * For administrator only, Terminate the scheduler by terminating running tasks.
```
fviale: Terminate the scheduler after all currently running tasks are finished.

PR #3418, opened 2018-11-15
Add custom GUI (specific to ProActive) to IAM

nebilbenmabrouk:
A new GUI theme (that is specific to ProActive) is added to IAM.
The new theme is configured in '$pa_home/config/iam/application-proactive.properties', and its resources are located under $pa_home/config/iam/ui. It contains two sub-folders:
i) static, which contains css, images and static resources, and
ii) templates, which contains html files.

There are no code changes in this PR :)


nebilbenmabrouk:
re-test this please

PR #3424, opened 2018-11-22
EmailNotification Improuvement:: 

sophiesongge:
send to multiple receivers + add task status


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -160,11 +161,16 @@ private String getBody() {
         String jobID = jobState.getId().value();
         String status = jobState.getStatus().toString();
         String hostname = "UNKNOWN";
+        List<TaskState> tasks = jobState.getTasks();
+        StringBuilder taskStatus = new StringBuilder();
+        for (TaskState task : tasks) {
+            taskStatus.append("Task : " + task.getId().toString() + " Status: " + task.getStatus().toString() + "\n");
```
fviale: I prefer task.getId().getReadableName() + " (" + task.getId().toString() +") Status: " + ...

PR #3427, opened 2018-11-23
EmailNotification Improvements: multiple receivers + more detailed mail

sophiesongge:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -160,11 +163,17 @@ private String getBody() {
         String jobID = jobState.getId().value();
         String status = jobState.getStatus().toString();
         String hostname = "UNKNOWN";
+        List<TaskState> tasks = jobState.getTasks();
+        StringBuilder taskStatus = new StringBuilder();
```
jrochas: You could rename the variable to e.g. `allTaskStatusesString`.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -160,11 +163,17 @@ private String getBody() {
...
+        for (TaskState task : tasks) {
+            taskStatus.append(task.getId().getReadableName() + " (" + task.getId().toString() + ") Status: " +
+                              task.getStatus().toString() + "\n");
+        }
```
jrochas: You could replace this for loop and the StringBuilder by:
`String.join(System.lineSeparator(), tasks.stream().map(task -> task.getId().getReadableName() + " (" + task.getId().toString() + ") Status: " + task.getStatus().toString()).collect(Collectors.toList()));`

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -142,12 +144,13 @@ private static String getFrom() throws JobEmailNotificationException {
         return from;
     }
 
-    private String getTo() throws JobEmailNotificationException {
+    private String[] getTo() throws JobEmailNotificationException {
```
jrochas: Why using an array and not a List?

jrochas: I think that hiding the array directly after the split by creating a List from the array would make the code more resilient and make it more usable.

PR #3430, opened 2018-11-27
remove File connectors from scheduler-examples module

Aminelouati:
In this Pr, we remove File connectors (FTP and SFTP) from scheduler-examples module as the implementation is moved inside the workflows.


MaelAudren:
Can you put explanation in the PR why those class were removed ?

PR #3431, opened 2018-11-28
Add the possibility to submit job from catalog with object type CATALOG_OBJECT

MaelAudren:


rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/SchedulerClient.java
```diff
@@ -803,26 +805,53 @@ public JobId submitFromCatalog(String catalogRestURL, String bucketName, String
...
+        Optional<String> revision = getRevisionFromCalledWorkflow(calledWorkflow);
+
+        HttpGet httpGet = revision.map(r -> new HttpGet(catalogRestURL + "/buckets/" +
+                                                        getBucketFromCalledWorkflow(calledWorkflow) + "/resources/" +
+                                                        getNameFromCalledWorkflow(calledWorkflow) + "/revisions/" + r +
+                                                        "/raw"))
+                                  .orElseGet(() -> new HttpGet(catalogRestURL + "/buckets/" +
+                                                               getBucketFromCalledWorkflow(calledWorkflow) +
+                                                               "/resources/" +
+                                                               getNameFromCalledWorkflow(calledWorkflow) + "/raw"));
```
vinseon: Duplicate code can be removed, like this:

```
        Optional<String> revision = getRevisionFromCalledWorkflow(calledWorkflow);
        String workflowURL = catalogRestURL + "/buckets/" + getBucketFromCalledWorkflow(calledWorkflow) + "/resources/" + getNameFromCalledWorkflow(calledWorkflow);
        if (revision.isPresent()) {
            workflowURL +=  "/revisions/" + revision.get();
        }
        HttpGet httpGet = new HttpGet(workflowURL + "/raw");
```

But actually the 3 methods `getRevision`, `getBucket` and `getName` don't do much. I would replace them by a single one generating `workflowURL` directly.

MaelAudren: Thank you for noticing that duplication. I improved the code and removed the duplication.
However the methods `getBucketFromCalledWorkflow` and `getNameFromCalledWorkflow`  are used in two places. I would have to duplicate the code if remove them so I prefer to keep them.

PR #3434, opened 2018-12-07
Adding logging aronud main scheduling loop operations

marcocast:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulingMainLoopTimingLogger.java
```diff
@@ -0,0 +1,73 @@
+package org.ow2.proactive.scheduler.util;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.apache.log4j.Logger;
+import org.python.google.common.collect.Maps;
```
fviale: use guava instead of python

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -160,6 +163,8 @@ private void releaseNodes(InternalJob job, NodeSet nodeSet) throws RMProxyCreati
      * @return the number of tasks that have been started
      */
     public int schedule() {
+        schedulingMainLoopTimingLogger = new SchedulingMainLoopTimingLogger(logger);
```
fviale: the logger should be changed to SchedulingMethodImpl.class

marcocast: Why not passing the already created logger?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/util/SchedulingMainLoopTimingLogger.java
marcocast: ok

PR #3436, opened 2018-12-10
Adding logging around main scheduling loop operations

ShatalovYaroslav:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -119,6 +118,8 @@
 
     private CheckEligibleTaskDescriptorScript checkEligibleTaskDescriptorScript;
 
+    private SchedulingMainLoopTimingLogger schedulingMainLoopTimingLogger;
```
fviale: could you change the following line in this class:

public static final Logger logger = Logger.getLogger(SchedulingService.class);

to

public static final Logger logger = Logger.getLogger(SchedulingMethodImpl.class)

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -162,6 +163,8 @@ private void releaseNodes(InternalJob job, NodeSet nodeSet) throws RMProxyCreati
      * @return the number of tasks that have been started
      */
     public int schedule() {
+        schedulingMainLoopTimingLogger = new SchedulingMainLoopTimingLogger(Logger.getLogger(SchedulingMethodImpl.class));
```
fviale: with the change mentionned above, you can use new SchedulingMainLoopTimingLogger(logger);

PR #3439, opened 2018-12-13
send notifications with job logs as attachment

sophiesongge:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -122,17 +149,19 @@ public boolean doCheckAndSend() throws JobEmailNotificationException {
...
-                boolean sent = doCheckAndSend();
+                boolean sent = doCheckAndSend(withAttachment);
                 if (sent) {
                     jlogger.info(jobState.getId(), "sent notification email for finished job to " + getTo());
                 }
             } catch (JobEmailNotificationException e) {
                 jlogger.warn(jobState.getId(),
                              "failed to send email notification to " + e.getEmailTarget() + ": " + e.getMessage());
                 logger.trace("Stack trace:", e);
+            } catch (Exception e) {
```
fviale: please add jlogger.warn(jobState.getId(),
                             "failed to send email notification to " + e.getEmailTarget() + ": " + e.getMessage());	                             "failed to send email notification to " + e.getEmailTarget() + ": " + e.getMessage());

sophiesongge: I can not add this, because e doesn't have the method e.getMessage(), it is a method of JobEmailNotificationException, so I separate these two.

fviale: you mean e does not have e.getEmailTarget(), indeed.

But you can simply do:

jlogger.warn(jobState.getId(),
                             "failed to send email notification;" + e.getMessage());	                             "failed to send email notification: " + e.getMessage());

sophiesongge: it's done, thanks

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -178,4 +210,54 @@ private String getBody() {
...
+        JobId jobID = jobState.getId();
+        List<TaskState> tasks = jobState.getTasks();
+        String attachLogPath = null;
+
+        try {
+            JobResult result = dbManager.loadJobResult(jobID);
+
+            while (result == null) {
+                try {
+                    Thread.sleep(1000);
```
fviale: you need to do again loadJobResult

PR #3440, opened 2018-12-13
Improve Variable replacement performance

fviale:
- Include initial StrSubstitutor fnd associated classes rom apache commons lang source code with copyrights
   (version 3.4)
 - some tests are failing because StrSubstitutor does not support the
   $variable notation (empty suffixes are not supported)
 - Updated the xml schema to add contraints on variable
 and generic information names (free form is not acceptable)
 - also, allow usage of patterns inside START_AT generic info
 - fix infinite definition in TestStaxJobFactory

 - improve StrSubstitutor to handle variable patterns with empty suffixes, such as $variable
 - added NOTICE information to explain the changes

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/VariableSubstitutor.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/VariableSubstitutor.java
@@ -29,6 +29,8 @@ import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.ow2.proactive.scheduler.common.util.text.StrMatcher;
+import org.ow2.proactive.scheduler.common.util.text.StrSubstitutor;
 import org.ow2.proactive.scripting.Script;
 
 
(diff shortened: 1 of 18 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/VariableSubstitutor.java
```diff
@@ -171,14 +175,74 @@ public static void filterAndUpdate(Script<?> script,
...
             }
         }
 
-        return replacements;
+        StrSubstitutor substitutor = new StrSubstitutor(replacements, SUBSITUTE_PREFIX, SUBSTITUTE_SUFFIX, (char) 0);
+
+        // match the beginning of a variable, can either be $ or ${
+        substitutor.setVariablePrefixMatcher(new StrMatcher() {
+            @Override
+            public int isMatch(char[] buffer, int pos, int bufferStart, int bufferEnd) {
```
Mykhailenko: this method is hard to read. Could you please put javadoc

PR #3447, opened 2018-12-19
send email notification for all the events

sophiesongge:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -107,7 +107,22 @@ public boolean doCheckAndSend(boolean withAttachment)
...
+                                              "JOB_PENDING_TO_RUNNING",
+                                              "JOB_RESTARTED_FROM_ERROR",
+                                              "JOB_RESUMED",
+                                              "JOB_RUNNING_TO_FINISHED",
+                                              "JOB_SUBMITTED")
+                                      .stream()
+                                      .map(status -> status.toLowerCase())
+                                      .collect(Collectors.toList());
+            } else {
+                jobStatusList = Arrays.asList(jobStatus.toLowerCase().split("\\s*,\\s*"));
```
Mykhailenko: And could you please describe what does this pattern mean?

sophiesongge: This line is what exist before. It parse the value of the key "NOTIFICATION EVENTS" in generic info. This value specifies the events that a user want to receive an email notification, and the events are separated by ",". The case of the events does not metter (you can type either lower case or upper case or even a mix)

Mykhailenko: sorry i meant to put comment in the code

sophiesongge: okay

PR #3448, opened 2018-12-27
DB tasks pooled connection using KikariCP

Aminelouati:
In this PR, we
1. Add a new package `connectionpooling` in the scheduler-examples project to enable pooled connection to any external SQL database using HikariCP Framework.
2. Add some unit tests
3. Add a functional test

### DIFF
--- a/scheduler/scheduler-api/build.gradle
+++ b/scheduler/scheduler-api/build.gradle
@@ -13,6 +13,7 @@ dependencies {
     compile project(':common:common-api')
 
     compile 'org.apache.commons:commons-lang3:3.5'
+    compile group: 'com.zaxxer', name: 'HikariCP', version: '2.3.2'
 
     runtime 'com.fasterxml.woodstox:woodstox-core:5.0.1'
     runtime 'msv:relaxngDatatype:20050913'
--- a/scheduler/scheduler-api/build.gradle
+++ b/scheduler/scheduler-api/build.gradle
@@ -13,7 +13,6 @@ dependencies {
     compile project(':common:common-api')
 
     compile 'org.apache.commons:commons-lang3:3.5'
-    compile group: 'com.zaxxer', name: 'HikariCP', version: '2.3.2'
 
     runtime 'com.fasterxml.woodstox:woodstox-core:5.0.1'
     runtime 'msv:relaxngDatatype:20050913'
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/db/connectionpooling/DBConnectionPoolsHolder.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/db/connectionpooling/DBConnectionPoolsHolder.java
@@ -55,6 +55,9 @@ public class DBConnectionPoolsHolder {
     private static final int CACHE_EXPIRATION_MINUTES = Integer.parseInt(System.getProperty("external.db.pool.cache.expiration.minutes",
                                                                                             "10"));
 
+    private DBConnectionPoolsHolder() {
+    }
+
     private static final Logger logger = Logger.getLogger(DBConnectionPoolsHolder.class);
 
     private final LoadingCache<DBConnectionDetails, HikariDataSource> dbConnectionPoolMap = CacheBuilder.newBuilder()
@@ -83,9 +86,10 @@ public class DBConnectionPoolsHolder {
      *
      * @param dbConnectionDetails
      * @param queryStr
-     * @return a cached result set of a given query to an external DB
+     * @return a cached result set for a given query to an external DB
      * @throws SQLException
      */
+
     public ResultSet executeQuery(DBConnectionDetails dbConnectionDetails, String queryStr) {
         ResultSet resultSet = null;
         try (Connection pooledConnection = dbConnectionPoolMap.get(dbConnectionDetails).getConnection();
@@ -113,7 +117,7 @@ public class DBConnectionPoolsHolder {
      *
      * @param dbConnectionDetails
      * @param updateStr
-     * @return @return either (1) the row count for SQL Data Manipulation Language (DML) statements
+     * @return either (1) the row count for SQL Data Manipulation Language (DML) statements
      *         or (2) 0 for SQL statements that return nothing
      */
     public int executeUpdate(DBConnectionDetails dbConnectionDetails, String updateStr) {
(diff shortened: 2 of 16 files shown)


scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SQLPooledConnections.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public class SQLPooledConnections {
+
+    private static Map<String, Connection> SQLPooledConnectionsMap = new HashMap<>();
+
+    public static void addConnection(String jobId, Connection connection) {
+        SQLPooledConnectionsMap.put(jobId, connection);
+    }
+
+    public static Connection getSQLPooledConnection(String jobId) {
```
amouhoub: maybe rename the method `getSQLPooledConnection` to `getConnection` : 
1) To cope with the `addConnection` method
2) Because the method will be used statically, therefore by calling SQLPooledConnections.getConnection (the class name already contains PooledConnections so no need to recall it in the method name)

Aminelouati: Thanks @amouhoub for the review. I will change the method name as you suggested.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SQLPooledConnections.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Map;
+
+
+/**
+ * @author ActiveEon Team
+ * @since 20/12/2018
+ */
+public class SQLPooledConnections {
+
+    private static Map<String, Connection> SQLPooledConnectionsMap = new HashMap<>();
```
Mykhailenko: you dont use any kind of synchronization.
I dont think HashMap is tread-safe. You can replace it with ConcurrentHashMap

Aminelouati: Thanks @Mykhailenko for the review. You are right. I will replace it with ConcurrentHashMap.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SQLPooledConnections.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static void addConnection(String jobId, Connection connection) {
```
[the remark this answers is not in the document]
amouhoub: @marcocast Good point, especially the last one about including tests in PRs to see the purpose of doing non-standard stuff
In fact, for this PR, there is a need for one of our prospects to use a pooled db connection within a proactive workflow across different tasks (think of a loop, replicate, ...). Therefore, we create a concurrent map that stores connection objects (say we need multiple pooled connections for different dbms/databases) and reuse them across the tasks of a job.
If you have a different proposal that surpasses the shortcomings of using statics then it is most welcome !!!
Thanks a lot !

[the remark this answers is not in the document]
fviale: @amouhoub you can use a singleton pattern (or lazy singleton) to avoid static methods everywhere.

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/SQLPooledConnections.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public class SQLPooledConnections {
+
+    private static Map<String, Connection> SQLPooledConnectionsMap = new ConcurrentHashMap<>();
+
+    public static void addConnection(String jobId, Connection connection) {
+        SQLPooledConnectionsMap.put(jobId, connection);
+    }
+
+    public static Connection getConnection(String jobId) {
```
fviale: I thought you were going to use a framework (hikari) to manage connection pools. A framework could make things a bit easier (because you need to handle many aspects of the db connection: idle time, connection timeouts, etc).

scheduler/scheduler-api/src/main/java/org/ow2/proactive/db/connectionpooling/DBConnectionPoolsHolder.java
```diff
@@ -0,0 +1,129 @@
...
+import com.zaxxer.hikari.HikariDataSource;
+
+
+/**
+ * @author ActiveEon Team
+ * @since 28/12/2018
+ */
+public class DBConnectionPoolsHolder {
+
+    private static final int CACHE_MAXIMUM_SIZE = Integer.parseInt(System.getProperty("external.db.pool.cache.maximum.size",
```
marcocast: a private constructor is missing

scheduler/scheduler-api/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     compile project(':common:common-api')
 
     compile 'org.apache.commons:commons-lang3:3.5'
+    compile group: 'com.zaxxer', name: 'HikariCP', version: '2.3.2'
```
marcocast: We can add this in the Parent Bom project

fviale: scheduling don't use parent-bom, (it is far too complicated to do so)

marcocast: ok

Aminelouati:
re-test this please

Aminelouati:
re-test this please

scheduler/scheduler-examples/src/main/java/org/ow2/proactive/scheduler/examples/connectionpooling/DBConnectionPoolsHolder.java
```diff
@@ -0,0 +1,130 @@
...
+                                                                                                        .expireAfterAccess(CACHE_EXPIRATION_MINUTES,
+                                                                                                                           TimeUnit.MINUTES)
+                                                                                                        .build(new CacheLoader<DBConnectionDetails, HikariDataSource>() {
+                                                                                                            public HikariDataSource
+                                                                                                                    load(DBConnectionDetails key)
+                                                                                                                            throws SQLException {
+                                                                                                                System.out.println("New connection to an external DB is created " +
```
marcocast: please replace System.out.println with a proper logger....maybe taskLogger?

fviale: We cannot as this is executed by the Task. Any logger.info will be printed to the Node log file and not to the task output.

So if we want to see this in the task output there is currently no other way than to do System.out.println

PR #3449, opened 2018-12-28
Lazy loading of finished jobs

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -1245,12 +1241,29 @@ public UserData getCurrentUserData() throws NotConnectedException {
         return PASchedulerProperties.getPropertiesAsHashMap();
     }
 
+    synchronized ClientJobState getClientJobState(JobId jobId) {
+        if (!jobsMap.containsKey(jobId)) {
+            List<InternalJob> internalJobs = dbManager.loadInternalJob(Arrays.asList(jobId.longValue()));
+            if (!internalJobs.isEmpty()) {
+                InternalJob internalJob = internalJobs.get(0);
+                ClientJobState clientJobState = new ClientJobState(internalJob);
+                jobsMap.put(jobId, clientJobState);
```
fviale: Beware that by doing jobsMap.put(), the job will be kept in memory without housekeeping

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerStateRecoverHelper.java
```diff
@@ -130,7 +130,7 @@ public RecoveredSchedulerState recover(long loadJobPeriod, RMProxy rmProxy) {
             }
         }
 
-        finishedJobs.addAll(dbManager.loadFinishedJobs(false, loadJobPeriod));
+        //        finishedJobs.addAll(dbManager.loadFinishedJobs(false, loadJobPeriod));
```
fviale: We shouldn't do this but use the property instead

Mykhailenko: Ok

Mykhailenko:
@sophiesongge you are right, i will change to make it true LRU

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -1245,12 +1253,33 @@ public UserData getCurrentUserData() throws NotConnectedException {
...
+    synchronized ClientJobState getClientJobState(JobId jobId) {
+        if (!jobsMap.containsKey(jobId)) {
+            if (!finishedJobsLRUCache.containsKey(jobId)) {
+                List<InternalJob> internalJobs = dbManager.loadInternalJob(jobId.longValue());
+                if (!internalJobs.isEmpty()) {
+                    InternalJob internalJob = internalJobs.get(0);
+                    ClientJobState clientJobState = new ClientJobState(internalJob);
+                    finishedJobsLRUCache.put(jobId, clientJobState);
+                }
+            }
```
sophiesongge: If the cache already contains the key, we need to remove it first, then add it to the head of the linkedhashmap

PR #3450, opened 2018-12-31
Allow job and task to share the same license

mbenguig:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -186,17 +187,30 @@ private boolean isTaskExecutableUsingLicense(EligibleTaskDescriptor task) {
...
 
                 final String currentRequiredLicense = requiredLicensesArray[i];
+
+                // If the license is already obtained at the job level, use it without getting a new one
+                LinkedBlockingQueue<JobDescriptorImpl> eligibleJobsDescriptorsLicense = eligibleJobsDescriptorsLicenses.get(currentRequiredLicense);
+                if (containsJob(eligibleJobsDescriptorsLicense,
+                                ((EligibleTaskDescriptorImpl) task).getInternal().getJobId())) {
+                    jobAlreadyUseLicense[i] = true;
+                    continue;
+                }
```
MaelAudren: I think this block of code could be moved to second iteration. It would avoid to execute that block when returning false. You could merge the two conditions to return false only when it is required.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -186,17 +187,30 @@ private boolean isTaskExecutableUsingLicense(EligibleTaskDescriptor task) {
 
             // Do not execute if one of the required license can not be obtained
             final String[] requiredLicensesArray = requiredLicenses.split(",");
-            for (int i = 0; i < requiredLicensesArray.length; i++) {
+            final int nbRequiredLicenses = requiredLicensesArray.length;
+            boolean[] jobAlreadyUseLicense = new boolean[nbRequiredLicenses];
+
+            for (int i = 0; i < nbRequiredLicenses; i++) {
 
                 final String currentRequiredLicense = requiredLicensesArray[i];
```
MaelAudren: The variable initialization should be done outside de loop.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -186,17 +187,30 @@ private boolean isTaskExecutableUsingLicense(EligibleTaskDescriptor task) {
...
                 if (!canGetTaskLicense(currentRequiredLicense)) {
                     logger.debug("License for " + currentRequiredLicense + " not available, keep task pending");
                     return false;
                 }
             }
             // Can be executed! Give it all required software licenses
-            for (int i = 0; i < requiredLicensesArray.length; i++) {
-                eligibleTasksDescriptorsLicenses.get(requiredLicensesArray[i]).add(task);
+            for (int i = 0; i < nbRequiredLicenses; i++) {
+                if (!jobAlreadyUseLicense[i])
```
MaelAudren: In the java convention a condition has always brackets, it improves the readability.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
mbenguig: Ok

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
mbenguig: Ok

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
mbenguig: We cannot shortcut the loop (return false if the task cannot take the current license) before testing before if  the job already got the current license. To improve the code readability, i removed the continue in the "if" and added a "else if" just after. But i decided to keep the result of the "if" (job already got the current license or not) for the next iteration, for not performing the test again before registering the task.

mbenguig: Finally i can simplify it following your suggestions. Thx!

scheduler/scheduler-server/src/test/java/functionaltests/policy/license/TestLicensePolicy.java
```diff
@@ -177,4 +177,39 @@ public void testLicensePolicyJobAndTaskLevelExclusiveSingleLicenseAtTaskLevel()
...
+    public void testTaskUseShareSameLicenseOfItsJobAndFinish() throws Throwable {
+
+        JobId jobId = schedulerHelper.submitJob(new File(JobSimpleJobAndTaskLicensePolicy.toURI()).getAbsolutePath(),
+                                                ImmutableMap.of("LICENSES_JOB",
+                                                                "software_A",
+                                                                "LICENSES_TASK",
+                                                                "software_A"));
+        log("Waiting for job finished");
+        schedulerHelper.waitForEventJobFinished(jobId);
+        Assert.assertTrue(true);
```
fviale: nice one :-)

PR #3451, opened 2019-01-03
add eventType name as value in the generic info

sophiesongge:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -144,7 +144,8 @@ public boolean doCheckAndSend(boolean withAttachment)
             logger.debug("Notification emails disabled, doing nothing");
             return false;
         }
-        if (!jobStatusList.contains(eventType.toString().toLowerCase()) || !jobStatusList.contains(eventType.name().toLowerCase())) {
+        if (!jobStatusList.contains(eventType.toString().toLowerCase()) ||
```
fviale: && !

PR #3453, opened 2019-01-04
increased the pa.scheduler.job.logs.max.size to 10MB

marcocast:


MaelAudren:
I highlight a big issue. If could fix it before merging, it would avoid big problem in Activeeon for next the years. Solving the issue may conduct to some refactoring. It it takes too much time I suggest you to fix the problem in another pull request. You may have to postpone you plans.

PR #3454, opened 2019-01-07
LoadPackage : create new revision when upgrading the scheduler

fviale:
When an existing installation of the scheduler is upgraded to a new version, the current behavior is that all catalog objects present already before the upgrade are not modified.

This is a problem as the new version may bring new improvements of the catalog workflows, patches, etc.

This commit changes the behavior and creates a new revision of the existing catalog object.

The administrator can then decide using the catalog portal if he wants to keep the update or rollback if necessary.

### DIFF
--- a/tools/LoadPackage.groovy
+++ b/tools/LoadPackage.groovy
@@ -22,6 +22,7 @@ class LoadPackage {
     private final String GLOBAL_SPACE_PATH
     private final String SCHEDULER_REST_URL
     private final String SCHEDULER_HOME
+    private final String SCHEDULER_VERSION
     private final File WORKFLOW_TEMPLATES_DIR
     private final String WORKFLOW_TEMPLATES_DIR_PATH
     private final String CATALOG_URL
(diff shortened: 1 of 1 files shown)


tools/LoadPackage.groovy
```diff
@@ -234,23 +251,26 @@ class LoadPackage {
...
+            def commitMessageEncoded = java.net.URLEncoder.encode(metadata_map.get("commitMessage") + " (" + SCHEDULER_VERSION + ")", "UTF-8")
 
-            def builder = org.apache.http.entity.mime.MultipartEntityBuilder.create()
-            builder.setBoundary(boundary);
-            builder.setMode(org.apache.http.entity.mime.HttpMultipartMode.BROWSER_COMPATIBLE)
-            builder.addPart("file", new org.apache.http.entity.mime.content.FileBody(object_file))
-            post.setEntity(builder.build())
+            // For POST queries
+            this.class.getClass().getResource(new File(this.SCHEDULER_HOME, "dist/lib/httpclient-4.5.2.jar").absolutePath);
+            def boundary = "---------------" + UUID.randomUUID().toString();
```
Mykhailenko: why do we generate boundary like this?

fviale: I have no idea honestly, I did not want to touch this bit

fviale:
Retest this please

PR #3462, opened 2019-01-17
add dockerfile.properties

cdiop:


config/scriptengines/dockerfile.properties
```diff
@@ -0,0 +1,6 @@
+docker.file.command=/usr/local/bin/docker
```
fviale: this will work only on linux

docker.file.command=docker 

would work on both linux and windows

fviale: could you also change the docker-compose.properties accordingly ?

fviale: I'm not 100% sure that using simply "docker" will work on linux, even more so "docker-compose" for the other script engine. But I know for sure that it will not work on windows.

See the following commit I did to fix this:
https://github.com/ow2-proactive/scriptengine-docker-compose/commit/4eb054210f8620f31aab95efa38206d2eba55633

A possibility could be to have two properties:
docker.file.command=/usr/local/bin/docker
docker.file.command.windows=docker

@cdiop @mboussaa What do you think ?

Mykhailenko:
Retest this please

MaelAudren:
@cdiop what is the status of this pull request ?

PR #3463, opened 2019-01-21
update docker command

cdiop:


cdiop:
@MaelAudren It is better to have one config file per scriptengine

PR #3475, opened 2019-02-12
Add IAM filters to jetty webapp contexts (contexts of PA microservices)

nebilbenmabrouk:
- JettyStarter updated. It adds IAM authentication and authorization filters to the web contexts of some PA microservices (namely, the studio, scheduler, rm and automation-dashboard).
- keystore updated. It contains a SSL certificate signed by the Certification Authority (ActiveeonCA). The root certificate of ActiveeonCA (called 'ActiveeonRootCA.crt') is also added under config/web.
- Slight change in IAM starter the parameter 'PA_SERVER_NAME' is no longer hard-coded.


scheduler/scheduler-server/src/main/java/org/ow2/proactive/utils/JettyStarter.java
```diff
@@ -377,6 +399,84 @@ private WebAppContext createWebAppContext(String contextPath, String[] virtualHo
...
+                             EnumSet.of(DispatcherType.REQUEST,
+                                        DispatcherType.INCLUDE,
+                                        DispatcherType.FORWARD,
+                                        DispatcherType.ASYNC,
+                                        DispatcherType.ERROR));
+
+            // Add CAS HttpServletRequest Wrapper Filter
+            webApp.addFilter(org.jasig.cas.client.util.HttpServletRequestWrapperFilter.class,
+                             "/*",
+                             EnumSet.of(DispatcherType.REQUEST,
```
Mykhailenko: These EnumSet repeat 4 times. Maybe you can re-use it

nebilbenmabrouk: done. Thanks

scheduler/scheduler-node/src/test/java/functionaltests/ForkedTaskExecutorTest.java
```diff
@@ -324,6 +326,17 @@ private Decrypter createCredentials(String username) throws NoSuchAlgorithmExcep
 
     @Before
     public void setUp() throws Exception {
+
```
fviale: normally all tests are started in gradle with the correct property paths set. Did you have an issue while running the test in intellij ? If yes, then you simply need to add these properties as test jvm parameters rather than change the code.

nebilbenmabrouk:
re-test this please

PR #3480, opened 2019-02-21
Delay shutdown when remove or undeploy NS

Mykhailenko:


fviale:
after looking at the code, it seems quite difficult to double check shutdown has been called inside a test

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -2567,7 +2603,8 @@ private void validateNodeSourceNameOrFail(String nodeSourceName) {
         }
         if (this.definedNodeSources.containsKey(nodeSourceName) ||
             this.deployedNodeSources.containsKey(nodeSourceName) ||
-            this.delayedNodeSourceRemovalEvents.containsKey(nodeSourceName)) {
+            this.delayedNodeSourceRemovalEvents.containsKey(nodeSourceName) ||
```
fviale: this check is a bit strange, is it meant to handle the following scenario ?

1) user removes a ns, waiting on busy nodes
2) It takes too long, user try again to remove the ns.

Mykhailenko: @fviale did not get your point. This check here is to ensure we dont create new NS with a name that is already exist

fviale: sorry, code folding misunderstanding

Issue #3469, opened 2019-01-29
running script on nodes can fail with already locked message


fviale:
When trying to run a script on a node source, the following error message appears

Error: java.lang.IllegalStateException: Node script cannot be executed atomically since the node is already locked: pnp://assoopy.mgps.inra.fr:34921/assoopy__assoopy__4_1
Caused by: Node script cannot be executed atomically since the node is already locked: pnp://assoopy.mgps.inra.fr:34921/assoopy__assoopy__4_1

The node does not appear as locked in the RM portal interface.
Unlocking all nodes of the node source does not solve the issue.

Selecting another node source shows the same error message.

Retrying several times still show the same message. No job was running during the test. Node source recovery ws disabled.

PR #3481, opened 2019-02-21
Fix node locking (not actually)

Mykhailenko:
Issue: https://github.com/ow2-proactive/scheduling/issues/3469
No idea how to fix it. Just make nodes unlocking safer.


fviale:
@Mykhailenko your pull request contain unrelated changes

Mykhailenko:
@fviale yes you are right. Fixing it

PR #3482, opened 2019-02-22
Fix node locking (not actually)

Mykhailenko:
Issue https://github.com/ow2-proactive/scheduling/issues/3469

This PR just make node locking/unlocking safer.

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
@@ -2709,7 +2709,7 @@ public class RMCore implements ResourceManager, InitActive, RunActive {
 
                 result &= operation.apply(rmnode);
             } catch (Exception e) {
-                logger.error("Exception during mapOnNodeUrlSet.", e);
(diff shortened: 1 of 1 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -2696,27 +2691,27 @@ boolean internalLockNode(RMNode rmNode, Client lockInitiator) {
...
+                if (rmnode == null) {
+                    logger.warn("Cannot " + operationName + ", unknown node: " + url);
+                    result = false;
+                    continue;
+                }
 
-            result &= operation.apply(rmnode);
+                result &= operation.apply(rmnode);
+            } catch (Exception e) {
+                logger.error("Exception during mapOnNodeUrlSet.", e);
```
fviale: it would be better to log a similar error message as the above warning, with the exception

fviale:
```suggestion
                logger.error("Error during " + operationName + " on node " + url, e);
```

PR #3487, opened 2019-03-01
Enhance jetty performance

nebilbenmabrouk:
- JettyStarter uses the class 'QuickStartWebApp' to enhance the timeliness of deploying PA microservices

- Atmosphere Framework (used in the 'rest' webapp) is configured to improve its initialization time


Mykhailenko:
So cool, can you tell how much faster it is going?

Issue #3494, opened 2019-03-05
RMNodeStarter: after reconnecting to a resource manager which has been restarted, JMX authentication does not work


fviale:
JMXAuthentication tries to connect to the resource manager using the previous reference of the resource manager. When the resource manager has been restarted, it is not valid any more.

Accordingly, many operations on the node do not work any more (node locking, node thread dump, etc).

fviale:
fixed

PR #3495, opened 2019-03-05
Fix for #3494 : In RMNodeStarter, after reconnecting to a resource manager which has been restarted, JMX authentication does not work

fviale:
This fix does the following:

 - shutdown and restart JMX monitoring upon resource manager reconnection
 - store various information such as nodes and resource manager interface as attributes
 - sanitize reconnection logging messages


rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/RMNodeStarter.java
```diff
@@ -1209,17 +1207,32 @@ private ResourceManager loginToResourceManager(final Credentials credentials, fi
...
+        ResourceManager rm = loginToResourceManager(credentials, rmAuth);
 
-        RMAuthentication auth = joinResourceManager(rmURL);
-        final ResourceManager rm = loginToResourceManager(credentials, auth);
-        SigarExposer sigarExposer = null;
+        startMonitoring(rmAuth);
 
+        for (final Node node : nodes) {
+            nodeSetJmxUrl(sigarExposer, node);
+            addNodeToResourceManager(rmURL, node, rm);
```
Mykhailenko: Question: can `addNodeToResourceManager` throw some kind of exception, so we will skip next calls of `addNodeToResourceManager`, if so, maybe, we can catch them?

rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/RMNodeStarter.java
```diff
@@ -504,8 +514,8 @@ public boolean createNodesAndConnect(final String nodeName) {
         return createdNodeNames;
     }
 
-    private Map<String, Node> createNodes(String nodeName) {
-        Map<String, Node> nodes = new HashMap<>(workers);
+    private ConcurrentHashMap<String, Node> createNodes(String nodeName) {
```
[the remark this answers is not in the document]
fviale: Yes, but I was thinking precisely to make the return type explicit as to know it's a concurrent version.

rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/RMNodeStarter.java
fviale: Normally, no, this would be an abnormal behavior. In that case, the global try catch in the main method will exit the JVM with a clear error message.

PR #3501, opened 2019-03-06
Install tools needed by the scripts automatically

fviale:
The installation script needs a few system tools (rsync, zip, unzip, git) which are not always installed by default.

This change asks the user to confirm that these tools will be installed

### DIFF
--- a/tools/install_base.sh
+++ b/tools/install_base.sh
@@ -205,6 +205,45 @@ config/web/truststore
     cd "$OLD_PWD"
 }
 
+install_binaries()
+{
+    if ! which rsync > /dev/null 2>&1; then
+        echo "rsync is not installed on this computer and is required by the ProActive installation."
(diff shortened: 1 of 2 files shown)


tools/install_base.sh
```diff
@@ -205,6 +205,45 @@ config/web/truststore
...
+        echo "unzip is not installed on this computer and is required by the ProActive installation."
+        if confirm "Do you want to install it? [Y/n] " ; then
+           $PKG_TOOL -y install unzip
+        else
+           echo "Installation aborted."; exit 1 ;
+        fi
+    fi
+
+    if ! which git > /dev/null 2>&1; then
+        echo "Git is not installed on this computer and is required by the ProActive installation."
```
Mykhailenko: all names starts with small but Git starts with capital

PR #3510, opened 2019-03-14
Json model

mklkun:
In this pull request we:
- introduce a new variable model PA:JSON for variables with valid JSON objects/arrays,
- add relevant unit tests.

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
@@ -42,7 +42,7 @@ import org.ow2.proactive.scheduler.common.job.factories.spi.model.exceptions.Val
  */
 public class JSONValidator implements Validator<String> {
 
-    public static final String JSON_EMPTY_MODEL = "^[/{][/}]$";
+//    public static final String JSON_NEEDED_CAR_REGEX = "[0-9a-zA-Z\{\}\":]";
 
     public JSONValidator() {
         /**
@@ -53,14 +53,11 @@ public class JSONValidator implements Validator<String> {
     @Override
     public String validate(String parameterValue, ModelValidatorContext context) throws ValidationException {
 
-        Pattern pattern = Pattern.compile(JSON_EMPTY_MODEL);
-        Matcher matcher = pattern.matcher(parameterValue);
-
         try {
-            if ((parameterValue.matches(JSON_EMPTY_MODEL) && matcher.find()) || !isValidJSON(parameterValue)) {
+            if (!isValidJSON(parameterValue)) {
                 throw new ValidationException("Expected value should match JSON format, received " + parameterValue);
             }
-        } catch (Exception er) {
+        } catch (ValidationException er) {
             throw new ValidationException(er);
         }
         return parameterValue;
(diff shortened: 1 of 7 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
```diff
@@ -0,0 +1,80 @@
...
+         * ProActive Empty constructor.
+         */
+    }
+
+    @Override
+    public String validate(String parameterValue, ModelValidatorContext context) throws ValidationException {
+
+        try {
+            if (!isValidJSON(parameterValue)) {
+                throw new ValidationException("Expected value should match JSON format, received " + parameterValue);
```
fviale: I think this part should be improved.

isValidJSON returns a boolean and also throws a ValidationException. You should rather decide on a single strategy, either return true false or return void and throw an exception.

I prefer the latter, as an exception contains more meaningful error messages than true/false.

In that case the method should be renamed checkValidJson

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
```diff
@@ -0,0 +1,80 @@
...
+            }
+        } catch (ValidationException er) {
+            throw new ValidationException("Validator error from JsonParser: Expected value should match JSON format, received " +
+                                          parameterValue);
+        }
+        return parameterValue;
+    }
+
+    public boolean isValidJSON(final String json) throws ValidationException {
```
fviale: this method should be private as it's only used in the current class as far as I see

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
```diff
@@ -0,0 +1,80 @@
...
+        boolean valid;
+        try {
+            final JsonParser parser = new ObjectMapper().getJsonFactory().createJsonParser(json);
+            while (parser.nextToken() != null) {
+            }
+            valid = true;
+        } catch (JsonParseException jpe) {
+            throw new ValidationException("Validator error for JSON type : " + jpe);
```
fviale: it is better to catch Exception here rather than specific (are we sure only checked exception will be thrown by jackson ?

additionnally, you must chain the exception cause correctly:

throw new ValidationException("Validator error for JSON type", jpe);

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
mklkun: I agree. I thought that I had to leave the content of isValidJSON as faithful as possible to the version you sent us.
I will fix this.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
mklkun: Ok.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/JSONValidator.java
mklkun: Same here. I will fix that.
Thanks!

mklkun:
Great. Thank you!

fviale:
You will need a second review (don't hesitate to ask).

Before you merge, you need to (I suggest to do it in this order).

1) squash your commit:  you can use git rebase -i (and squash commits together), and then git push -f to push force on your fork

2) synchronize your fork with the master (you do this by creating a pull request from ow2-proactive/scheduling to your fork, ask around if you're not sure how to do this)

3) wait for the test to finish

4) merge

mklkun:
Thanks Amine!

PR #3529, opened 2019-04-12
New shiny ProgressFileReaderPoller 

Mykhailenko:


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/ProgressFileReaderPoller.java
```diff
@@ -0,0 +1,222 @@
...
+                                  UUID.randomUUID() + ".progress";
+
+        return start(workingDir, progressFileName);
+    }
+
+    boolean start(File workingDir, String filename) {
+        try {
+            createProgressFile(workingDir, filename);
+
+            watchDogThread = new Thread(new ProgressFileReaderThread(Duration.ofMillis(PASchedulerProperties.SCHEDULER_PROGRESS_FILE_READER_POLL_TIMEOUT.getValueAsInt())));
```
fviale: you cannot use PASchedulerProperties in this context as this code is executed on the node.

fviale: why not use ScheduledExecutorService.scheduleAtFixedRate instead of Thread ?

config/scheduler/settings.ini
```diff
@@ -402,4 +402,4 @@ pa.scheduler.synchronization.db=data/synchronization
 #-------------------------------------------------------
 #----------------  PORTAL PROPERTIES  ------------------
 #-------------------------------------------------------
-pa.scheduler.portal.configuration=config/portal/scheduler-portal-display.conf
+pa.scheduler.portal.configuration=config/portal/scheduler-portal-display.conf
```
fviale: could you remove this new line change ?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/ProgressFileReaderPoller.java
```diff
@@ -0,0 +1,201 @@
...
+            progress = 0;
+
+            return true;
+        } catch (IOException e) {
+            logger.warn("Error while creating progress file. Progress will not be reported.", e);
+            return false;
+        }
+    }
+
+    private void createProgressFile(File workingDir, String progressFileName) throws IOException {
```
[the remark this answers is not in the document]
Mykhailenko: i reused the code from the ProgressFileReader. I think it is right thing to do. We dont want to remove file if it exists neither it is a bug to call createProgressFile method when file already exists

PR #3538, opened 2019-04-19
Add the possibility to exclude users or groups from node sources

mbenguig:


common/common-client/src/main/java/org/ow2/proactive/permissions/PrincipalPermission.java
```diff
@@ -151,11 +211,41 @@ public void add(Permission permission) {
...
-        for (Permission p : permissions) {
-            if (p.implies(permission)) {
-                return true;
+
+        if (!(permission instanceof PrincipalPermission)) {
+            return false;
+        }
+
+        PrincipalPermission permissionToRespect = (PrincipalPermission) permission;
+        if (permissionToRespect.principals.isEmpty())
```
Mykhailenko: could you please put `{`

common/common-client/src/main/java/org/ow2/proactive/permissions/PrincipalPermission.java
```diff
@@ -151,11 +211,41 @@ public void add(Permission permission) {
...
+
+        int nbPrincipals = permissionToRespect.principals.size();
+        long nbExcludedPrincipals = permissionToRespect.principals.stream()
+                                                                  .filter(c -> c instanceof NotIdentityPrincipal)
+                                                                  .count();
+
+        if (nbExcludedPrincipals == 0) {
+            // Give access if INCLUSION permissions to respect include/implie one of the permission requiring access
+            // ex: "Bob" or "GroupBob" get access if "Bob" or "GroupBob" is part of permissions to respect
+            for (Permission p : permissions) {
```
Mykhailenko: you can use `Stream::anyMatch` method

common/common-client/src/main/java/org/ow2/proactive/permissions/PrincipalPermission.java
mbenguig: sure

common/common-client/src/main/java/org/ow2/proactive/permissions/PrincipalPermission.java
mbenguig: sure

PR #3543, opened 2019-05-03
Rename filters

Mykhailenko:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/task/TaskStatus.java
```diff
@@ -175,9 +175,11 @@ public String toString() {
                     return Stream.of(TaskStatus.PENDING);
                 case "running":
                 case "active":
```
amouhoub: @Mykhailenko Maybe remove `active` and `terminated` as they are not valid and used anymore (plus they are semantically wrong)

PR #3545, opened 2019-05-03
(WILL BE REMOVED SOON) Housekeeping performance

Mykhailenko:
- Housekeeping takes a lot of time due to DB loading
 - this PR removes heavy DB loading
 - functional tests are still missing (there are no test which covers housekeeping)

### DIFF
--- a/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/core/SchedulerStateImpl.java
+++ b/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/core/SchedulerStateImpl.java
@@ -25,23 +25,30 @@
  */
 package org.ow2.proactive.scheduler.core;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.Vector;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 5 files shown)


scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/core/SchedulerStateImpl.java
```diff
@@ -364,17 +374,22 @@ public synchronized void pendingToRunning(T js) {
...
     }
 
     public synchronized void runningToFinished(T js) {
         runningJobs.remove(js);
-        finishedJobs.add(js);
+        finishedJobs.put(js.getId(), js);
     }
 
     public synchronized void removeFinished(T js) {
         finishedJobs.remove(js);
```
fviale: it looks like we should remove js.getId()

scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/core/SchedulerStateImpl.java
```diff
@@ -60,13 +68,13 @@
 public final class SchedulerStateImpl<T extends JobState> implements SchedulerState<T> {
 
     /** Pending jobs */
-    private Set<T> pendingJobs = Collections.synchronizedSet(new LinkedHashSet<T>());
+    private Set<T> pendingJobs = new ConcurrentHashSet<>();
```
fviale: by using ConcurrentHashSet we remove the insertion iteration order, which changes the orders of vectors returned by getPending, getRunning, etc.

PR #3551, opened 2019-05-21
Improve error message when server can not connect to node

ShatalovYaroslav:


rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/RMNodeStarter.java
```diff
@@ -1178,7 +1177,8 @@ private RMAuthentication joinResourceManager(String rmURL) {
             logger.info("Resource Manager joined.");
             return auth;
         } catch (Throwable t) {
-            logger.error("Unable to join the Resource Manager at " + rmURL, t);
+            logger.error("Unable to join the Resource Manager at " + rmURL
+            + "\n ProActive server can not send reply to Node (usually due to firewall). Please check server log for more information", t);
```
mboussaa: ...can not request the Node...?

aminelouati: cannot connect to the Node

fviale: it's not here that the error message should be changed

ShatalovYaroslav:
retest it please

ShatalovYaroslav:
test this please

MaelAudren:
@ShatalovYaroslav can you squash your commit before merging ?

PR #3552, opened 2019-05-21
Avoid deadlocks and heavy CPU usage

fviale:
- use read/write locks instead of synchronized
 - use Thread.sleep to avoid aggressive loop
 - use ArrayBlockingQueue instead of LinkedBlockingQueue (minor)
 - use fairness to respect FIFO thread access to the queue.

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/utils/appenders/AsynchChachedFileAppender.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/utils/appenders/AsynchChachedFileAppender.java
@@ -42,19 +42,21 @@ public class AsynchChachedFileAppender extends AsynchFileAppender {
     @Override
     public void append(String cacheKey, LoggingEvent event) {
         try {
-            synchronized (queue) {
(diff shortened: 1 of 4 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/utils/appenders/AsynchChachedFileAppender.java
```diff
@@ -42,26 +42,30 @@
     @Override
     public void append(String cacheKey, LoggingEvent event) {
```
amouhoub: Maybe add `@SuppressWarnings("squid:S2142")` since we purposely chose to not reiterate the Interruption state and/or exception

rm/rm-server/src/main/java/org/ow2/proactive/utils/appenders/AsynchFileAppender.java
```diff
@@ -52,18 +60,24 @@ public AsynchFileAppender() {
 
     // non blocking
     public void append(String cacheKey, LoggingEvent event) {
```
amouhoub: Same as above

PR #3554, opened 2019-05-22
upgrade resteasy to 3.0.26

fviale:
last upgrade did not solve the memory leak issue due to mime4j temporary file. The version upgraded for rest easy libraries was the wrong version. https://issues.jboss.org/browse/RESTEASY-1467
 Version 3.0.21 contains the fix instead of 3.0.19


fviale:
Retest this please

PR #3564, opened 2019-05-31
encode third part credential key on delete

aminelouati:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/ThirdPartyCredentials.java
```diff
@@ -63,5 +63,5 @@ void putThirdPartyCredential(String key, String value)
      * @throws NotConnectedException if you are not authenticated.
      * @throws PermissionException if you can't access this particular method.
      */
-    void removeThirdPartyCredential(String key) throws NotConnectedException, PermissionException;
+    void removeThirdPartyCredential(String key) throws NotConnectedException, PermissionException, KeyException;
```
fviale: I don't understand why you need to add KeyException everywhere

aminelouati: The decode method added to SchedulerStateRest throws an Exception thus I needed to change the signature of the super method. If not, a compilation error is arisen.

fviale: in SchedulerStateRest, the new exception that you throw is a SchedulerRestException and not a KeyException.

aminelouati: Yes, you are right. However, similar to `putThirdPartyCredential` method, in SchedulerClient we catch SchedulerRestException and throw KeyException to avoid adding this dependency everywhere. 

```
import org.ow2.proactive_grid_cloud_portal.scheduler.exception.SchedulerRestException;
```
But if you prefer that, I will do so.

aminelouati: I need to add this dependency to build.gradle of scheduler-api module

aminelouati: which implies a circular dependency....

PR #3568, opened 2019-06-06
Rationalize some error messages

fviale:
- avoid using printStackTrace whenever possible

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/InfrastructureManagerFactory.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/InfrastructureManagerFactory.java
@@ -49,6 +49,8 @@ public class InfrastructureManagerFactory {
     /** list of supported infrastructures */
     private static Collection<Class<?>> supportedInfrastructures;
 
+    private static final Logger logger = Logger.getLogger(InfrastructureManagerFactory.class);
+
     /**
      * Creates new infrastructure manager using reflection mechanism.
      *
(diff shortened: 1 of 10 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/policy/NodeSourcePolicyFactory.java
```diff
@@ -136,7 +139,7 @@ public static NodeSourcePolicy activate(NodeSourcePolicy policy, Object[] policy
                 properties.load(stream);
             }
         } catch (Exception e) {
-            e.printStackTrace();
+            logger.error("Error when reading node source policies in file : " + propFileName, e);
```
Mykhailenko:
```suggestion
            logger.error("Error when reading node source policies in file: " + propFileName, e);
```

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/InfrastructureManagerFactory.java
```diff
@@ -122,7 +125,7 @@ public static InfrastructureManager recover(NodeSourceDescriptor nodeSourceDescr
                 Class<?> cls = Class.forName(className.toString());
                 supportedInfrastructures.add(cls);
             } catch (ClassNotFoundException e) {
-                Logger.getLogger(InfrastructureManagerFactory.class).warn("Cannot find class " + className.toString());
+                logger.error("Cannot find infrastructure class " + className.toString());
```
Mykhailenko: `className` is not String?

PR #3570, opened 2019-06-11
Delete badSelectionScript.js

mhow2:
Not a valid javascript file. I suggest to delete it.


mhow2:
I realize it could be intended. Sorry about this.

PR #3574, opened 2019-06-24
Robust endpoint for resultMap and preciousResults

Mykhailenko:
- avoid sending empty maps
- Fix NPE

### DIFF
--- a/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
+++ b/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
@@ -540,7 +540,7 @@ public class SchedulerStateRest implements SchedulerRestInterface {
     }
 
     // request:  http://localhost:8080/rest/scheduler/jobs/resultmap?jobsid=102&jobsid=152
-    // response: {"102": {"var1" : "val1", "var2" : "val2"},"152": {}}
+    // response: {"102": {"var1" : "val1", "var2" : "val2"}}
     @Override
     public Map<Long, Map<String, String>> jobResultMaps(String sessionId, List<String> jobsId)
             throws NotConnectedRestException, UnknownJobRestException, PermissionRestException {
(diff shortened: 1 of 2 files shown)


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -540,7 +540,7 @@ public JobResultData jobResult(@HeaderParam("sessionid") String sessionId, @Path
     }
 
     // request:  http://localhost:8080/rest/scheduler/jobs/resultmap?jobsid=102&jobsid=152
-    // response: {"102": {"var1" : "val1", "var2" : "val2"},"152": {}}
+    // response: {"102": {"var1" : "val1", "var2" : "val2"}}
```
amouhoub: 1) How can one check  if all his job ids have or don't have resultMaps ? this will add in the calling-code a test for all ids sent for this endpoint to see if they have a result or not. I'm not sure that this is the proper way to do this.

2) It is not a good idea to use this endpoint the way it is now with multiple job ids because HTTP/GET calls are limited to 2000 chars. Therefore, if you have 6 figures job id (say 10100, 10101, etc...) and you send a query for 100 jobs, your url will be at least:(`"&jobsid=".length` = 8) 
`(8+6)*100` = 1400 !,
At Komatsu this will break in less than a week 😢 
If you cannot expect an HTTP/GET max length as is the case here they you should do an HTTP/POST

Mykhailenko: Wow, thanks
can you send a link to `limited ti 2000 chars`

amouhoub:
I have one other side-comment regarding the expected usage of the new/modified endpoints:
ResultMap is a string that can bear up to 200mb (average observed max with default db) and the Machine Learning workflows are using them to store long JSON strings.
So if for all "past jobs" and just for the sake of displaying or not the map/download icon we get all resultMaps before the user asks for them (I believe they are displayed on-click only in a modal), this would still be heavy at a scale and this would lead to either : store in memory all the resultsMaps to avoid asking for them again or drop the results of this call once we display the icons then ask for them again when the user clicks.
Just saying 🙄 ...

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
amouhoub: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
I remember that Job Planner (during @paraita's days) had an issue that caused this and made the job planner fail at scale

fviale: 1) Alternatively, the endpoint could return hasResultMap=true/false but in that case, from a pure REST API perspective, the endpoint has no great value. Here it gives the opportunity to receive all in a single call.
2) the intent of this endpoint is to restrict it to a small number of jobids (<= 100).

This is why it is implemented using GET and it matches the workflow automation portal requirement which calls this on less than 100 jobs.

If ever in the future we really need to perform this call on a large number of jobs, then we will probably need to implement it via POST (we can actually have two endpoints with the same path and different methods)

fviale:
@amouhoub why the hell are you using result map to store such big strings. It has never been designed for this usage.

amouhoub:
@fviale I didn't say that we use it this way and I know that we don't for a fact. We just use it to store some average son objects.
I'm saying that we're downloading twice at least the same info or storing it for in memory just because we need to decide wether or not to display an icon.
I would suggest using a flag, not in these 2 endpoints because that would be pointless, but to add the flags to the jobInfoList endpoint results as we already call it to get everything we need here except the results. So why not just include all the info in one place as we need it all together now.
As per for this PR, the provided endpoints are great for getting results massively if we ever need to do that in one-shot calls in some other portal/etc (not recurrently as in wf automation)

PR #3575, opened 2019-06-26
Add support for HTTPS to SchedulerRestClient

nebilbenmabrouk:
Enable SchedulerRestClient to interact with Scheduler in SSL mode

### DIFF
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
@@ -122,7 +122,7 @@ public class SchedulerRestClient {
             sslContext = SSLContexts.custom().loadTrustMaterial(null, acceptingTrustStrategy).build();
 
         } catch (KeyStoreException | KeyManagementException | NoSuchAlgorithmException e) {
-            e.printStackTrace();
+            throw new RuntimeException(e.getMessage());
         }
     }
 
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
@@ -31,11 +31,14 @@ import java.io.*;
 import java.lang.reflect.*;
 import java.net.HttpURLConnection;
 import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
+import java.security.KeyManagementException;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
+import javax.net.ssl.*;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.Produces;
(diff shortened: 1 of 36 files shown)


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
```diff
@@ -74,17 +84,31 @@
...
+
     public SchedulerRestClient(String restEndpointURL) {
         this(restEndpointURL, null);
     }
 
     public SchedulerRestClient(String restEndpointURL, ClientHttpEngine httpEngine) {
-        this.restEndpointURL = restEndpointURL;
+
+        if (sslContext == null) {
+            setBlindTrustSSLContext();
```
fviale: should it not be inside the following if block ? Because it will always be null here.

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
```diff
@@ -74,17 +84,31 @@
...
+        }
+
```
fviale: alternatively you can do:
this.httpEngine = httpEngine;

if (this.httpEngine == null) {

this.httpEngine = ...
}

Not a big deal, but modifying parameters should be avoided.

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
nebilbenmabrouk: You're right. Fixed, thanks

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
```diff
@@ -95,6 +119,20 @@ public SchedulerRestClient(String restEndpointURL, ClientHttpEngine httpEngine)
         scheduler = createRestProxy(providerFactory, restEndpointURL, httpEngine);
     }
 
+    private void setBlindTrustSSLContext() {
+        try {
+            TrustStrategy acceptingTrustStrategy = (cert, authType) -> true;
+            sslContext = SSLContexts.custom().loadTrustMaterial(null, acceptingTrustStrategy).build();
+
+        } catch (KeyStoreException | KeyManagementException | NoSuchAlgorithmException e) {
+            e.printStackTrace();
```
fviale: What about throwing a runtime exception instead of printStackTrace ?

fviale: This would imply to change the calls like:

org.ow2.proactive_grid_cloud_portal.cli.cmd.sched.FreezeCommand

```java
   @Override
    public void execute(ApplicationContext currentContext) throws CLIException {
        SchedulerRestInterface scheduler = currentContext.getRestClient().getScheduler();
        try {
       } catch (Exception e) {
            handleError("Error occurred while trying to freeze the scheduler:", e, currentContext);
        }
    }
```
to
```java
   @Override
    public void execute(ApplicationContext currentContext) throws CLIException {
   
        try {
      SchedulerRestInterface scheduler = currentContext.getRestClient().getScheduler();

       } catch (Exception e) {
            handleError("Error occurred while trying to freeze the scheduler:", e, currentContext);
        }
    }
```

fviale: the problem is that e.printStackTrace() will most likely go to /dev/null :-)

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
```diff
@@ -95,6 +116,20 @@ public SchedulerRestClient(String restEndpointURL, ClientHttpEngine httpEngine)
         scheduler = createRestProxy(providerFactory, restEndpointURL, httpEngine);
     }
 
+    private void setBlindTrustSSLContext() {
+        try {
+            TrustStrategy acceptingTrustStrategy = (cert, authType) -> true;
+            sslContext = SSLContexts.custom().loadTrustMaterial(null, acceptingTrustStrategy).build();
+
+        } catch (KeyStoreException | KeyManagementException | NoSuchAlgorithmException e) {
+            throw new RuntimeException(e.getMessage());
```
fviale: throw new RuntimeException(e.getMessage(),e);

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
nebilbenmabrouk: I prefer keeping it like that to avoid adding an else block.
Actually, I modified a null parameter.

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
nebilbenmabrouk: Done.

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
fviale: ok

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/SchedulerRestClient.java
fviale: you need to change calls to currentContext.getRestClient() according to my comment to avoid uncatched exceptions

nebilbenmabrouk: Done.

PR #3584, opened 2019-07-15
Add the task parameter fork 

tao-xinxiu:
Support to configure whether the task is run in a forked JVM for each task.

Global configuration should override the task parameter. 

In case of conflicts:
 - conflict between global configuration of runAsMe and fork throws a `SchedulerConfigurationException` to stop the scheduler to be started
 - conflict between global configurations and task parameters cause a warning in the job log
 - conflict between the task parameters runAsMe and fork throws a `JobCreationEXCEPTION` to stop the job to be submitted


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/task/Task.java
```diff
@@ -778,4 +802,45 @@ public void setForkEnvironment(ForkEnvironment forkEnvironment) {
...
     }
 
+    /**
+     * Returns whether the task should be run in a forked JVM, depending on the global configuration and task property.
+     * The global configuration has a higher priority than the task property; And runasme true implies fork true.
+     *
+     * @return whether the task should be run in a forked JVM.
+     */
+    public boolean isForkingTask() {
+        if (PASchedulerProperties.TASK_RUNASME.getValueAsBoolean()) {
```
fviale: this is part of the scheduler-api module which is not necessarily used server-side

tao-xinxiu: I just moved this part (i.e., calculating task fork and runAsMe based on its parameter and global config) into a separated class in `scheduler-server/.../util/TaskConfiguration.java`.

p.s. I didn't move it into `internalTask`, because some of these functions are called before the internalTask is initialized.

PR #3588, opened 2019-07-18
add license report plugin in build.gradle

MaelAudren:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue May 10 09:48:49 CEST 2016
+#Thu Jul 18 15:36:59 CEST 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.13-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-bin.zip
```
aminelouati: I am wondering if we should upgradegradle-wrapper  the other project consequently ?

MaelAudren: Indeed I am upgrading the gradle wrapper of every project. Moreover I can't merge this pull request until I upgraded release gradle-wrapper otherwise when building Proactive it will be compiled using relase gradle version and it will fail.

aminelouati: Thanks for the clear response

PR #3591, opened 2019-07-25
Improve Node Source permission handling

fviale:
- created new permission for full rm core rights but which does not override node restrictions
 - created a new unused permission to override node restrictions (just in case)
 - created a new group for scheduler full permission
 - "admin" and "demo" users now belong to rmcore and scheduler admin group instead of having "AllPermissions". This allows to enforce node usage restrictions on those accounts
 - improved tests in rm-server/permissions


config/authentication/group.cfg
```diff
@@ -1,8 +1,10 @@
-admin:admin
-admin:nsadmins
-demo:admin
+admin:scheduleradmins
+admin:rmcoreadmins
+demo:scheduleradmins
+demo:rmcoreadmins
 guest:guests
 nsadmin:nsadmins
+nsadmin2:nsadmins
```
mbenguig: Thanks to you, we have now well defined groups: scheduleradmins, rmcoreadmins, nsadmins...  with well defined members: admin, demo... but cant these latter ones replace nsadmin and nsadmins users?

fviale: @mbenguig nsadmins users are only allowed to control node sources, not the full resource manager

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/authentication/Client.java
```diff
@@ -258,12 +259,32 @@ public boolean checkPermission(final Permission permission, String errorMessage)
                 return null;
             }, null);
         } catch (SecurityException ex) {
-            throw new SecurityException(errorMessage, ex);
+            if (!checkSuperUserPermission(superUserPermissions)) {
```
mbenguig: A single-line comment would have been apreciated ..

fviale: ok

fviale:
Retest this please

PR #3595, opened 2019-07-31
minor [forkTask]: add function to convert task fork from taskType in TaskData

tao-xinxiu:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskData.java
```diff
@@ -1227,4 +1225,14 @@ public TaskState toTaskState() {
         return taskState;
     }
 
+    @Transient
+    public boolean isForkTask() {
+        if (taskType.equals(SCRIPT_TASK)) {
```
Mykhailenko: if i am not totally sure that String is not null, then i do:
```suggestion
        if (SCRIPT_TASK.equals(taskType)) {
```

tao-xinxiu: Good point, just applied.

PR #3598, opened 2019-08-01
No need creds

Mykhailenko:


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/BatchJobInfrastructure.java
```diff
@@ -496,7 +497,7 @@ public void configure(Object... parameters) {
                 persistedInfraVariables.put(CREDENTIALS_KEY,
                                             Credentials.getCredentialsBase64((byte[]) parameters[index++]));
             } catch (KeyException e) {
-                throw new IllegalArgumentException("Could not retrieve base64 credentials", e);
+                logger.debug("Could not retrieve base64 credentials", e);
```
fviale: if credentials are provided and incorrect, maybe we could trigger the error ?

Mykhailenko: you are right

rm/rm-policy/rm-policy-scheduler/src/main/java/org/ow2/proactive/scheduler/resourcemanager/nodesource/policy/SchedulerAwarePolicy.java
```diff
@@ -88,7 +89,11 @@ public BooleanWrapper configure(Object... policyParameters) {
         schedulerUrl = policyParameters[2].toString();
 
         if (policyParameters[3] == null) {
-            throw new IllegalArgumentException("Credentials must be specified");
+            try {
+                policyParameters[3] = nodeSource.getAdministrator().getCredentials().getBase64();
+            } catch (KeyException e) {
+                logger.info("Cannot use default admin credentials.", e);
+            }
```
MaelAudren: If instead throwing an exception when `policyParameters[3] == null` you log the error. I think it will return a NullPointerException when you do `credentials = (byte[]) policyParameters[3]`.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
```diff
@@ -251,6 +251,14 @@ private void startNodeProcess(int numberOfNodes) {
         }
     }
 
+    private String getCredentials() throws KeyException {
```
MaelAudren: Why did you add `throws KeyException`

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructure.java
```diff
@@ -337,14 +338,11 @@ public void configure(Object... parameters) {
...
             // credentials
-            if (parameters[index] == null) {
-                throw new IllegalArgumentException("Credentials must be specified");
-            }
             try {
                 persistedInfraVariables.put(CREDENTIALS_KEY,
                                             Credentials.getCredentialsBase64((byte[]) parameters[index++]));
             } catch (KeyException e) {
-                throw new IllegalArgumentException("Could not retrieve base64 credentials", e);
+                logger.debug("Provided credentials are not compatible with Base64 format.");
```
mcompastie: The comment from @fviale also applies here. I  bet user should be invited to propose their own credential instead.

rm/rm-policy/rm-policy-scheduler/src/main/java/org/ow2/proactive/scheduler/resourcemanager/nodesource/policy/SchedulerAwarePolicy.java
Mykhailenko: thanks. i changed

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructure.java
```diff
@@ -337,12 +338,11 @@ public void configure(Object... parameters) {
...
             // credentials
-            if (parameters[index] == null) {
-                throw new IllegalArgumentException("Credentials must be specified");
-            }
             try {
-                persistedInfraVariables.put(CREDENTIALS_KEY,
-                                            Credentials.getCredentialsBase64((byte[]) parameters[index++]));
+                if (parameters[index] != null) {
+                    persistedInfraVariables.put(CREDENTIALS_KEY,
+                                                Credentials.getCredentialsBase64((byte[]) parameters[index++]));
```
fviale: this is semantically different, as you increase the index only when the parameter is not null !

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
```diff
@@ -279,7 +287,7 @@ protected void configure(Object... args) {
         try {
             this.credentials = Credentials.getCredentialsBase64((byte[]) args[index++]);
         } catch (KeyException e1) {
-            throw new IllegalArgumentException("Cannot decrypt credentials", e1);
+            logger.debug("Cannot decrypt credentials", e1);
```
fviale: why logger debug here ?

PR #3605, opened 2019-08-08
Add new endpoint getNodesMBeanHistory

aminelouati:


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
```diff
@@ -639,19 +687,28 @@ void setMBeanInfo(@HeaderParam("sessionid") String sessionId, @PathParam("name")
...
+     * 	"ConfigNodesCount",
+     * 	"DownNodesCount",
+     * 	"LostNodesCount",
+     * 	"AverageActivity" };
+     * 	</pre>
      *
      *
      * @param sessionId a valid session
-     * @param range a String of 5 chars, one for each stat history source, indicating the time range to fetch
+     * @param range a String of 9 chars, one for each stat history source, indicating the time range to fetch
```
mcompastie: 10 Chars ?

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -611,6 +605,26 @@ public String getNodeMBeanHistory(@HeaderParam("sessionid") String sessionId,
         return rmProxy.getNodeMBeanHistory(nodeJmxUrl, objectName, attrs, range);
     }
 
+    @Override
+    @GET
+    @GZIP
+    @Produces("application/json")
+    @Path("nodes/mbean/history")
+    public String getNodesMBeanHistory(@HeaderParam("sessionid") String sessionId,
```
mcompastie: The typing is not aligned with the declaration in RMRestInterface interface. Please confirm it's OK.

aminelouati: You are right. It has to be Object as in the declaration in RMRestInterface interface.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
aminelouati: It is  9 which is number of the different datasource (AvailableNodesCount, FreeNodesCount.. ).

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -31,13 +31,7 @@
...
 import java.net.URISyntaxException;
 import java.security.KeyException;
-import java.util.AbstractMap;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
mboussaa: @aminelouati could you please import only the map class from java.util instead of importing all classes? thx

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -611,6 +605,26 @@ public String getNodeMBeanHistory(@HeaderParam("sessionid") String sessionId,
...
+            throws InstanceNotFoundException, IntrospectionException, ReflectionException, IOException,
+            NotConnectedException, MalformedObjectNameException, NullPointerException, MBeanException {
+
+        // checking that still connected to the RM
+        RMProxyUserInterface rmProxy = checkAccess(sessionId);
+        Map nodesMBeanHistoryMap = new HashMap<String, String>();
+        for (String nodeJmxUrl : nodesJmxUrl) {
+            nodesMBeanHistoryMap.put(nodeJmxUrl, rmProxy.getNodeMBeanHistory(nodeJmxUrl, objectName, attrs, range));
+        }
+        return nodesMBeanHistoryMap.toString();
```
mboussaa: @aminelouati if we return "nodesMBeanHistoryMap", it will call automatically toString() right? so maybe we don't need to put toString(). (I'm not sure though)

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -611,6 +605,26 @@ public String getNodeMBeanHistory(@HeaderParam("sessionid") String sessionId,
...
+    @GZIP
+    @Produces("application/json")
+    @Path("nodes/mbean/history")
+    public String getNodesMBeanHistory(@HeaderParam("sessionid") String sessionId,
+            @QueryParam("nodejmxurl") Set<String> nodesJmxUrl, @QueryParam("objectname") String objectName,
+            @QueryParam("attrs") List<String> attrs, @QueryParam("range") String range)
+            throws InstanceNotFoundException, IntrospectionException, ReflectionException, IOException,
+            NotConnectedException, MalformedObjectNameException, NullPointerException, MBeanException {
+
+        // checking that still connected to the RM
```
mboussaa: //checking that still connected to the RM to remove it was previously explained in other endpoints

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
mcompastie: Ok

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
```diff
@@ -459,11 +459,18 @@ Object getNodeMBeanInfo(@HeaderParam("sessionid") String sessionId, @QueryParam(
...
+     * @param attrs set of mbean attributes
      * @param range a String of 5 chars, one for each stat history source, indicating the time range to fetch
      *      for each source. Each char can be:<ul>
      *            <li>'a' 1 minute
      *            <li>'m' 10 minutes
+     *            <li>'n' 5 minutes
+     *            <li>'t' 30 minutes
      *            <li>'h' 1 hour
+     *            <li>'j' 2 hours
+     *            <li>'k' 4 hours
```
mboussaa: @aminelouati I didn't understand from where u brought 'n' 't' 'j' 'k' parameters

mboussaa: Don't bother about this comment. I understood why you added them.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
aminelouati: Maybe better to return a JSON object. I am updating the PR. 
Thanks

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
aminelouati: Indeed, the IDE replaced automatically 
import java.util.Collection;	
import java.util.Collections;	
import java.util.Iterator;	
import java.util.List;	
import java.util.Map;	
import java.util.Set;

by import java.util.*;

It is no less efficient to use a wildcard import -- all those extra classes will only be loaded if they are used.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
aminelouati: ok

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
```diff
@@ -488,6 +495,47 @@ Object getNodeMBeanHistory(@HeaderParam("sessionid") String sessionId, @QueryPar
...
+     *            <li>'M' 1 month
+     *            <li>'y' 1 year</ul>
+     * @return a JSON object containing a key for each source
+     * @throws InstanceNotFoundException
+     * @throws IntrospectionException
+     * @throws ReflectionException
+     * @throws IOException
+     * @throws MalformedObjectNameException
+     * @throws NullPointerException
+     * @throws NotConnectedException
```
Mykhailenko: There are rules which I follow when I write Java Docs. 
I never leave @throws, @param, or @return without comments - cause it is only make code longer.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -600,7 +602,7 @@ public Object getNodeMBeanInfo(@HeaderParam("sessionid") String sessionId,
     @GZIP
     @Produces("application/json")
     @Path("node/mbean/history")
-    public String getNodeMBeanHistory(@HeaderParam("sessionid") String sessionId,
+    public Object getNodeMBeanHistory(@HeaderParam("sessionid") String sessionId,
```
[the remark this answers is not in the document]
aminelouati: In the super Method it is Object. Besides, the result is a json object.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -975,4 +998,59 @@ private int getParametersLength(String[] parameters) {
         }
     }
 
+    private Map<String, Object> processHistoryResult(String mbeanHistory, String range) throws IOException {
+        mbeanHistory = removingInternalEscaping(mbeanHistory);
+        Map<String, Object> jsonMap = parseJSON(mbeanHistory);
+        Map<String, Object> processedJsonMap = new HashMap<>();
+        if (jsonMap == null) {
+            return null;
+        }
```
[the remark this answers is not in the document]
aminelouati: you are right. Will do that. 
Thanks

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -975,4 +998,59 @@ private int getParametersLength(String[] parameters) {
...
+        long currentTime = new Date().getTime() / 1000;
+        long duration = StatHistory.Range.valueOf(range).getDuration();
+        int size = Optional.ofNullable(((ArrayList) jsonMap.entrySet().iterator().next().getValue()).size()).orElse(0);
+        long step = duration / size;
+
+        List<String> labels = new ArrayList<>();
```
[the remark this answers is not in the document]
aminelouati: It is rather the first item of the map. 
The map contains the history of one of several metrics. 
Here is an example:
```{
    "TotalMem": [
        17179869184,
        17179869184,
        17179869184,
        17179869184
    ],
    "FreeMem": [
        182292480,
        179892224,
        137711616,
        137711616,
}```

File: rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java [hunk 11]
```diff
@@ -975,4 +998,59 @@ private int getParametersLength(String[] parameters) {
...
+        result = result.replace("\\\"", "\"");
+        result = result.replace("\"{", "{");
+        result = result.replace("}\"", "}");
+        return result;
+    }
+
+    private Map<String, Object> parseJSON(final String jsonString) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            final JsonParser parser = mapper.getJsonFactory().createJsonParser(jsonString);
```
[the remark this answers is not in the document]
aminelouati: Yes, these methods are deprecated since 2.1 version. However, we are using an old version (build.gradle of scheduler-api project).      
compile group: 'org.codehaus.jackson', name: 'jackson-mapper-asl', version: '1.9.12'
Should I upgrade the used version ? or import  the new version in the build.gradle of the rest-server project.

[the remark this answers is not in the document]
MaelAudren: Jackson libraries are doing things well. They already implemented other method that do the same job. I show you an example in my upper comment how you could rewrite the code with methods that are not deprecated.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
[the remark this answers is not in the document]
MaelAudren: Yes I understand this was the response of one of my previous comment. However I maintain that you should set the list capacity.

PR #3608, opened 2019-08-20
Refactor workflow variable types to enum

tao-xinxiu:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/ModelValidator.java
```diff
@@ -86,48 +89,33 @@ public String validate(String parameterValue, ModelValidatorContext context) thr
...
+        uppercaseModel = removePrefix(uppercaseModel);
+        for (ModelType type : ModelType.values()) {
+            if (uppercaseModel.startsWith(type.name())) {
+                try {
+                    return (ParserValidator) type.getTypeParserValidator()
+                                                 .getDeclaredConstructor(String.class)
+                                                 .newInstance(removePrefix(model));
+                } catch (InstantiationException | IllegalAccessException | InvocationTargetException
+                        | NoSuchMethodException e) {
+                    if (e instanceof InvocationTargetException) {
```
MaelAudren: You could catch InvocationTargetException without the other Exceptions it would avoid to check the instantiation of the exception. However by doing so you avoid duplicate the final throw. Was it your purpose there ?

tao-xinxiu: Yes, that's the reason.

PR #3610, opened 2019-08-26
Run PA Nodes on z/OS

nebilbenmabrouk:


rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/RMNodeStarter.java
```diff
@@ -461,7 +461,11 @@ protected String configure(final String args[]) {
         configureSecurityManager();
         configureRMAndProActiveHomes();
         configureProActiveDefaultConfigurationFile();
-        loadSigarIfRunningWithOneJar();
+        try {
+            loadSigarIfRunningWithOneJar();
```
fviale: this will not be sufficient, as we expect to run PA nodes without onejar (node.jar) as well

nebilbenmabrouk: Actually the issue comes from the method 'getNativeLibraryName()', which appears only in loadSigarIfRunningWithOneJar(). So, I deem that this is the only place where to add this change

fviale: You are right, I didn't read the full code. Do you think we should set disabledMonitoring = true if an error such as this occurs ?

nebilbenmabrouk: We already set disabledMonitoring = true (we use -dm), but this error is not catched by the 'disabledMonitoring' property.

fviale: indeed, but when this error occurs, we could automatically disable monitoring, don't you think ?

nebilbenmabrouk: I prefer to let  the user choose (to disable it explicitly), and we don't really need it here as the error is catched and a simple warning is displayed

fviale: all right ...

PR #3613, opened 2019-09-02
feat: add workflow variable type `PA:CREDENTIAL`

tao-xinxiu:
Add a workflow variable type `PA:CREDENTIAL`. 
This variable type uses _Scheduler Third-Party Credentials_ to store credentials which can be later accessed through its key (e.g. `credentials.get(variables.get("VAR_KEY")`).

The variable value is required to be a **key** which exists in the _Scheduler Third-Party Credentials_.

Note, the workflow validate request is not always sending from the clients which has the access of _Scheduler Third-Party Credentials_, for example, the client "catalog". In such case, the variable validity check is skipped.

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
@@ -25,6 +25,10 @@
  */
 package org.ow2.proactive.scheduler.common.job.factories.spi.model.validator;
 
+import java.util.Set;
+
+import org.ow2.proactive.scheduler.common.exception.NotConnectedException;
+import org.ow2.proactive.scheduler.common.exception.PermissionException;
 import org.ow2.proactive.scheduler.common.job.factories.spi.model.ModelValidatorContext;
 import org.ow2.proactive.scheduler.common.job.factories.spi.model.exceptions.ValidationException;
 
@@ -36,7 +40,16 @@ import org.ow2.proactive.scheduler.common.job.factories.spi.model.exceptions.Val
 public class CredentialValidator implements Validator<String> {
     @Override
     public String validate(String parameterValue, ModelValidatorContext context) throws ValidationException {
-        // TODO if parameterValue not exists in 3rd-party credentials, throw ValidationException
+        // if parameterValue not exists in 3rd-party credentials, throw ValidationException
+        try {
+            Set<String> credentialKeys = context.getScheduler().thirdPartyCredentialsKeySet();
+            if (!credentialKeys.contains(parameterValue)) {
+                throw new ValidationException("Expected value should exist in the third-party credentials.");
+            }
+        } catch (NotConnectedException | PermissionException e) {
+            e.printStackTrace(); //TODO
+            throw new ValidationException("Exception during getting the third-party credentials.", e);
+        }
         return parameterValue;
     }
 
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
@@ -27,6 +27,7 @@ package org.ow2.proactive.scheduler.common.job.factories.spi.model.validator;
 
 import java.util.Set;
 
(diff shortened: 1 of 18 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
```diff
@@ -0,0 +1,66 @@
...
+ * @author ActiveEon Team
+ * @since 27/08/19
+ */
+public class CredentialValidator implements Validator<String> {
+    private static final Logger logger = Logger.getLogger(CredentialValidator.class);
+
+    @Override
+    public String validate(String parameterValue, ModelValidatorContext context) throws ValidationException {
+        if (context == null || context.getScheduler() == null) {
+            // Sometimes the workflow is submitted without scheduler instance (e.g., submitted from catalog).
```
fviale: in case of the catalog, a workflow is not submitted, a workflow is just parsed and stored in the catalog. the logger warn is not necessary (you can put it as debug if you want) as it would pollute the catalog log file (basically, we know what we are doing).

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
```diff
@@ -0,0 +1,66 @@
...
+            logger.warn(String.format("Can't check the validity of the variable value, because missing the access to the scheduler third-party credentials",
+                                      parameterValue));
+            return parameterValue;
+        }
+
+        // if parameterValue not exists in 3rd-party credentials, throw ValidationException
+        try {
+            Set<String> credentialKeys = context.getScheduler().thirdPartyCredentialsKeySet();
+            if (!credentialKeys.contains(parameterValue)) {
+                throw new ValidationException("Expected value should exist in the third-party credentials.");
```
fviale: please output the credential key in the error message. This give the user the ability to read the error message from the studio, copy the missing key, and create the missing credential (the original credential key stored in the variable value might contain references to other variables, making it not handy for copy/paste. The server error message will show, on the other hand, the final value, with all references resolved).

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        } catch (NotConnectedException | PermissionException e) {
+            throw new ValidationException("Exception during getting the third-party credentials.", e);
```
fviale: Could not read third party-credentials from the scheduler, make sure you are connected and you have permission right to read third-party credentials

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
fviale: it would be useful to test on the studio, if you can read this message when trying to validate the job (and see the missing credentials key)

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
tao-xinxiu: Done. Changed to debug.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
tao-xinxiu: Done. The error msg changed to `String.format("Could not find the key named [%s] in the third-party credentials. Please add it into the scheduler third-party credentials or change the variable value to a valid key.", parameterValue)`

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CredentialValidator.java
tao-xinxiu: Done. Thanks.

PR #3615, opened 2019-09-03
Nodes History endpoint

Mykhailenko:
Group by node source name, then hostname, then node name.
Sort by start time.
If start time before window then set it to window start.
if end time after window (or zero) then set it to window end.

### DIFF
--- a/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
+++ b/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
@@ -770,6 +770,7 @@ public interface RMRestInterface {
     @GZIP
     @Path("nodes/history")
     @Produces("application/json")
-    List<RMNodeHistory> getNodesHistory(@HeaderParam("sessionid") String sessionId, @HeaderParam("windowStart") long windowStart,
-                           @HeaderParam("windowEnd") long windowEnd) throws NotConnectedException;
+    List<RMNodeHistory> getNodesHistory(@HeaderParam("sessionid") String sessionId,
+            @HeaderParam("windowStart") long windowStart, @HeaderParam("windowEnd") long windowEnd)
+            throws NotConnectedException;
 }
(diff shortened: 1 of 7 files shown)


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRestInterface.java
```diff
@@ -765,4 +766,10 @@ String getStatHistory(@HeaderParam("sessionid") String sessionId, @QueryParam("r
     String getNodeThreadDump(@HeaderParam("sessionid") String sessionId, @QueryParam("nodeurl") String nodeUrl)
             throws NotConnectedException;
 
+    @GET
+    @GZIP
+    @Path("nodes/history")
+    @Produces("application/json")
+    List<RMNodeHistory> getNodesHistory(@HeaderParam("sessionid") String sessionId, @HeaderParam("windowStart") long windowStart,
```
aminelouati: Very good job !
I suggest to return a Map<String, List<RMNodeHistory>> to ease the work from the client side.

Mykhailenko: good. i will do it

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/db/RMDBManager.java
```diff
@@ -814,4 +814,14 @@ public RMDBManagerBuffer getBuffer() {
         return rmdbManagerBuffer;
     }
 
+    public List<NodeHistory> getNodesHistory(long windowStart, long windowEnd) {
+        return executeReadTransaction(session -> {
+            Query query = session.createQuery("FROM NodeHistory " +
```
fviale: endtime is by default null, how will this query behave for :windowStart <= endTime and endTime <= :windowEnd

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/db/RMDBManager.java
```diff
@@ -814,4 +814,21 @@ public RMDBManagerBuffer getBuffer() {
...
     }
 
+    public List<NodeHistory> getNodesHistory(long windowStart, long windowEnd) {
+        return executeReadTransaction(session -> {
+            //            this expression is built as negation to expression that
+            //            retrieves all items that we do not want to have in the response.
+            //            just to clarify the events we include
+            //            we discard only events :
+            //            - ending before the windows start
+            //            - starting after the window start
```
aminelouati: starting after the window End ?

Mykhailenko: yes you are right

PR #3618, opened 2019-09-05
Remove credentials generating in rm-node

Mykhailenko:
It might brake agents. But we need to remove this horrible credentials.


Mykhailenko:
Retest this please

PR #3620, opened 2019-09-05
Add usage info to RMNodeData

Mykhailenko:
To add usage info in the RMNodeData


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -642,6 +653,15 @@ protected NodeSet getRMNodes(Map<JobId, JobDescriptor> jobMap, int neededResourc
         }
     }
 
+    private Map<String, String> getUsageInfo(InternalTask internalTask0) {
+        Map<String, String> metaInfoAboutTask = new HashMap<>();
```
MaelAudren: Set  a capacity to the Map: `new HashMap<>(4)`

PR #3623, opened 2019-09-06
Add server timezone endpoint

Mykhailenko:
GET http://localhost:8080/rest/scheduler/servertimezone
HEADER: sessionid
Returns something like this:
`Europe/Paris +02:00`


Mykhailenko:
Nobody need it

PR #3627, opened 2019-09-16
minor changes in javadoc

fviale:
- following last change in the documentation, it was necessary to add the rm-server to the javadoc generation.
 - minor changes to fix the javadoc of the rm-server


build.gradle
```diff
@@ -493,6 +493,7 @@ def exportedProjects= [
         ":rest:rest-client",
         ":rest:rest-smartproxy",
         ":rm:rm-client",
+        ":rm:rm-server"
```
MaelAudren: What will be the impact to add rm-server here ?

fviale: this is to export rm server javadoc only, it does not create any cross-project dependencies

PR #3628, opened 2019-09-16
rm-api

Mykhailenko:
- create RMNodeClient
- which uses RMRestClient


rest/rest-api/build.gradle
```diff
@@ -3,4 +3,6 @@ dependencies {
     compile 'org.jboss.resteasy:resteasy-multipart-provider:3.0.26.Final'
     compile "org.objectweb.proactive:programming-util:${programmingVersion}"
     compile project(':scheduler:scheduler-api')
+    compile "org.objectweb.proactive:programming-core:${programmingVersion}"
+    compile project(':rm:rm-client')
```
fviale: these two imports are really problematic because they add a huge set of jars to all projects using rest-api (and there are a lot)

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMClientExample.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class RMClientExample {
+
+    public static void main(String[] args) throws LoginException, KeyException, ActiveObjectCreationException, NodeException, RMException {
+        System.out.println("Hello from example.");
+
+//        HttpClientBuilder httpClientBuilder = HttpClientBuilder.create().useSystemProperties();
+        RMRestClient client = new RMRestClient("http://localhost:8080/rest/", null);//, new ApacheHttpClient4Engine(httpClientBuilder.build()));
+        RMRestInterface rm = client.getRm();
+        String sessionId = rm.rmConnect("admin", "admin");
```
fviale: This is ok as a simple proxy (and in the way the SchedulerRestClient is made).

To be elegantly usable in a task like the schedulerapi, a second proxy should be made.

For example, the schedulerapi binding can be used like this inside a task:

schedulerapi.connect() (automatically uses the scheduler url transfered to the task by variable, and the current user credentials)
schedulerapi.submitJob(myJob)  (schedulerapi implements the scheduler interface).

This proxy is created in the class org.ow2.proactive.scheduler.task.client.SchedulerNodeClient

rest/rest-api/build.gradle
Mykhailenko: Then I would need to move all these classes to rest-api:
import org.ow2.proactive.resourcemanager.common.NSState;
import org.ow2.proactive.resourcemanager.common.RMState;
import org.ow2.proactive.resourcemanager.common.event.RMNodeHistory;
import org.ow2.proactive.resourcemanager.common.event.RMNodeSourceEvent;
import org.ow2.proactive.resourcemanager.common.event.dto.RMStateDelta;
import org.ow2.proactive.resourcemanager.common.event.dto.RMStateFull;
import org.ow2.proactive.resourcemanager.exception.RMException;
import org.ow2.proactive.resourcemanager.frontend.topology.Topology;
import org.ow2.proactive.resourcemanager.nodesource.common.NodeSourceConfiguration;
import org.ow2.proactive.resourcemanager.nodesource.common.PluginDescriptor;

rest/rest-api/build.gradle
```diff
@@ -3,4 +3,6 @@ dependencies {
     compile 'org.jboss.resteasy:resteasy-multipart-provider:3.0.26.Final'
     compile "org.objectweb.proactive:programming-util:${programmingVersion}"
     compile project(':scheduler:scheduler-api')
+//    compile "org.objectweb.proactive:programming-core:${programmingVersion}"
```
fviale: could you remove the commented out line?

settings.gradle
```diff
@@ -26,3 +26,7 @@ include 'rest:rest-client'
 include 'rest:rest-smartproxy'
 
 include 'war-wrapper'
+include 'rm-api'
+include 'rm:rm-api'
+include 'rm:rm-api'
```
fviale: there seem to be a duplicate

Mykhailenko: sure

rest/rest-api/build.gradle
Mykhailenko: sure

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
```diff
@@ -0,0 +1,428 @@
...
+        init(new ConnectionInfo(schedulerRestUrl, userCreds.getLogin(), userCreds.getPassword(), null, true));
+    }
+
+    @Override
+    public void init(ConnectionInfo connectionInfo) throws Exception {
+        HttpClient client = new HttpClientBuilder().insecure(connectionInfo.isInsecure()).useSystemProperties().build();
+
+        RMRestClient restApiClient = new RMRestClient(connectionInfo.getUrl(), new ApacheHttpClient4Engine(client));
+
+        //        ResteasyProviderFactory factory = ResteasyProviderFactory.getInstance();
```
fviale: this commented out code should be removed

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    @Override
+    public String getSession() {
+        return sessionId;
+    }
+
+    @Override
+    public String rmConnect(String username, String password) throws KeyException, LoginException, RMException {
+        return rm.rmConnect(username, password);
```
fviale: this should throw UnsupportedOperationException as the connection is handled by init only

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
```diff
@@ -0,0 +1,428 @@
...
+    @Override
+    public boolean removeNode(String sessionId, String nodeUrl, boolean preempt)
+            throws NotConnectedException, PermissionRestException {
+        return false;
+    }
+
+    @Override
+    public boolean removeNodeSource(String sessionId, String sourceName, boolean preempt)
+            throws NotConnectedException, PermissionRestException {
+        return false;
```
fviale: why do you return false and don't forward the code in many methods ?

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMClientExample.java
```diff
@@ -0,0 +1,48 @@
...
+import org.ow2.proactive_grid_cloud_portal.common.RMRestInterface;
+
+
+public class RMClientExample {
+
+    public static void main(String[] args) throws LoginException, KeyException, RMException {
+        System.out.println("Hello from example.");
+
+        //        HttpClientBuilder httpClientBuilder = HttpClientBuilder.create().useSystemProperties();
+        RMRestClient client = new RMRestClient("http://localhost:8080/rest/", null);//, new ApacheHttpClient4Engine(httpClientBuilder.build()));
```
MaelAudren: Explain why you keep the commented code or remove it please.

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMRestClient.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        this.httpEngine = httpEngine;
+        this.restEndpointURL = restEndpointURL;
+
+        providerFactory = ResteasyProviderFactory.getInstance();
+        if (!providerFactory.isRegistered(RMRestClient.JacksonContextResolver.class)) {
+            providerFactory.registerProvider(RMRestClient.JacksonContextResolver.class);
+        }
+        //        registerGzipEncoding(providerFactory);
```
MaelAudren: Explain why you keep this code commented or remove it please.

rest/rest-client/src/test/java/functionaltests/RMClientTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+
+public class RMClientTest extends SchedulerClientTest {
+
+    @Test(timeout = MAX_WAIT_TIME)
+    public void testSchedulerNodeClient() throws Throwable {
+        ISchedulerClient client = clientInstance();
+        Job job = nodeClientJob("/functionaltests/descriptors/rm_client_node.groovy", null, null);
+        JobId jobId = submitJob(job, client);
+        TaskResult tres = client.waitForTask(jobId.toString(), "NodeClientTask", TimeUnit.MINUTES.toMillis(5));
```
MaelAudren: tres -> tRes

rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/frontend/topology/TopologyInfo.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Gets the distances associated to the host.
+     * Note that it does not return information about all hosts handled by the resource manager
+     * but rather the information about hosts added before the one specified.
+     *
+     * @param hostAddress - inet address of the host
+     * @return distances map to nodes added before this one
+     */
+    HashMap<InetAddress, Long> getHostTopology(InetAddress hostAddress);
```
MaelAudren: Why not Map ?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/ForkedTaskVariablesManager.java
```diff
@@ -125,6 +132,17 @@ public SchedulerNodeClient createSchedulerNodeClient(TaskContext container) {
         return null;
     }
 
+    public RMNodeClient createRMNodeClient(TaskContext container) {
```
MaelAudren: It would be good to document the case where this method return null value.

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMClientExample.java
Mykhailenko: This class is called *Example.
It is for us

rest/rest-client/src/test/java/functionaltests/RMClientTest.java
Mykhailenko: ok

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMRestClient.java
Mykhailenko: thanks

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
```diff
@@ -0,0 +1,651 @@
...
+        }
+    }
+
+    @Override
+    public ConnectionInfo getConnectionInfo() {
+        return connectionInfo;
+    }
+
+    @Override
+    public void setSession(String sid) {
```
fviale: I think the setSession cannot be used like in the schedulerNodeClient, better remove this method

rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/frontend/topology/TopologyInfo.java
Mykhailenko: I dont know, it was taken from Topology class. We splitted this class into Topology and TopologyInfo

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
```diff
@@ -0,0 +1,651 @@
...
+     */
+    public void disconnect() throws NotConnectedException {
+        rm.rmDisconnect(sessionId);
+    }
+
+    /**
+     * @return Returns the state of the Resource Manager
+     */
+    public RMState getState() throws NotConnectedException {
+        return rm.getState(sessionId);
```
fviale: what happens when the connect() did not succeed or was not called, what kind of exception is thrown?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
Mykhailenko: ok

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/ForkedTaskVariablesManager.java
Mykhailenko: ok

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
Mykhailenko: no exception. Methods return nulls.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
fviale: because setSession does nothing. In the SchedulerNodeClient, one can do
schedulerapi.setSession("mycustomid")
schedulerapi.renewSession()
And then use the api.

Here it is not possible to do so

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
fviale: Then we need to find a way to handle when the user is not connected to the rm and throw exceptions if not

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
Mykhailenko: here we can setSession. and then call anymethod, and new sessionid would be used

fviale: Ok I didn't see it, but please note that this.rm is set in the init method only, so you need to call init first.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
Mykhailenko: ok. i will call checkSession() method in the beginning of each method

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMClientExample.java
MaelAudren: Every comments are for us. I don't understand your argument.

rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/frontend/topology/TopologyInfo.java
MaelAudren: ok

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMClientExample.java
Mykhailenko: There are 2 ways of creating the objects. One way commented out, and another is used.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
Mykhailenko: actually no need to call init. we can just change session id and become another user of the same url.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
```diff
@@ -111,6 +107,12 @@ public void renewSession() {
         }
     }
 
+    private void checkSessionNoNull() throws NotConnectedException {
+        if (sessionId == null) {
+            throw new NotConnectedException("You are not connected to RM. Call 'connect()' first.");
```
fviale: You are not connected to the Resource Manager. Call 'connect()' first before using the interface.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
```diff
@@ -111,6 +107,12 @@ public void renewSession() {
         }
     }
 
+    private void checkSessionNoNull() throws NotConnectedException {
```
fviale: checkNonEmptySession

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMClientExample.java
MaelAudren: Can you add a comment explaining that please. I think it should be crystal clear for other developers ?

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/client/RMRestClient.java
MaelAudren: Identified issue is still there.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
Mykhailenko: Ok

scheduler/scheduler-node/src/main/java/org/ow2/proactive/resourcemanager/task/client/RMNodeClient.java
Mykhailenko: checkNorris

PR #3631, opened 2019-09-20
Fix rmapi

Mykhailenko:


rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/frontend/topology/TopologyImpl.java
```diff
@@ -124,6 +124,10 @@ public boolean onSameHost(Node node, Node node2) {
         return distances.keySet();
     }
 
+    public HashMap<String, InetAddress> getGetHosts() {
```
fviale: isn't there a mistake ? getHosts() no ?

PR #3636, opened 2019-09-25
Add new rm rest endpoints to return the list of hosts/node sources as…

fviale:
… variable model

Also:
 - skip model fetching in PA:MODEL_FROM_URL if the url contains variables
 - minor constants refactoring

### DIFF
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/RMRestInterface.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/RMRestInterface.java
@@ -178,6 +178,24 @@ public interface RMRestInterface {
             throws NotConnectedException, PermissionRestException;
 
     /**
+     * Returns a task variable model string containing the list of registered hosts in the resource manager
+     * @return a model containing the list of hosts, including an empty name. e.g. PA:LIST(,hostname1,hostname2)
+     */
+    @GET
+    @Path("model/hosts")
+    @Produces("text/plain")
+    String getModelHosts() throws PermissionRestException;
+
+    /**
+     * Returns a task variable model string containing the list of registered node sources in the resource manager
+     * @return a model containing the list of node sources name, including an empty name and the default node source e.g. PA:LIST(,Default,LocalNodes)
+     */
+    @GET
+    @Path("model/nodesources")
+    @Produces("text/plain")
+    String getModelNodeSources() throws PermissionRestException;
+
+    /**
      * Returns true if the resource manager is operational.
      *
      * @param sessionId
--- a/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
+++ b/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
@@ -70,7 +70,9 @@ import org.ow2.proactive.authentication.UserData;
 import org.ow2.proactive.authentication.crypto.CredData;
 import org.ow2.proactive.authentication.crypto.Credentials;
 import org.ow2.proactive.resourcemanager.common.NSState;
+import org.ow2.proactive.resourcemanager.common.RMConstants;
 import org.ow2.proactive.resourcemanager.common.RMState;
+import org.ow2.proactive.resourcemanager.common.event.RMNodeEvent;
 import org.ow2.proactive.resourcemanager.common.event.RMNodeHistory;
 import org.ow2.proactive.resourcemanager.common.event.RMNodeSourceEvent;
 import org.ow2.proactive.resourcemanager.common.event.dto.RMStateDelta;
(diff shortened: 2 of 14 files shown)


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/RMRestInterface.java
```diff
@@ -177,6 +177,24 @@ RMStateDelta getRMStateDelta(@HeaderParam("sessionid") String sessionId,
...
             throws NotConnectedException, PermissionRestException;
 
+    /**
+     * Returns a task variable model string containing the list of registered hosts in the resource manager
+     * @return a model containing the list of hosts, including an empty name
+     */
+    @GET
+    @Path("model/hosts")
+    @Produces("text/plain")
+    String getModelHosts() throws PermissionRestException;
```
MaelAudren: Could you add an exmple in the javadoc of the format. As it is a text/plain format it is really not clear what will be the response.

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/RMRestInterface.java
```diff
@@ -177,6 +177,24 @@ RMStateDelta getRMStateDelta(@HeaderParam("sessionid") String sessionId,
...
+
+    /**
+     * Returns a task variable model string containing the list of registered node sources in the resource manager
+     * @return a model containing the list of node sources name, including an empty name and the default node source
+     */
+    @GET
+    @Path("model/nodesources")
+    @Produces("text/plain")
+    String getModelNodeSources() throws PermissionRestException;
```
MaelAudren: Same as upper comment.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -229,6 +230,30 @@ public RMStateFull getRMStateFull(String sessionId) throws NotConnectedException
         return orThrowRpe(RMStateCaching.getRMStateFull());
     }
 
+    @Override
+    public String getModelHosts() throws PermissionRestException {
+        RMStateFull state = orThrowRpe(RMStateCaching.getRMStateFull());
```
Mykhailenko: Are you sure this method does not return REMOVED node events?

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -229,6 +230,30 @@ public RMStateFull getRMStateFull(String sessionId) throws NotConnectedException
...
+        return "PA:LIST(," +
+               state.getNodesEvents()
+                    .stream()
+                    .map(event -> event.getHostName())
```
Mykhailenko:
```suggestion
                    .map(RMNodeEvent::getHostName)
```

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
```diff
@@ -229,6 +230,30 @@ public RMStateFull getRMStateFull(String sessionId) throws NotConnectedException
...
+                    .distinct()
+                    .collect(Collectors.joining(",")) +
+               ")";
```
MaelAudren: Is it normal that `PA:LIST(,` start with a comma ?
I think it would be more readable using String.format
```
return String.format("PA:LIST(%s)",
               state.getNodesEvents()
                    .stream()
                    .map(event -> event.getHostName())
                    .distinct()
                    .collect(Collectors.joining(","));

File: rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java [hunk 6]
```diff
@@ -229,6 +230,30 @@ public RMStateFull getRMStateFull(String sessionId) throws NotConnectedException
...
+    public String getModelNodeSources() throws PermissionRestException {
+        RMStateFull state = orThrowRpe(RMStateCaching.getRMStateFull());
+        return "PA:LIST(," + RMConstants.DEFAULT_STATIC_SOURCE_NAME +
+               "," + state.getNodeSource()
+                          .stream()
+                          .map(event -> event.getSourceName())
+                          .distinct()
+                          .collect(Collectors.joining(",")) +
+               ")";
+    }
```
MaelAudren: Same comment as upper:
```
        return String.format("PA:LIST(%s,%s",RMConstants.DEFAULT_STATIC_SOURCE_NAME,
                          state.getNodeSource()
                          .stream()
                          .map(event -> event.getSourceName())
                          .distinct()
                          .collect(Collectors.joining(","));
```

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/RMRestInterface.java
fviale: ok

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/RMRestInterface.java
fviale: ok

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
fviale: ok for format.

Yes for PA:LIST it is normal (unfortunately). I need to produce a list which allows an empty choice. And there is no other way to do it than directly inside the endpoint.

Empty choice is needed because of a bootstrap issue. When someone wants to define a workflow or task variable with PA:MODEL_FROM_URL pointing to this endpoint. The variable must have a valid default value (otherwise the studio does not allow to create it), which thus must be a valid hostname or node source. This makes the workflow non-portable and it could not be stored in the catalog proactive-examples.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
fviale: same answer

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/rm/RMRest.java
fviale: doubled-checked and REMOVED node or node source events are already filtered from the full state

PR #3641, opened 2019-10-01
Selection Scripts late binding

fviale:
- SelectionScript: modified constructors, hash, equals, digest methods. Added a new static function to return a resolve copy of a selection script
 - ProbabilisticSelectionManager: fetch script content if not present and minor try/catch changes
 - SchedulingMethodImpl: use resolved copies of the original selection scripts
 - added new rm functional test

### DIFF
--- a/common/common-api/src/main/java/org/ow2/proactive/scripting/SelectionScript.java
+++ b/common/common-api/src/main/java/org/ow2/proactive/scripting/SelectionScript.java
@@ -33,7 +33,6 @@ import java.io.StringReader;
 import java.net.URL;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.List;
 
 import javax.script.Bindings;
 
@@ -75,7 +74,7 @@ public class SelectionScript extends Script<Boolean> {
     /**
      * Hash digest of the script
      */
-    protected byte[] id_;
+    protected byte[] hash = null;
 
     /** ProActive needed constructor */
     public SelectionScript() {
@@ -93,7 +92,6 @@ public class SelectionScript extends Script<Boolean> {
      */
     public SelectionScript(String script, String engineName) throws InvalidScriptException {
         super(script, engineName, "SelectionScript");
-        buildSelectionScriptId();
     }
 
     /** Directly create a script with a string.
@@ -105,7 +103,6 @@ public class SelectionScript extends Script<Boolean> {
     public SelectionScript(String script, String engineName, boolean dynamic) throws InvalidScriptException {
         super(script, engineName, "SelectionScript");
         this.dynamic = dynamic;
-        buildSelectionScriptId();
     }
 
     /** Directly create a script with a string.
@@ -119,7 +116,6 @@ public class SelectionScript extends Script<Boolean> {
             throws InvalidScriptException {
         super(script, engineName, parameters, "SelectionScript");
         this.dynamic = dynamic;
-        buildSelectionScriptId();
     }
 
     /** Create a selection script from a file.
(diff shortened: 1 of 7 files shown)


common/common-api/src/main/java/org/ow2/proactive/scripting/SelectionScript.java
```diff
@@ -374,29 +373,21 @@ public boolean compareByteArray(byte[] array1, byte[] array2) {
...
-     * String representation
-     */
+    @Override
     public String toString() {
-        return hashCode() + "\n" + script;
+        if (script != null) {
+            return "" + new String(getHash()) + "\n" + script;
+        } else {
+            return "" + new String(getHash()) + "\n" + url;
+        }
```
nebilbenmabrouk: Can be replaced by: 
return script != null ? "" + new String(getHash()) + "\n" + script : "" + new String(getHash()) + "\n" + url;

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/statistics/ProbablisticSelectionManager.java
```diff
@@ -108,56 +108,55 @@ public ProbablisticSelectionManager(RMCore rmcore) {
...
+                SelectionScript scriptWithReplacedBindings = replaceBindings(script, bindings);
+                String digest = "";
+                try {
+                    digest = new String(scriptWithReplacedBindings.digest());
+                } catch (NoSuchAlgorithmException e) {
+                    logger.error(e.getMessage(), e);
+                }
+                if (probabilities.containsKey(digest) && probabilities.get(digest).containsKey(rmnode.getNodeURL())) {
+                    double probability = probabilities.get(digest).get(rmnode.getNodeURL()).value();
+                    if (probability == 0) {
```
Mykhailenko:
```suggestion
                 if (Double.compare(probability, 0) < 0.000001) {
```

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/statistics/ProbablisticSelectionManager.java
```diff
@@ -108,56 +108,55 @@ public ProbablisticSelectionManager(RMCore rmcore) {
...
+        Set<RMNode> nodeSet = intersectionMap.keySet();
+        List<RMNode> res = new ArrayList<>(nodeSet.size());
+        res.addAll(nodeSet);
+        Collections.sort(res, new NodeProbabilityComparator(intersectionMap));
+
+        if (logger.isDebugEnabled()) {
+            logger.debug("The following nodes are selected for scripts execution (time is " +
+                         (System.currentTimeMillis() - startTime) + " ms) :");
+            if (res.size() > 0) {
+                for (RMNode rmnode : res) {
```
Mykhailenko:
```suggestion
                res.forEach(rmnode -> {
```

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/rmnode/RMNode.java
```diff
@@ -81,6 +81,7 @@
      * and the responses given.
      * @return the map of Script and status
      */
+    @Deprecated
```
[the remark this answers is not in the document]
fviale: Actually, it's more because it is not used at all.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/statistics/ProbablisticSelectionManager.java
fviale: but Double.compare returns an int ...

PR #3644, opened 2019-10-08
Allow docker fork environment on windows

fviale:
- add new fork environment property "isDockerWindowsToLinux" to control path transformations
 - transform paths if needed in ForkedJvmTaskExecutionCommandCreator
 - also add the possibility to define the preJavaCommand as list (more robust)
 - added some unit tests


mboussaa:
Code not obvious to understand but LGTM

PR #3645, opened 2019-10-09
Improve handling of deserialization timeouts

fviale:
- increased default timeout to 2 minutes (to tolerate general cpu overload of the machine)
 - disable timeout for serialization
 - add a data size threshold for deserialization (timeout will be disabled for large data size)

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/AllObjects2BytesConverterHandler.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/AllObjects2BytesConverterHandler.java
@@ -46,7 +46,11 @@ public class AllObjects2BytesConverterHandler {
 
     private final static String TIMEOUT_JAVA_PROPERTY = "pa.max.deserialization.seconds";
 
-    private final static Long secondsToWait = Long.getLong(TIMEOUT_JAVA_PROPERTY, 30L);
+    private final static String DESERIALIZATION_THRESHOLD = "pa.max.deserialization.threshold";
+
+    private final static Long secondsToWait = Long.getLong(TIMEOUT_JAVA_PROPERTY, 120L);
+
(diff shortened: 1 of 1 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/AllObjects2BytesConverterHandler.java
```diff
@@ -58,30 +62,46 @@ private AllObjects2BytesConverterHandler() {
...
         try {
-            resultMap = future.get(secondsToWait, TimeUnit.SECONDS);
+            if (totalSize < deserializationThreshold) {
+                // If the amount of bytes to deserialize is small, the deserialization should not exceed a timeout,
+                // otherwise it means a deadlock occurs (most likely due to jython types)
+                // timeout is disabled for serialization
+                resultMap = future.get(secondsToWait, TimeUnit.SECONDS);
+            } else {
+                resultMap = future.get();
+            }
```
nebilbenmabrouk: I don't understand why the deserialization is still executed when the totalSize is greater than deserializationThreshold

fviale: we don't want to prevent deserialization for big size data. In that case, we don't use a timeout as we know the deserialization can take some time. The timeout is used for small size data, where it's not normal to exceed a default timeout.

nebilbenmabrouk: what if the pb of deadlock occurs for a big size data? how to detect that?

fviale: indeed, we cannot detect that

PR #3648, opened 2019-10-17
Improve SchedulerFrontend concurrency

fviale:
- SchedulerFrontendState: use read/write locks instead of synchronized on all methods
 - SchedulerFrontend: use dynamic instead of fixed thread pools. Use ImmediateService whenever possible
 - settings : increase maximum number of thread parameters (dynamicity allows to increase these values)
 - make SchedulerProxyUserInterface use also @ImmediateService to allow concurrent operations in the same user session.
 - test helpers: add current time in log output

### DIFF
--- a/common/common-api/src/main/java/org/ow2/proactive/utils/Lambda.java
+++ b/common/common-api/src/main/java/org/ow2/proactive/utils/Lambda.java
@@ -44,6 +44,11 @@ public class Lambda {
     private Lambda() {
     }
 
+    /**
+     * Execute a runnable with lock acquire/release
+     * @param lock the lock to acquire
+     * @param runnable code to execute
+     */
     public static void withLock(Lock lock, Runnable runnable) {
         lock.lock();
         try {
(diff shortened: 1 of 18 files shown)


fviale:
Retest this please

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -708,87 +738,94 @@ synchronized void checkJobOwner(String methodName, IdentifiedJob IdentifiedJob,
...
+                if (t.getReadableName().equals(taskName)) {
+                    taskId = t;
+                }
             }
-            return ts;
-        }
+            if (taskId == null) {
+                throw new UnknownTaskException(taskName, jobId);
+            }
+            synchronized (jobState) {
```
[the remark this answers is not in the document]
fviale: yes, because we synchronize on a different object than the lock.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -598,27 +615,32 @@ synchronized void disconnect() throws NotConnectedException, PermissionException
...
+        Lambda.withLockInterruptible(usersWriteLock, () -> {
+            credentials.remove(id);
+            ListeningUser ident = identifications.remove(id);
+            if (ident != null) {
+                // remove listeners if needed
+                ident.clearListener();
+                // remove this user to the list of connected user if it has not
+                // already been removed
+                ident.getUser().setToRemove();
+                Lambda.withLock(stateWriteLock, () -> sState.getUsers().update(ident.getUser()));
```
tao-xinxiu: We could add a comment in the declaration of userLock and stateLock about their order, to reduce risk of deadlock in the future dev.

common/common-api/src/main/java/org/ow2/proactive/utils/Lambda.java
```diff
@@ -64,6 +64,59 @@ public static void withLock(Lock lock, Runnable runnable) {
...
+            }
+            throw new RuntimeException(e);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    public static <T, E1 extends Exception> T withLockException1(Lock lock, Callable<T> callable, Class<E1> type1)
```
tao-xinxiu: Function could be renamed as `withLockNotWrappingException(lock, callable, notWrappingExceptionType)`

common/common-api/src/main/java/org/ow2/proactive/utils/Lambda.java
```diff
@@ -64,6 +64,59 @@ public static void withLock(Lock lock, Runnable runnable) {
...
+                throw (E1) e;
+            }
+            throw new RuntimeException(e);
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    public static <T, E1 extends Exception, E2 extends Exception> T withLockException2(Lock lock, Callable<T> callable,
+            Class<E1> type1, Class<E2> type2) throws E1, E2 {
```
tao-xinxiu: Function could be renamed as `withLockNotWrappingException(lock, callable, notWrappingExceptionType1, notWrappingExceptionType2)`

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
fviale: Very good remark, and indeed I made a mistake somewhere regarding lock hierarchy

fviale:
Retest this please

PR #3650, opened 2019-11-04
ParentId as a field

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -523,9 +525,23 @@ InternalJob createJob(Job userJob, UserIdentificationImpl ident)
         // setting the job properties
         job.setOwner(ident.getUsername());
 
+        fillParentJobIdIfExistsInGenInfo(userJob, job);
+
         return job;
     }
 
+    private void fillParentJobIdIfExistsInGenInfo(Job userJob, InternalJob job) {
+        if (userJob.getGenericInformation().containsKey(PARENT_JOB_ID)) {
```
fviale: I think generic information map can be null

PR #3653, opened 2019-11-14
Add maximum login attempts mechanism

fviale:
Configurable mechanism with shared properties. By default 3 unsuccesful attempts on a 10 minutes window disables
further authentication attempts until the window is expired.

### DIFF
--- a/common/common-client/src/main/java/org/ow2/proactive/authentication/FileLoginModule.java
+++ b/common/common-client/src/main/java/org/ow2/proactive/authentication/FileLoginModule.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.security.KeyException;
 import java.security.PrivateKey;
+import java.util.Date;
 import java.util.Map;
 import java.util.Properties;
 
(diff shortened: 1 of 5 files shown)


common/common-client/src/main/java/org/ow2/proactive/authentication/FileLoginModule.java
```diff
@@ -179,19 +186,34 @@ public boolean login() throws LoginException {
...
+        if (isNotFallbackAuthentication) {
+            removeOldFailedAttempts(username);
+
+            if (tooManyFailedAttempts(username)) {
+                String message = "Too many failed login/attempts, please try again in " +
+                                 retryInHowManyMinutes(username) + " minutes.";
+                logger.warn("[" + FileLoginModule.class.getSimpleName() + "] " + message);
+                throw new FailedLoginException(message);
+            }
+        }
```
nebilbenmabrouk: What happens if the fallback mechanism is used? the user can try to login without any limitation?

fviale: in the fallback mechanism the file login module is checked first then the main module is checked next (ldap or pam).

I disabled the maximum failed attempts on the file login module in that case (but it is kept when both login attempts fail), otherwise accounts would be blocked all the time for non file-based users.

nebilbenmabrouk: Fair enough :)

PR #3654, opened 2019-11-15
Add helper binding for task progress

fviale:
used with progress.set() in workflow tasks

Also fixed some paths in docker on windows fork environment

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
@@ -1,0 +1,82 @@
+/*
+ * ProActive Parallel Suite(TM):
+ * The Open Source library for parallel and distributed
+ * Workflows & Scheduling, Orchestration, Cloud Automation
+ * and Big Data Analysis on Enterprise Grids & Clouds.
+ *
+ * Copyright (c) 2007 - 2017 ActiveEon
+ * Contact: contact@activeeon.com
+ *
+ * This library is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
+ * as published by the Free Software Foundation: version 3 of
+ * the License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * If needed, contact us to obtain a release under GPL Version 2 or 3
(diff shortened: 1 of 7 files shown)


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
```diff
@@ -0,0 +1,84 @@
...
+ * @author ActiveEon Team
+ * @since 24/10/2019
+ */
+public class TaskProgressImpl implements TaskProgress {
+
+    File progressFile;
+
+    transient PrintStream output;
+
+    transient PrintStream error;
```
MaelAudren: VisibleForTesting ?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public TaskProgressImpl(String progressFilePath, PrintStream output, PrintStream error) {
+        if (progressFilePath != null) {
+            progressFile = new File(progressFilePath);
+        } else {
+
+        }
```
MaelAudren: else empty ?

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
fviale: we don't use VisibleForTesting in the scheduling project

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
fviale: right it's a mistake

nebilbenmabrouk: else {
        throw new IllegalArgumentException("Invalid path for the progress file: " + progressFilePath);
}

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
MaelAudren: ok

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/TaskProgressImpl.java
fviale: actually, I've removed the else case, this path could eventually be null in tests (in which case we do not care).

fviale:
@nebilbenmabrouk this class is not called internally. It can be called by a workflow task using:

progress.set(23)

Basically, it's a binding inside workflow tasks

PR #3666, opened 2019-12-01
Ossindex integration

mcompastie:
- OSS-Index plugin integration
- Applying several maintenance update for enforcing identified security fixes


common/common-api/build.gradle
```diff
@@ -7,7 +7,7 @@ dependencies {
     compile group: 'org.jasypt', name: 'jasypt', version: '1.9.3'
 
     testRuntime 'org.jruby:jruby-complete:9.0.5.0'
-    testRuntime 'org.python:jython-standalone:2.7.0'
+    testRuntime 'org.python:jython-standalone:2.7.1'
```
fviale: We cannot update jython, this has caused many issues when we tried to do this in the past

rm/rm-node/build.gradle
```diff
@@ -21,7 +21,7 @@ dependencies {
 
     // script engines
     runtime 'org.jruby:jruby-complete:9.0.5.0'
-    runtime 'org.python:jython-standalone:2.7.0'
+    runtime 'org.python:jython-standalone:2.7.1'
```
fviale: same

rm/rm-server/build.gradle
```diff
@@ -22,7 +22,7 @@ dependencies {
     testCompile project(':common:common-api').sourceSets.test.output
 
     testCompile 'org.jruby:jruby-complete:9.0.5.0'
-    testCompile 'org.python:jython-standalone:2.7.0'
+    testCompile 'org.python:jython-standalone:2.7.1'
```
fviale: same

PR #3670, opened 2019-12-03
Fix zipping directories bug

mklkun:


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/utils/Zipper.java
```diff
@@ -159,9 +159,11 @@ private static void zipFiles(List<File> files, String basepath, OutputStream os)
                 ZipOutputStream zos = new ZipOutputStream(os);
                 closer.register(zos);
                 for (File file : files) {
-                    FileInputStream inputStream = new FileInputStream(file);
-                    closer.register(inputStream);
-                    writeZipEntry(zipEntry(basepath, file), inputStream, zos);
+                    if (file.isFile()) {
+                        FileInputStream inputStream = new FileInputStream(file);
+                        closer.register(inputStream);
+                        writeZipEntry(zipEntry(basepath, file), inputStream, zos);
```
fviale: it seems strange not to write a zip entry when the pathname refers to a directory. Maybe it's how it's supposed to be, or automatically done when children files are added, but what about empty directories ? I guess they would be skipped

mklkun: Yes, it is automatically done when children are added.  You're right too about empty folders, they are skipped since that they don't contain input files. Is this behaviour correct or not for you ?

fviale: I think it would be better to add them, if it is not complicated

mklkun: Ok. I will work on it.

PR #3672, opened 2019-12-04
Add PendingCurrentPast job and task stats per username in MBeans

amouhoub:
In this PR I add a few metrics to the MBeans responsible for keeping job and task statistics for the `stats/myaccount` endpoint (stats per user)


scheduler/scheduler-server/src/test/java/functionaltests/db/schedulerdb/TestReadSchedulerAccount.java
```diff
@@ -50,10 +50,21 @@
 
         int taskCount;
 
+        // TODO complete the singleJobScenario with task the following statistics
```
MaelAudren: I think it is better to create a task than to keep a TODO in the code

amouhoub: Sure, the todo was just for during the review to avoid getting asked to remove unused variables. I'll complete the test before merging

PR #3675, opened 2019-12-06
fix running tasks count for stats/myaccount endpoint

amouhoub:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -661,7 +661,7 @@ public SchedulerAccount readAccount(final String username) {
             }
 
             int pendingTasksCount = getTaskCountForUser(TaskStatus.PENDING_TASKS, username);
-            int currentTasksCount = getTaskCountForUser(TaskStatus.RUNNING_TASKS, username);
+            int currentTasksCount = getTaskCountForUser(Collections.singleton(TaskStatus.RUNNING), username);
```
MaelAudren: Why only this one take a list as a parameter and not the upper and the under lines ?

amouhoub: because the others are already lists

amouhoub:
@MaelAudren I swear I did spotlessApply this time 😆

PR #3678, opened 2019-12-10
Fix bug root path when zipping inputs

mklkun:


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/client/utils/Zipper.java
```diff
@@ -232,14 +232,18 @@ public static void unzipEntry(ZipInputStream zis, OutputStream os) throws IOExce
             }
         }
 
-        private static ZipEntry zipEntry(String basepath, File file) throws IOException {
-            String realAbsolutePath = Paths.get(file.getAbsolutePath()).toRealPath().toString();
+        private static String getRealRelativeFilePath(String basepath, String absolutePath) throws IOException {
+            String realAbsolutePath = Paths.get(absolutePath).toRealPath().toString();
             String realBasePath = Paths.get(basepath).toRealPath().toString();
+            return basepath.endsWith("/") ? realAbsolutePath.substring(realBasePath.length())
```
mcompastie: basepath.endWith(Files.separator)

mcompastie: You should use the dedicated Java datastruct. representing path instead of string to bring a fully crossplatform implementation.

PR #3679, opened 2019-12-11
Load addon classes with child-first delegation mechanism

tao-xinxiu:
**Goal**: Use child-first class loading mechanism to load the addon classes. It allows different addons to use their proper version of dependent library. The PR resolves the potential library version conflict between the different addons and server components.

**Scope**: Currently the addon class are mainly infrastructure plugins. Specifically, a class is taken as an addon if found its jar located under class simple name in lowercase in server `addons` directory. Currently the child-first delegation mechanism is only used for addon classes.

**Implemented changes**: To ensure the correct behavior of the classes which loaded by child-first mechanism. Its thread context class loader is switched to be identical as its class loader during its class loading (and initializing), instance creation, and methods execution. And it is switched back after the addon code finishes its execution. Specifically, proxy mechanism is used for wrapping the addons class methods execution to switch the thread context class loader. Note, the static methods are not wrapped. 

**Note**, from now on, it's expected to always use `AddonClassUtils` for loading and instantiating the addon classes. And calling the addons static method outside the addons jar (which is rarely the case) should be wrapped through `ContextClassLoaderInjector.switchContextClassLoader`.

### DIFF
--- a/rm/rm-server/build.gradle
+++ b/rm/rm-server/build.gradle
@@ -16,7 +16,6 @@ dependencies {
     )
 
     compile 'com.google.guava:guava:19.0'
-    compile group: 'org.javassist', name: 'javassist', version: '3.15.0-GA'
 
     testCompile "org.objectweb.proactive:programming-extension-pnp:${programmingVersion}"
     testCompile "org.objectweb.proactive:programming-extension-pnpssl:${programmingVersion}"
--- a/rm/rm-server/build.gradle
+++ b/rm/rm-server/build.gradle
@@ -16,6 +16,7 @@ dependencies {
     )
 
     compile 'com.google.guava:guava:19.0'
+    compile group: 'org.javassist', name: 'javassist', version: '3.15.0-GA'
 
     testCompile "org.objectweb.proactive:programming-extension-pnp:${programmingVersion}"
     testCompile "org.objectweb.proactive:programming-extension-pnpssl:${programmingVersion}"
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/utils/AddonPathUtils.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/utils/AddonPathUtils.java
@@ -26,9 +26,11 @@
 package org.ow2.proactive.resourcemanager.utils;
 
 import java.io.File;
-import java.io.FilenameFilter;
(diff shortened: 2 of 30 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/utils/AddonPathUtils.java
```diff
@@ -0,0 +1,89 @@
...
+     * @return
+     */
+    public static URL[] findJarsInPath(String path) {
+        File[] jarArray = new File(path).listFiles(new FilenameFilter() {
+            @Override
+            public boolean accept(File dir, String name) {
+                return name.endsWith(".jar");
+            }
+        });
+        HashSet<URL> jarNames = new HashSet<>();
```
Mykhailenko: nit-picking...
```suggestion
        Set<URL> jarNames = new HashSet<>();
```

tao-xinxiu:
retest this please

rm/rm-server/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
     )
 
     compile 'com.google.guava:guava:19.0'
+    compile group: 'org.javassist', name: 'javassist', version: '3.15.0-GA'
```
fviale: programming dependencies should already include javassist, normally we should reuse this version and not overwrite it

tao-xinxiu: Done. It's removed.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/utils/AddonPathUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        File infraAddonDir = new File(getPathInAddons(dirName));
+        return infraAddonDir.exists() && infraAddonDir.isDirectory();
+    }
+
+    /**
+     * Find all the jar files in the path
+     * @param path
+     * @return
+     */
+    public static URL[] findJarsInPath(String path) {
```
fviale: I think this method could be simplified using something like:
try (DirectoryStream<Path> jars =
    Files.newDirectoryStream(pluginsDir, "*.jar")) {

    for (Path jar : jars) {
        urlSet.add(jar.toUri().toURL());
    }
}

PR #3683, opened 2019-12-19
Fix db backup

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerBackupRunner.java
```diff
@@ -108,16 +111,18 @@ private void performBackup() throws IOException {
...
-              .forEach(file -> {
-                  LOGGER.info("Removing old backup: " + file.getName());
-                  file.delete();
-              });
-
+        File[] files = fodlerWhereBackups.listFiles();
+        if (files != null) {
+            Stream.of(files)
+                  .filter(File::isFile)
+                  .filter(file -> file.getName().startsWith(PREFIX))
```
aminelouati: As explained here https://dzone.com/articles/single-filter-perform-better-than-multiple-one-in, maybe it is better to use a single filter instead of many.

Mykhailenko: Thanks Amine, I did not know it. However in this case I think performance improvement is really neglactable compairng to code clarity. I will leave as it is. Sorry

PR #3684, opened 2019-12-20
Resolve dead lock in scheduler backup runner

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/synchronization/AOSynchronization.java
```diff
@@ -128,6 +128,15 @@ public AOSynchronization() {
         // ProActive empty no arg constructor
     }
 
+    public void freeze() throws IOException {
+        recordManager.close();
```
fviale: some work needs to be done in runActivity method in order to make freeze coherent inside the api. 
Right now, after calling freeze, all other methods will fail. I would prefer that they are not executed until resume is called.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerBackupRunner.java
```diff
@@ -54,34 +55,50 @@
...
         } catch (IOException | InterruptedException e) {
             LOGGER.error("", e);
         } finally {
             scheduler.resume();
         }
     }
 
-    private boolean noOtherTaskIsRunning() {
-        return scheduler.getJobs().getRunningTasks().isEmpty();
+    private boolean otherTaskIsRunning() {
```
fviale: noTaskIsRunning

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerBackupRunner.java
```diff
@@ -54,34 +55,50 @@
...
+            if (!otherTaskIsRunning()) {
+                synchronizationAPI.freeze();
+                try {
+                    removeOldBackups();
+                    performBackup();
+                } finally {
+                    synchronizationAPI.resume();
+                }
+            } else {
+                LOGGER.info("Backup will not be performed because there is still some jobs running");
```
fviale: LOGGER.warn

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerBackupRunner.java
```diff
@@ -54,34 +55,50 @@
...
 
     @Override
     public void run() {
+        scheduler.freeze();
         try {
-            scheduler.freeze();
-
-            while (!noOtherTaskIsRunning()) {
-                Thread.sleep(10000);
+            DateTime deadLine = DateTime.now().plusSeconds(possibleDealyInSeconds);
```
fviale: possibleDelay

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/synchronization/AOSynchronization.java
```diff
@@ -51,6 +51,7 @@
 import org.objectweb.proactive.InitActive;
 import org.objectweb.proactive.RunActive;
 import org.objectweb.proactive.Service;
+import org.objectweb.proactive.annotation.ImmediateService;
```
fviale: you can remove this import

PR #3689, opened 2019-12-27
Add email notification for JOB_RUNNING_TO_FINISHED_WITH_ERRORS

mboussaa:
Add email notification for jobs having the notification event `JOB RUNNING TO FINISHED WITH ERRORS`:

-Consider a job `running to finished with errors` if it has one or more `faulty` tasks
-Add event to SchedulerEvent (`JOB_RUNNING_TO_FINISHED_WITH_ERRORS` or `Job running to finished with errors` could be introduced as GI)
-For now, the filtering concerns only past finished jobs, which means jobs having finished as status.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
@@ -237,7 +237,7 @@ public class JobEmailNotification {
     }
 
     private String getSubject(boolean withErrors) {
-        String event = withErrors ? SchedulerEvent.JOB_RUNNING_TO_FINISHED_WITH_ERRORS.toString().toLowerCase()
+        String event = withErrors ? SchedulerEvent.JOB_RUNNING_TO_FINISHED_WITH_ERRORS.toString()
                                   : eventType.toString();
         String jobID = jobState.getId().value();
         return String.format(SUBJECT_TEMPLATE, jobID, event);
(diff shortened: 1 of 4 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -125,50 +126,83 @@ public boolean doCheckAndSend(boolean withAttachment)
...
                                                     "Error sending email: " + e.getMessage(),
                                                     e);
         }
     }
 
+    private boolean hasFaultyTasks() {
+        List<TaskState> tasks = jobState.getTasks();
+        boolean withErrors = tasks.stream()
+                                  .map(task -> task.getStatus().name().toLowerCase())
+                                  .anyMatch(t -> t.equals("Faulty".toLowerCase()));
```
fviale: why not using JobInfo contained in notification data?

Besides this, we are interested in getNumberOfFailedTasks, getNumberOfFaultyTasks and getNumberOfInErrorTasks

mboussaa: @fviale thanks for notifying this. The new commit should have fixed that.

fviale:
as a matter of fact, maybe it could be useful to add a unit test for the new feature you introduced

scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/core/JobEmailNotificationTest.java
```diff
@@ -302,6 +302,26 @@ public void testFinished() throws Exception {
         verifyNoMoreInteractions(stubbedSender);
     }
 
+    @Test
+    public void testFinishedWithErrors() throws Exception {
```
fviale: you should add a test to verify that, if a job contains no error, with NOTIFICATION_EVENTS sets the same as this test, no notification should occur.

PR #3691, opened 2020-01-02
Yet another fix

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerBackupRunner.java
```diff
@@ -109,32 +116,27 @@ private boolean someTaskIsRunning() {
...
-            LOGGER.error("Cannot save a backup because backup destination folder does not exist: " + destination);
+            LOGGER.error("Cannot save a backup because backup configuredDestination folder does not exist: " +
+                         configuredDestination);
         }
     }
 
     private void removeOldBackups() {
-        File fodlerWhereBackups = new File(destination);
-        File[] files = fodlerWhereBackups.listFiles();
+        File[] files = resolvedDestination.listFiles();
```
fviale: I think this can throw an IOException on first run.

1) removeOldBackups is called before performBackup
2) performBackup calls mkdirs
so if the backup directory does not exist, on first run, an IOException will be thrown when reaching resolvedDestination.listFiles();

fviale: ok, listFiles returns null if destination does not exist

PR #3693, opened 2020-01-06
Handle notification event ALL

mboussaa:
-Fix the issue when the NOTIFICATION_EVENT is ALL, finished jobs without errors are not notified

-Fix as well when JOB_RUNNING_TO_FINISHED_WITH_ERRORS and JOB_RUNNING_TO_FINISHED are added together as NOTIFICATION_EVENT

-Add comments

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
@@ -107,6 +107,12 @@ public class JobEmailNotification {
 
     public boolean doCheckAndSend(boolean withAttachment)
             throws JobEmailNotificationException, IOException, UnknownJobException, PermissionException {
+        String eventFinishedWithErrorsName = SchedulerEvent.JOB_RUNNING_TO_FINISHED_WITH_ERRORS.name().toLowerCase();
+        String eventFinishedWithErrorsMethod = SchedulerEvent.JOB_RUNNING_TO_FINISHED_WITH_ERRORS.toString()
+                                                                                                 .toLowerCase();
+        String eventFinishedName = SchedulerEvent.JOB_RUNNING_TO_FINISHED.name().toLowerCase();
+        String eventFinishedMethod = SchedulerEvent.JOB_RUNNING_TO_FINISHED.toString().toLowerCase();
+
         String jobStatus = jobState.getGenericInformation().get(GENERIC_INFORMATION_KEY_NOTIFICATION_EVENT);
         List<String> jobStatusList = new ArrayList<>();
         if (jobStatus != null) {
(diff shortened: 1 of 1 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -136,22 +142,39 @@ public boolean doCheckAndSend(boolean withAttachment)
...
                         sendEmail(withAttachment, true);
+                        //if not we do not send any notification as JOB_RUNNING_TO_FINISHED is not included
                     } else {
                         return false;
                     }
+
+                    // second case: check if JOB_RUNNING_TO_FINISHED is provided along with JOB_RUNNING_TO_FINISHED_WITH_ERRORS (e.g., the case of 'All' event)
+                } else if ((jobStatusList.contains(eventFinishedWithErrorsName) ||
+                            jobStatusList.contains(eventFinishedWithErrorsMethod)) &&
+                           (jobStatusList.contains(eventFinishedName) || jobStatusList.contains(eventFinishedMethod))) {
```
mklkun: This test 
`&& (jobStatusList.contains(eventFinishedName) || jobStatusList.contains(eventFinishedMethod))`
can be deleted since we are in the `else if` bloc.

mboussaa: @mklkun that's true. I fix it. Thx))

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -107,6 +107,12 @@ public JobEmailNotification(JobState js, NotificationData<JobInfo> notification,
...
     public boolean doCheckAndSend(boolean withAttachment)
             throws JobEmailNotificationException, IOException, UnknownJobException, PermissionException {
+        String eventFinishedWithErrorsName = SchedulerEvent.JOB_RUNNING_TO_FINISHED_WITH_ERRORS.name().toLowerCase();
+        String eventFinishedWithErrorsMethod = SchedulerEvent.JOB_RUNNING_TO_FINISHED_WITH_ERRORS.toString()
+                                                                                                 .toLowerCase();
+        String eventFinishedName = SchedulerEvent.JOB_RUNNING_TO_FINISHED.name().toLowerCase();
+        String eventFinishedMethod = SchedulerEvent.JOB_RUNNING_TO_FINISHED.toString().toLowerCase();
+
         String jobStatus = jobState.getGenericInformation().get(GENERIC_INFORMATION_KEY_NOTIFICATION_EVENT);
         List<String> jobStatusList = new ArrayList<>();
```
aminelouati: I think that you can specify the initial size of the ArrayList which is 10. It is more efficient as explained here https://stackoverflow.com/questions/15430247/why-start-an-arraylist-with-an-initial-capacity

mboussaa: True but that `jobStatusList ` has a static content defined as a stream of static notification events so its size won't grow up. Though, It was like that before this PR :p

PR #3701, opened 2020-01-16
Dirty fix to preserve checkbox values

Mykhailenko:


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/NodeSourceParameterHelper.java
```diff
@@ -112,8 +112,16 @@ private void updateDynamicParameterIfNotEqual(List<Serializable> mergedParameter
         String cleanNewValue = newValue.trim();
         String cleanOldValue = oldValue.trim();
 
-        if (configurableField.getMeta().dynamic() && !cleanNewValue.equals(cleanOldValue)) {
-            mergedParameters.set(valueIndex, cleanNewValue);
+        // if new value is empty string and configurable field is checkbox
+        // then we keep old value.
+        // we do it, because I cannot figure out why GWT DynamicForm
+        // sends empty value for non-changed checkbox
```
fviale: does it mean it is not possible to "uncheck" a checkbox and have it acknowledged now ?

fviale: or maybe unchecking will generate a new value containing "false" ?

Mykhailenko: When DynamicForm is not changed. Checkbox value comes as "". Nevertheless, it was true or false before. Not sure, what do you mean by "uncheck".

PR #3702, opened 2020-01-17
Improve email notification

mboussaa:


fviale:
@mboussaa 
Not bad but is it configurable generally in a config file?

config/scheduler/email.template
```diff
@@ -0,0 +1,16 @@
+Dear ProActive User,
+
+This notification is to inform you that the ProActive Job ID: %1$s is %2$s.
+
+Below, you find more details about the job tasks and their statuses:
+%3$s
```
fviale: can you not use named patterns that you could replace?
For example %JOB_ID%, etc?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
```diff
@@ -307,7 +303,11 @@ private String getBody() {
         } catch (UnknownHostException e) {
             logger.debug("Could not get hostname", e);
         }
-        return String.format(BODY_TEMPLATE, jobID, status, allTaskStatusesString, hostname);
+
+        final Properties properties = EmailConfiguration.getConfiguration().getProperties();
+        String templatePath = properties.getProperty(EmailConfiguration.TEMPLATE_PATH);
+        String bodyTemplate = new String(Files.readAllBytes(Paths.get(PASchedulerProperties.getAbsolutePath(templatePath))));
```
fviale: Files.toString(file, Charset.defaultCharset()) seems better to me

config/scheduler/email.template
mboussaa: Done. Check it out here https://github.com/ow2-proactive/scheduling/pull/3702/commits/047b89c3e172dd503f55e95dfd88d6abaa0f1f39

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/JobEmailNotification.java
mboussaa: Done.

mboussaa:
Retest this please

PR #3705, opened 2020-01-20
Improve Scheduler stats & add All Users

Mykhailenko:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/JobData.java
```diff
@@ -701,7 +701,10 @@ JobUsage toJobUsage() {
                                          getProjectName(),
                                          jobId.value(),
                                          getJobName(),
-                                         getFinishedTime() - getStartTime());
+                                         getFinishedTime() - getStartTime(),
+                                         getStatus().toString(),
```
mboussaa: no need for toString here

Mykhailenko: We do need `toString()` because constructor requires String while `getStatus()` returns JobStatus.

PR #3710, opened 2020-01-22
Add additional informations to node source

mbenguig:


rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/common/event/RMNodeSourceEvent.java
```diff
@@ -150,6 +154,10 @@ public String getSourceDescription() {
         return this.nodeSourceDescription;
     }
 
+    public String getAdditionalInformations() {
+        return additionalInformations;
```
fviale: why not having additionalInformations as Map<String, String> and handle this on the client side (e.g. portal)?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
```diff
@@ -94,6 +91,35 @@
 @ActiveObject
 public class NodeSource implements InitActive, RunActive {
 
+    public enum AdditionalInformation {
+
+        ACTUAL_COST,
+        COST_LIMIT
+    }
+
+    private Map<AdditionalInformation, String> additionalInformations = null;
```
fviale: additional information should be generic and a Map<String, String>

rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/common/event/RMNodeSourceEvent.java
mbenguig: Ok

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
mbenguig: Ok

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
```diff
@@ -558,6 +571,57 @@ public void setStatus(NodeSourceStatus status) {
...
+
+    public void removeAndPersistAdditionalInformation(String... keys) {
+        for (String key : keys) {
+            this.additionalInformation.remove(key);
+        }
+        persistAdditionalInformation();
+    }
+
+    private void setRmDbManager(RMDBManager dbManager) {
+        this.dbManager = dbManager;
```
fviale: this setter seems useless to me

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
```diff
@@ -558,6 +571,57 @@ public void setStatus(NodeSourceStatus status) {
...
+            this.descriptor.getAdditionalInformation().put(key, value);
+            this.monitoring.nodeSourceEvent(new RMNodeSourceEvent(RMEventType.NODESOURCE_UPDATED,
+                                                                  this.administrator.getName(),
+                                                                  this.name,
+                                                                  this.getDescription(),
+                                                                  this.additionalInformation,
+                                                                  this.administrator.getName(),
+                                                                  this.getStatus().toString()));
+            // Persist additional information
+            persistAdditionalInformation();
```
fviale: persist should be called before sending the event

fviale:
two comments, otherwise LGTM

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
mbenguig: Ok

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/NodeSource.java
mbenguig: Ok

PR #3717, opened 2020-02-21
Add nodeSource system property to forked jvm task

mklkun:


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/ForkedJvmTaskExecutionCommandCreator.java
```diff
@@ -198,6 +207,14 @@ private void forwardProActiveProperties(List<String> jvmArguments, PAProperty...
         }
     }
 
+    private void forwardOtherProperties(List<String> jvmArguments, PAProperty... propertiesToForward) {
```
tao-xinxiu: Could you add a comment for the function purpose ? the function name is not quite straightforward.

mklkun: Yes, sure.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/executors/forked/env/ForkedJvmTaskExecutionCommandCreator.java
```diff
@@ -177,6 +180,12 @@
         return javaCommand;
     }
 
+    private PAPropertyString createNodeSourceProperty() {
+        PAPropertyString nodeSourceNameProperty = new PAPropertyString("proactive.node.nodesource", false, "local");
```
fviale: the default value for a node source not specified is "Default" (when the node is deployed using the proactive-node command or through an agent without speicifying the node-source parameter)

mklkun: Didn't know that. I will fix it. Thanks.

PR #3719, opened 2020-02-24
Add a script variable PA_NODE_SOURCE

mklkun:


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/TaskLauncher.java
```diff
@@ -524,4 +524,13 @@ private String getNodeName() {
         }
     }
 
+    private String getNodeSourceName() {
+        try {
+            return System.getProperty("proactive.node.nodesource");
+        } catch (Exception e) {
+            logger.debug("Failed to acquire task launcher node source information", e);
```
mbenguig: or logger.error ...  (up to you)

mklkun: I would have used `logger.error` too, but I preferred to follow what have been done in the other methods so far.

PR #3722, opened 2020-02-25
Added schema for notification-service

AlexFalessi:


scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/util/SchedulerHsqldbStarterTest.java
```diff
@@ -59,11 +59,9 @@ public void setUp() throws IOException {
 
     @Test
     public void testConfigureCatalogs() throws Exception {
-        verify(hsqldbServerMock, times(2)).addCatalog(Mockito.<Path> any(), Mockito.<Path> any());
+        verify(hsqldbServerMock, times(3)).addCatalog(Mockito.<Path> any(), Mockito.<Path> any());
 
         schedulerHsqldbStarter.configureCatalogs(hsqldbServerMock, "/scheduler/home/");
-
-        verify(hsqldbServerMock, times(4)).addCatalog(Mockito.<Path> any(), Mockito.<Path> any());
```
AlexFalessi: Removed this line as it was useless, test passed eaven if commented out. 
addCatalog() methods should be called 3 times.

config/notification-service/database.properties
```diff
@@ -0,0 +1,35 @@
+# Database connection settings.
+# To change the database provider, change the driver (hibernate.connection.driver_class),
+# the URL (hibernate.connection.url), and don't forget the dialect (hibernate.dialect)
+
+hibernate.connection.driver_class=org.hsqldb.jdbc.JDBCDriver
+hibernate.connection.url=jdbc:hsqldb:hsql://localhost:9001/notification-service
```
mboussaa: I don't like much having `-` in a DB name. 
Why not call it `notification`
@AlexFalessi @fviale

config/notification-service/database.properties
```diff
@@ -0,0 +1,35 @@
...
+hibernate.dialect=org.hibernate.dialect.HSQLDialect
+
+# Username and password
+hibernate.connection.username=notification-service
```
mboussaa: same

config/notification-service/database.properties
AlexFalessi: I dont mind

PR #3732, opened 2020-03-23
Fix FlowScript creation bug when reboot in case the script is by reference

mklkun:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/task/flow/FlowScript.java
```diff
@@ -453,6 +499,7 @@ public String getId() {
 
     @Override
     protected Reader getReader() {
+        logger.info("this.script: " + this.script);
```
fviale: please remove this log inside a getter

mklkun: Done.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/ScriptData.java
```diff
@@ -103,29 +103,68 @@ static ScriptData createForFlowScript(FlowScript script, TaskData taskData) {
...
-            return FlowScript.createReplicateFlowScript(getScript(), getScriptEngine());
+            return FlowScript.createReplicateFlowScript(getScript(), getScriptEngine(), parameters());
         } else {
             throw new DatabaseManagerException("Invalid flow script action: " + flowScriptActionType);
         }
     }
 
+    FlowScript createFlowScriptByURL() throws InvalidScriptException {
+        try {
+            URL inputUrl = new URL(url);
```
aminelouati: maybe initialize the URL outside the try block

mklkun: Actually the try block is there just because of the exception that can be thrown by the URL constructor. :p

fviale: intialize an URL can throw a checked exception so it makes sense to include it inside the try block (to allow throwing an InvalidScriptEception)

PR #3733, opened 2020-03-23
 Install.sh: Configure job planner housekeeping

fviale:


tools/install_base.sh
```diff
@@ -482,6 +482,8 @@ if [[ "$OLD_PADIR" == "" ]]; then
          sed "s/pa\.scheduler\.job\.removeFromDataBase=.*/pa.scheduler.job.removeFromDataBase=true/g"  -i "$PA_ROOT/default/config/scheduler/settings.ini"
          sed "s/^#pa\.rm\.history\.maxperiod=.*/pa.rm.history.maxperiod=$JOB_CLEANUP_SECONDS/g"  -i "$PA_ROOT/default/config/rm/settings.ini"
```
MaelAudren: Is it normal that rm housekeeping time is commented ?

fviale: well it is commented out in the config file, which means no default maxperiod

PR #3735, opened 2020-03-24
Remove checked exception in RMMonitoring

fviale:


rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMListenerProxy.java
```diff
@@ -145,8 +145,8 @@ private void checkCounter(RMEvent event) {
             logger.warn("Local event counter is " + counter + " vs. rm event counter " + event.getCounter());
             try {
                 this.target.getMonitoring().removeRMEventListener();
-            } catch (RMException e) {
-                logger.error(e.getMessage(), e);
+            } catch (Exception e) {
```
mcompastie: Can we be a bit more selective on the exception to be intercepted ?

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMListenerProxy.java
```diff
@@ -127,7 +127,7 @@ public boolean init(String url, Credentials credentials) throws RMException, Key
     public BooleanWrapper disconnect() {
         try {
             target.getMonitoring().removeRMEventListener();
-        } catch (RMException ignored) {
+        } catch (Exception ignored) {
```
mcompastie: Can we be a bit more selective on the exception to be intercepted ?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/policy/RMAwareNodeSourcePolicy.java
```diff
@@ -76,8 +76,8 @@ public void shutdown(Client initiator) {
         if (!rmShuttingDown) {
             try {
                 rmMonitoring.removeRMEventListener();
-            } catch (RMException e) {
-                logger.error("Error when removing event listener", e);
+            } catch (Exception e) {
```
mcompastie: Can we be a bit more selective on the exception to be intercepted ?

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMListenerProxy.java
fviale: not in the context of removal, close, clean, we want to be as large as possible

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/policy/RMAwareNodeSourcePolicy.java
fviale: same answer

PR #3737, opened 2020-03-25
Improve RM proxy sessions handling

fviale:
To fix an observed thread leak due to RMListenerProxy hanging on getMonitoring().removeRMEventListener() this change aims at rationalizing and adding timeouts to the linked requests.

Additionally, add a Thread.setDefaultUncaughtExceptionHandler to properly log uncaught exceptions in Scheduler.log

### DIFF
--- a/rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/frontend/RMMonitoring.java
+++ b/rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/frontend/RMMonitoring.java
@@ -28,7 +28,6 @@ package org.ow2.proactive.resourcemanager.frontend;
 import org.objectweb.proactive.annotation.PublicAPI;
 import org.ow2.proactive.resourcemanager.common.event.RMEventType;
 import org.ow2.proactive.resourcemanager.common.event.RMInitialState;
-import org.ow2.proactive.resourcemanager.exception.RMException;
 
 
 /**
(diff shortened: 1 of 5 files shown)


rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMListenerProxy.java
```diff
@@ -126,16 +131,19 @@ public boolean init(String url, Credentials credentials) throws RMException, Key
      */
     public BooleanWrapper disconnect() {
         try {
-            target.getMonitoring().removeRMEventListener();
+            getMonitoringOrFail().removeRMEventListener();
         } catch (Exception ignored) {
```
mcompastie: I guess to understand why this exception is so inclusive. However, for debugging purpose, it would great to notify somewhere the failure, and why. A debug in logs looks adapted to me

rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/frontend/RMMonitoring.java
```diff
@@ -42,7 +43,7 @@
  * @since ProActive Scheduling 0.9
  */
 @PublicAPI
-public interface RMMonitoring {
+public interface RMMonitoring extends Serializable {
```
mcompastie: Is it a firm strict constraint from `PAFuture.getFutureValue` method ?

fviale: In gneral an object returned by an active object method must be Serializable, but this interface exception is not really important here, I will probably remove it.

rm/rm-client/src/main/java/org/ow2/proactive/resourcemanager/common/util/RMListenerProxy.java
fviale: ok

PR #3743, opened 2020-04-09
publish rest-api javadoc

fviale:


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/CommonRestInterface.java
```diff
@@ -121,7 +121,7 @@ void logout(@HeaderParam("sessionid")
     /**
      * Get a UserData object associated to a session.
      *
-     * <br/> NOTE: <br/>
+     * <br> NOTE: <br>
```
mcompastie, line 124: Shouldn't it be <br /> to skip line ?

fviale: it doesn't work like this as this api is also used to generate the rest documentation, we need to use <br>

PR #3744, opened 2020-04-16
fix: not convert to URI in var PA:GLOBAL_FILE/PA:USER_FILE

tao-xinxiu:
No longer convert PA:GLOBAL_FILE/PA:USER_FILE variable value to URI, since the path name encoding is already handled in VFS. 
Converting it to URI, then check file existence with uri.toString cause the pb when the file path contains spaces.


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/converter/EscapeSpaceURIConverter.java
```diff
@@ -0,0 +1,42 @@
...
+import org.ow2.proactive.scheduler.common.job.factories.spi.model.exceptions.ConversionException;
+
+
+public class EscapeSpaceURIConverter extends URIConverter {
+
+    public final String encodedSpace = "%20";
+
+    @Override
+    public URI convert(String parameterValue) throws ConversionException {
+        String escapedValue = parameterValue.replace(" ", encodedSpace);
```
mcompastie, line 39: This looks valid to circumvent issue from space character. But what happen when other URL-unfriendly are met ? Shouldn't you use `URLEncoder.encode()` to address them all at once ?

tao-xinxiu: Personally I think it's more safe to disallow those characters instead of encoding them. To handle all these characters also requires work in frontend path encoding and backend vfs resolveFile decoding. Currently I don't feel it's really needed. 

@fviale What do you think ?

MaelAudren: By implementing ourself an encoder we take the risk to miss unsupported characters.

fviale: @tao-xinxiu I think we need to find the most robust solution

fviale: For example this seems to work for relative paths:

path = new File( "ab cd/toto titi/file.txt").toURI().toURL().toString()
root = new File("").toURI().toURL().toString()
println path.replace(root, "")

fviale: are we ever going to have absolute paths here ?

tao-xinxiu: I think it should always be relative path here, as the user may (or should) have no knowledge about the server dataspace root path.

tao-xinxiu: After verification, escaping special characters is already handled in vfs, so stopping convert variable value to URI fix the pb.

PR #3745, opened 2020-04-16
 Support LDAP StartTLS protocol

fviale:
- include library to download remote server certificate as a keystore
 - some small modifications regarding property paths resolution


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/webapp/PortalConfiguration.java
```diff
@@ -155,7 +156,12 @@ public static String getAbsolutePath(String userPath) {
         if (new File(userPath).isAbsolute()) {
             return userPath;
         } else {
-            return PortalConfiguration.REST_HOME.getValueAsString() + File.separator + userPath;
+            File pathName = new File(PortalConfiguration.REST_HOME.getValueAsString(), userPath);
```
ShatalovYaroslav, line 159: This code is appearing in several places for the same logic. Don't you want to extract it?

rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/core/properties/PAResourceManagerProperties.java
```diff
@@ -440,7 +441,12 @@ public static String getAbsolutePath(String userPath) {
         if (new File(userPath).isAbsolute()) {
             return userPath;
         } else {
-            return PAResourceManagerProperties.RM_HOME.getValueAsString() + File.separator + userPath;
+            File pathName = new File(PAResourceManagerProperties.RM_HOME.getValueAsString(), userPath);
+            try {
+                return pathName.getCanonicalPath();
```
ShatalovYaroslav, line 446: related to previuos comment

PR #3747, opened 2020-04-17
Remove usage of ProActiveCounter in ssh infrastructures

fviale:
ProActiveCounter creates different node names on successive deployments.
This generates multiple log and rrd files which can be very big, progressively eating all disk space.

Unicity of the node name is already guaranteed by node source, host name and worker node index, so the proactive counter does not bring much added value.

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/HostsFileBasedInfrastructureManager.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/HostsFileBasedInfrastructureManager.java
@@ -558,7 +558,7 @@ public abstract class HostsFileBasedInfrastructureManager extends Infrastructure
             String configuredAddress = hostTracker.getConfiguredAddress();
             String compliantAddress = convertToCompliantAddress(configuredAddress);
             return nodeSource.getName() + COMPLIANT_ADDRESS_DELIMITER + compliantAddress + COMPLIANT_ADDRESS_DELIMITER +
-                   ProActiveCounter.getUniqID();
+                   0;
         }
 
         protected String extractHostFromNode(String nodeNameOrUrl) {


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/HostsFileBasedInfrastructureManager.java
```diff
@@ -558,7 +558,7 @@ protected String generateNodeName(HostTracker hostTracker) {
             String configuredAddress = hostTracker.getConfiguredAddress();
             String compliantAddress = convertToCompliantAddress(configuredAddress);
             return nodeSource.getName() + COMPLIANT_ADDRESS_DELIMITER + compliantAddress + COMPLIANT_ADDRESS_DELIMITER +
-                   ProActiveCounter.getUniqID();
+                   0;
```
mcompastie, line 561: So, there will be one node name per host controlled by SSH ? I am worried about the case when compliantAdress will be shared among multiple host: In recently, I had the case of nodes using sporadically the IP address associated to docker0 NIC instead of eth0. The IP address associated to docker0 was the same on all VMs In this case, I had multiple node claiming the same IP while being located on different VMs. In similar condition, I am afraid your modification could lead to some collision in node name, and I am not sure how the platform could react. Can we consider using another identifier than compliantAdress ?

fviale:
@mcompastie to reply to your deleted remark. in this case it is about a SSH infrastructure, where deployment to multiple hosts is controlled by a host file, containing, for example
host1 4
host2 8

We can use hostnames or ip addresses, but we are sure they will all be different. Docker is not involved inside ssh infrastructures.

For a given host, the root node name will be the same for avrious nodes deployed on the same host. A node index will be added dynamically .e.g

nodesource_mymachine_0
nodesource_mymachine_1

if two nodes are deployed on mymachine.

mcompastie:
> @mcompastie to reply to your deleted remark. in this case it is about a SSH infrastructure, where deployment to multiple hosts is controlled by a host file, containing, for example
> host1 4
> host2 8
> 
> We can use hostnames or ip addresses, but we are sure they will all be different. Docker is not involved inside ssh infrastructures.
> 
> For a given host, the root node name will be the same for avrious nodes deployed on the same host. A node index will be added dynamically .e.g
> 
> nodesource_mymachine_0
> nodesource_mymachine_1
> 
> if two nodes are deployed on mymachine.

Reading back the code invoking getNodeName(), I released that getNodeName was based on the IP provided at configuration time and was not inferred from the operated node, and the stated issue had no point to be there. Hence my comment's removal.

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/HostsFileBasedInfrastructureManager.java
mboussaa, line 561: why `0` ?  and why not removing `COMPLIANT_ADDRESS_DELIMITER + 0`

fviale: @mboussaa you need to look at HostsFileBased class to understand. COMPLIANT_ADDRESS_DELIMITER is used to extract the host name using a regexp. If I remove it , it will not be able to extract the host name. If I don't put 0, the node name will end with "__" which is ugly.

PR #3749, opened 2020-04-21
LDAP : allow using user id instead of distinguished name in group filter

fviale:
This change allows finding groups for recent OpenLDAP configurations


common/common-client/src/main/java/org/ow2/proactive/authentication/LDAPProperties.java
```diff
@@ -73,6 +73,9 @@
      **/
     public static final String LDAP_GROUP_FILTER = "pa.ldap.group.filter";
 
+    /** use member uid in group search */
+    public static final String LDAP_GROUPSEARCH_USE_UID = "pa.ldap.group.search.use.uid";
```
MaelAudren, line 77: Added but not used ?

fviale: why is it not used?

PR #3753, opened 2020-05-06
Load objects to Catalog: get all existing buckets for admin user

ShatalovYaroslav:
for startup: from Catalog get all buckets for admin user


MaelAudren:
Can you squash you commits before merging please ?

PR #3756, opened 2020-05-13
Disable jetty log file by default

fviale:


fviale:
Retest this please

fviale:
@mcompastie indeed, but according to what I read, the issue has been fixed in a much more recent version of jetty and the fixed as not been backported as a maintenance patch. Upgrading jetty version is difficult and cannot be done in a few days.

PR #3758, opened 2020-05-15
Global variables and generic information

fviale:
- define properties files which allow to define global variables and generic info
 - Adapt StaxJobFactory to handle these global values
 - add relevant tests in TestStaxJobFactory

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/StaxJobFactory.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/StaxJobFactory.java
@@ -45,6 +45,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.stream.Collectors;
 
 import javax.xml.stream.XMLInputFactory;
@@ -94,6 +95,7 @@ import org.ow2.proactive.topology.descriptor.ThresholdProximityDescriptor;
 import org.ow2.proactive.topology.descriptor.TopologyDescriptor;
 import org.ow2.proactive.utils.Tools;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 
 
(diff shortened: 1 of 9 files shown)


scheduler/scheduler-api/src/test/java/org/ow2/proactive/scheduler/common/job/factories/TestStaxJobFactory.java
```diff
@@ -136,6 +152,90 @@ public void testCreateJobShouldNotFailWhenParsingEmptyMetadata() throws Exceptio
         factory.createJob(jobDescriptorWithEmptyMetadata);
     }
 
+    @Test
+    public void testCreateJobWithNoVariablesShouldReferenceGlobalVariablesAndGenericInfoe() throws Exception {
```
mbenguig, line 156: Infoe -> Info olé ! (but no need to fix it)

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/StaxJobFactory.java
```diff
@@ -537,6 +560,55 @@ public JobType getType() {
...
+            } catch (IOException e) {
+                logger.error("Global Variable Configuration file: " + path + " cannot be read.", e);
+                return globalVariables;
+            }
+            Map<String, String> propertyMap = (Map) props;
+            for (Map.Entry<String, String> var : propertyMap.entrySet()) {
+                globalVariables.put(var.getKey(), new JobVariable(var.getKey(), var.getValue()));
+            }
+        }
+        return new LinkedHashMap<>(globalVariables);
```
mbenguig, line 580: Do we need to set and return globalVariables?

fviale: it is important to return a clone, otherwise, changes to the map are persisted (static variable) on subsequent workflow parsing.

getConfiguredGlobalJobVariablesAsReplacementMap() uses directly the static variable, and return as well a new map

fviale: I agree that in the second method, I could check that the variable is initialized instead of calling getConfiguredGlobalJobVariables() systematically

PR #3763, opened 2020-05-27
Stop the scheduler only in case of jython deadlocks

fviale:
It has been observed recently that shutdown of the scheduler server occurs in case of server lag, not due to python deadlocks. The goal of this change is to trigger a shutdown only when the cause is due to jython and not something else.

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/AllObjects2BytesConverterHandler.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/AllObjects2BytesConverterHandler.java
@@ -38,6 +38,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import org.apache.log4j.Logger;
+import org.objectweb.proactive.utils.StackTraceUtil;
 
 
 public class AllObjects2BytesConverterHandler {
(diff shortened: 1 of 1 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/util/AllObjects2BytesConverterHandler.java
```diff
@@ -94,17 +95,24 @@ public static Serializable convertByte2Object(final byte[] value) {
...
+                if (totalSize < deserializationThreshold) {
+                    // If the amount of bytes to deserialize is small, the deserialization should not exceed a timeout,
+                    // otherwise it means a deadlock occurs (most likely due to jython types)
+                    // timeout is disabled for serialization
+                    resultMap = future.get(secondsToWait, TimeUnit.SECONDS);
+                } else {
+                    resultMap = future.get();
+                }
+            } catch (TimeoutException e) {
+                if (StackTraceUtil.getAllStackTraces().contains("org.python.core")) {
```
mboussaa, line 108: dangerous to put only "contain org.python.core" maybe put more explicit message to check. Then, I don't see python-related messages in https://trello.com/c/uCAczDYP/5374-tryqa-deserialization-problem

fviale: I think this is less dangerous than what we have now. we want to stop the scheduler only when the specific python deadlock occur.
But you are right the error message must be changed.

PR #3771, opened 2020-06-15
MemoryLeak: do not keep finished jobs state in memory

fviale:
- SchedulerStateImpl : use a maximum size for finished jobs. As test is using the getState method a lot, it is necessary to keep some retro-compatibility. The default number of jobs kept in SchedulerStateImpl is 1000, but can be configured through a JVM property.
 - SchedulerFrontendState : delete job from jobsMap when the job is finished.

### DIFF
--- a/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/core/SchedulerStateImpl.java
+++ b/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/core/SchedulerStateImpl.java
@@ -64,6 +64,14 @@ public final class SchedulerStateImpl<T extends JobState> implements SchedulerSt
     /** Finished jobs */
     private Map<JobId, T> finishedJobs = Collections.synchronizedMap(new LinkedHashMap<JobId, T>());
 
+    /**
+     * Used to control the maximum size of finished jobs. This is mostly used for test compatibility,
+     * As tests use a lot the getState() request. Normal usage of the scheduler does not rely on getState()
+     */
+    private static Integer maxFinishedJobsSize;
+
+    public final static int DEFAULT_MAX_FINISHED_JOBS_SIZE = 1000;
+
     /** Scheduler status */
     private SchedulerStatus status = SchedulerStatus.STARTED;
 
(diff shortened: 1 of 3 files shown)


scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/core/SchedulerStateImpl.java
```diff
@@ -364,12 +373,22 @@ public synchronized void pendingToRunning(T js) {
 
     public synchronized void pendingToFinished(T js) {
         pendingJobs.remove(js);
+        putInFinished(js);
+    }
+
+    private void putInFinished(T js) {
+        if (finishedJobs.size() >= maxFinishedJobsSize) {
```
mcompastie, line 380: I wonder if this routine should be launch in initMaxFinishedJobs() as well:
1/ A user configures scheduler.state.max.finished.jobs to 1000, and used it so he reaches this threshold: Your implementation will work as expected, and limits the finishedJobs to 1000.
2/ The user update  scheduler.state.max.finished.jobs to 2000, and used it so he reaches this threshold: Your implementation will work as expected, and limits the finishedJobs to 2000.
3/ The user prefer to revert back this setting and set scheduler.state.max.finished.jobs back to 1000. Your implementation will still remain w/ 2000 finished Jobs, since it only removes one exceeding jobs when another one is set to be finished.
So ... shouldn't it be worthy to have all the finishedJobs exceeding the scheduler.state.max.finished.jobs threshold removed during initMaxFinishedJobs() to prevent that scenario from happening ?

fviale: this JVM setting is supposed to be defined at JVM initialization, not to be changed at runtime.
This is why the static field is private and initialized once. If the JVM is restarted, the new setting will be taken into account, and the number of finished jobs kept in SchedulerStateImpl will adapt accordingly. 
The scheduler actually loads 0 finished jobs from the database, the SchedulerStateImpl structure is populated only by jobs submitted after the start.

PR #3775, opened 2020-06-19
Generic Info REMOVE_DELAY to remove a job after a configurable time

fviale:
- Add functional test for housekeeping

- Also, fix issue with ServerJobAndTaskLogs while removing precious logs (access to SchedulerSpacesSupport from a non-active object context).
Fix minor issues appearing in the code

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
@@ -30,6 +30,7 @@ import java.security.KeyException;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.util.Collection;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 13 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -949,14 +954,28 @@ static RuntimeException launderThrowable(Throwable t) {
...
             if (SchedulingService.SCHEDULER_AUTO_REMOVED_JOB_DELAY > 0) {
-                long timeToRemove = System.currentTimeMillis() + SchedulingService.SCHEDULER_AUTO_REMOVED_JOB_DELAY;
+                removeDelay = SchedulingService.SCHEDULER_AUTO_REMOVED_JOB_DELAY;
+            }
+            if (jobGenericInfo != null && jobGenericInfo.containsKey(GENERIC_INFO_REMOVAL_DELAY)) {
+                try {
+                    removeDelay = Math.min(removeDelay,
+                                           Tools.formatDate(jobGenericInfo.get(GENERIC_INFO_REMOVAL_DELAY)));
+                } catch (Exception e) {
+                    logger.error("Error when parsing generic information " + GENERIC_INFO_REMOVAL_DELAY + " for job " +
```
mcompastie, line 971: I would advice to remind in the log the value of the default removal delay to ease the log exploitation in debugging session.

fviale: It is not necessary as I now print in INFO the exact data when the job is removed

aminelouati:
May be you can add a small paragraph in the doc about REMOVE_DELAY and its use

fviale:
@aminelouati sure, I will add it

PR #3777, opened 2020-06-23
Support the workflow variable to be defined as optional

tao-xinxiu:
Model Syntax is to add `?` at the end of the variable model `PA:MODEL?`
For example: `PA:URL?`, `PA:USER_FILE?`, `PA:DATETIME(yyyy-MM-dd)[2014, 2015]?`. Currently all the workflow variable model except PA:NOT_EMPTY_STRING support this syntax.

Optional variable accepts an blank string as the variable value. Only when the variable value is not blank, it is converted and validated by its original ParserValidator.

The expired variable models `PA:OPTIONAL_URI`, `PA:OPTIONAL_URL`, `PA:OPTIONAL_USER_FILE`, `PA:OPTIONAL_GLOBAL_FILE` are removed.


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/factory/OptionalParserValidator.java
```diff
@@ -25,21 +25,46 @@
...
+ * @author ActiveEon Team
+ * @since 14/08/19
+ */
+public class OptionalParserValidator<T> extends BaseParserValidator<T> {
+    @VisibleForTesting
+    BaseParserValidator<T> parentParserValidator;
 
-public class OptionalUserFileParserValidator extends UserFileParserValidator {
+    public OptionalParserValidator(String model, ModelType type) throws ModelSyntaxException {
+        super(removeSuffix(model), type, ".+");
```
fviale, line 50: I have the feeling this does not allow checking that the model string is valid.
For example, PA:DATETIME([? will produce a not understandable error

tao-xinxiu: This line `.+` allows all the string, but during generating the parentParserValidator `parentParserValidator = (BaseParserValidator<T>) ModelValidator.newParserValidator(type, this.model)` (here this.model is the model without suffix `?`), it will get the error `Model expression DATETIME([ doesn't match the expected pattern: ^(?i:DATETIME)\(([^)]+)\)$|^(?i:DATETIME)\(([^)]+)\)\[([^)]+)\]$`.

fviale: Ok, thanks for the clarification. PR looks very clean.

One word of caution though, as you removed existing PA:OPTIONAL.. I think you should do another PR in ow2-proactive/proactive-examples to change the corresponding workflows which were using these models, and merge them at the same time to avoid issues.

tao-xinxiu: Thanks for the reminder, currently none of the example workflows have used the optional variable.
I've just created the PR to remove the front-end UI where have used optional variables and to better handle variable null/undefined case.

PR #3780, opened 2020-06-29
Fix selection log max size

fviale:
This setting can overwrite pa.scheduler.job.logs.max.size=10MB from config/scheduler/settings.ini
Accordingly, they must be both equal


fviale:
Retest this please

PR #3784, opened 2020-07-06
Add tools to connect to HSQLDB CLI

mboussaa:


tools/hsqldb-connection/sqltool.rc
```diff
@@ -0,0 +1,7 @@
+urlid scheduler
+
+url jdbc:hsqldb:hsql://localhost:9001/scheduler
+
+username scheduler
+
+password
```
fviale, line 7: there should be a new line

tools/hsqldb-connection/db-connect.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# A Simple Shell Command To access HSQLDB CLI and connect by default to scheduler DB
+# If you would like to connect to another ProActive DB, please configure the sqltool.rc file before running the script
+# If java is not set on your system, please use java under [Schedulering_Path]/jre/bin/java
+
+java -jar sqltool-2.4.1.jar --rcFile=sqltool.rc scheduler
```
fviale, line 6: a new line is missing

tools/hsqldb-connection/db-connect.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
fviale, line 1: you should also create a corresponding .bat script

mboussaa: ok done. Could you check the .bat ?

tools/hsqldb-connection/db-connect.sh
fviale, line 6: actually for both windows and linux, you should use the jre present in ProActive

fviale: You can look at proactive-node.bat and proactive-node to see how the APP_HOME folder is determined.
Example in proactive-node.bat:
`set DIRNAME=%~dp0
if "%DIRNAME%" == "" set DIRNAME=.
set APP_BASE_NAME=%~n0
set APP_HOME=%DIRNAME%..`

PR #3786, opened 2020-07-16
Add getNodeTokens method to ResourceManage

fviale:
- additionally, check that provided tokens is not null or empty, as null tokens causes deep issues in the rm database and core.


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/core/RMCore.java
```diff
@@ -3144,6 +3159,9 @@ public void removeNodeToken(String nodeUrl, String token) throws RMException {
 
     @Override
     public void setNodeTokens(String nodeUrl, List<String> tokens) throws RMException {
+        if (tokens.stream().filter(token -> token == null || token.isEmpty()).findFirst().isPresent()) {
```
tao-xinxiu, line 3162: we can use `tokens.stream().anyMatch()` here.

fviale: yes, I pushed the modification

PR #3789, opened 2020-07-22
Fix housekeeping for large job list

fviale:
- partition job id list into smaller chunks of maximum 1000 items

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
@@ -40,6 +40,7 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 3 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -781,42 +782,48 @@ public void setTaskDataOwnerIfNull() {
     }
 
     public void executeHousekeepingInDB(final List<Long> jobIdList, final boolean shouldRemoveFromDb) {
-        executeReadWriteTransaction(new HousekeepingSessionWork(jobIdList, shouldRemoveFromDb));
+        List<List<Long>> jobIdSubSets = Lists.partition(jobIdList, 1000);
```
nebilbenmabrouk, line 785: is it possible to make the '1000' a configurable param, or at least a varibale in the class

fviale: Yes I willl look at this

fviale: it is now configurable

PR #3790, opened 2020-07-24
decode variables values when submitting workflow from URL

aminelouati:


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/util/WorkflowVariablesTransformer.java
```diff
@@ -25,17 +25,20 @@
...
 
+import com.google.common.base.Charsets;
 import com.google.common.collect.Maps;
 
 
 public class WorkflowVariablesTransformer {
 
-    public Map<String, String> getWorkflowVariablesFromPathSegment(PathSegment pathSegment) {
+    public Map<String, String> getWorkflowVariablesFromPathSegment(PathSegment pathSegment)
+            throws UnsupportedEncodingException {
```
mklkun, line 41: Why not a try/catch block?

PR #3793, opened 2020-07-30
Add server public url configuration

fviale:
Allows to adapt PASchedulerProperties.SCHEDULER_REST_URL and PASchedulerProperties.CATALOG_REST_URL when the ProActive server has a public url (for example, behind a reverse proxy or resolved with a domain name)


fviale:
Retest this please

PR #3794, opened 2020-07-31
Fix issues with PCA proxy

fviale:
- add jetty-util to server jars (class loader issue)
 - PCAProxyRule: handle the rare case where the endpoint path appears in the middle (and not the start) of the request uri

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/utils/PCAProxyRule.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/utils/PCAProxyRule.java
@@ -140,12 +140,22 @@ public class PCAProxyRule extends Rule implements Rule.ApplyURI {
                 logger.debug(String.format("Rewrote %s to %s", target, newTarget));
             }
             referrerCache.put(target, endpointPath);
-            if (request.getMethod().equals(HttpMethod.GET.asString())) {
+            if (HttpMethod.GET.is(request.getMethod())) {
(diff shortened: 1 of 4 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/utils/PCAProxyRule.java
```diff
@@ -140,12 +140,22 @@ private String rewriteTarget(String target, String endpointPath, HttpServletRequ
...
             referrerCache.put(target, endpointPath);
-            if (request.getMethod().equals(HttpMethod.GET.asString())) {
+            if (HttpMethod.GET.is(request.getMethod())) {
                 redirectGetRequest(target, endpointPath, request, response, newTarget);
             }
             return newTarget;
         } else {
             logger.trace("Target already contains endpoint");
+            if (!target.startsWith(endpointPath)) {
+                // endpoint is in the middle of the path (who did this?), let's bring it on top
```
MaelAudren, line 150: how do you get the endpoint ? Through the pca services ?

fviale: In the request directly, thanks to a regular expression (see previous pull request)

fviale: https://github.com/ow2-proactive/scheduling/commit/96ec7c0cc8dd3749937ce7df70572e453c5151ea

PR #3796, opened 2020-08-05
PCAProxyRule : improve uri handling

fviale:
- better handling of target containing the pca endpoint in the middle
 - added some unit tests

### DIFF
--- a/scheduler/scheduler-server/src/test/java/org/ow2/proactive/utils/PCAProxyRuleTest.java
+++ b/scheduler/scheduler-server/src/test/java/org/ow2/proactive/utils/PCAProxyRuleTest.java
@@ -1,0 +1,111 @@
+/*
+ * ProActive Parallel Suite(TM):
+ * The Open Source library for parallel and distributed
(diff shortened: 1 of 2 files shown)


scheduler/scheduler-server/src/test/java/org/ow2/proactive/utils/PCAProxyRuleTest.java
```diff
@@ -0,0 +1,115 @@
...
+            verify(response, times(1)).sendRedirect(redirectString.capture());
+            Assert.assertEquals(expectedTarget, redirectString.getValue());
+        } else {
+            verify(response, times(0)).sendRedirect(redirectString.capture());
+        }
+
+    }
+
+    @Test
+    public void applyURI() {
```
AlexFalessi, line 113: Is this test supposed to be empty ?

fviale: sorry, this is a mistake (and I will remove it probably)

PR #3804, opened 2020-10-08
Fill updated Variables and Models even when the workflow is not valid

fviale:


ShatalovYaroslav (approved):

mcompastie (approved):

aminelouati (approved):

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/util/ValidationUtil.java
```diff
@@ -107,6 +108,9 @@ public static JobValidationData validateJob(String jobFilePath, Map<String, Stri
             data.setTaskName(e.getTaskName());
             data.setErrorMessage(e.getMessage());
             data.setStackTrace(getStackTrace(e));
+            if (job != null && job instanceof TaskFlowJob) {
+                fillUpdatedVariables((TaskFlowJob) job, data);
```
amouhoub, line 112: why not do it in a finally block since calling fillUpdatedVariables requires the same conditions (not null & TaskFlowJob)

MaelAudren (approved):

amouhoub (approved):

PR #3805, opened 2020-10-13
Add configurable fetch mode for scripts

fviale:
Make fetch mode (lazy, eager) configurable. This allows, for example, script by reference in environments where http connection to the server is not guaranteed on the node side.

### DIFF
--- a/common/common-api/src/main/java/org/ow2/proactive/core/properties/PASharedProperties.java
+++ b/common/common-api/src/main/java/org/ow2/proactive/core/properties/PASharedProperties.java
@@ -79,9 +79,14 @@ public enum PASharedProperties implements PACommonProperties {
     /** comma-separated list of folder and/or files which need to be backup */
     SERVER_BACKUP_TARGETS("pa.server.backup.targets", PropertyType.STRING, "data,logs"),
 
-    /** backup mechanism may wait possible.delay (in seconds) untill all tasks are finished,
-     * if some task will still run then backup will not be performed */
-    SERVER_BACKUP_POSSIBLE_DELAY("pa.server.backup.possile.delay", PropertyType.INTEGER, "600");
+    /** backup mechanism may wait possible.delay (in seconds) until all currently running tasks are finished.
+     * if some tasks are still running after this delay, the backup will not be performed */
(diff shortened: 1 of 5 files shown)


common/common-api/src/main/java/org/ow2/proactive/core/properties/PASharedProperties.java
```diff
@@ -81,7 +81,12 @@
 
     /** backup mechanism may wait possible.delay (in seconds) untill all tasks are finished,
      * if some task will still run then backup will not be performed */
-    SERVER_BACKUP_POSSIBLE_DELAY("pa.server.backup.possile.delay", PropertyType.INTEGER, "600");
+    SERVER_BACKUP_POSSIBLE_DELAY("pa.server.backup.possile.delay", PropertyType.INTEGER, "600"),
```
mbenguig, line 84: possile ?

common/common-api/src/main/java/org/ow2/proactive/scripting/SimpleScript.java
```diff
@@ -82,7 +83,7 @@ public SimpleScript(File file, Serializable[] parameters) throws InvalidScriptEx
      */
     public SimpleScript(URL url, Serializable[] parameters) throws InvalidScriptException {
         // standard scripts (i.e. other than Selection scripts) use url late binding by default
-        super(url, parameters, false);
+        super(url, parameters, !PASharedProperties.LAZY_FETCH_SCRIPT.getValueAsBoolean());
```
mcompastie, line 86: Should it be worthy to put this eval in an attribute of Script abstract class ? => The property would be evaluate once (and failed at once if the user value isn't valid), and ease its re-usability in other formats class.

mcompastie (approved):

mbenguig (approved):
(cf my comment)

common/common-api/src/main/java/org/ow2/proactive/scripting/SimpleScript.java
fviale: Makes sense, Ok will change it.

common/common-api/src/main/java/org/ow2/proactive/core/properties/PASharedProperties.java
fviale: ugly indeed, I can change this

PR #3808, opened 2020-10-19
Improve Deployment Index handling

fviale:
Previous implementation was using an index counter always increasing.

Problem: this generate log files and data files which are big and can consume the disk space over time.

Solution: use a set of deployed indexes instead of a counter and find the first available index absent from this set

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
@@ -33,6 +33,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 
 import org.objectweb.proactive.core.config.CentralPAPropertyRepository;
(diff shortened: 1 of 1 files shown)


tao-xinxiu (approved):

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/LocalInfrastructure.java
```diff
@@ -486,9 +513,35 @@ private void decrementNumberOfLostNodesWithLockAndPersist() {
...
+            int maxIndex;
+            if (deployedNodeIndexes.isEmpty()) {
+                maxIndex = -1;
+            } else {
+                maxIndex = deployedNodeIndexes.last();
+            }
+            ContiguousSet<Integer> allIndexes = ContiguousSet.create(Range.closed(0, maxIndex + 1),
+                                                                     DiscreteDomain.integers());
+            TreeSet<Integer> availableIndexes = new TreeSet(Sets.difference(allIndexes, deployedNodeIndexes));
+            int newIndex = availableIndexes.first();
```
tao-xinxiu, line 532: What happens when availableIndexes is empty ?

fviale: it cannot be because allIndexes = [0..max(deployedNodeIndexes)+1] so availableIndexes contains at least max(deployedNodeIndexes)+1

mcompastie (approved):

PR #3810, opened 2020-10-28
Improve SchedulerFrontendState concurrency

fviale:
This improvement follows a dead lock observed in SchedulerFrontendState
 - use ConcurrentHashMap instead of global locks for user sessions. This reduces the lock complexity by having only a global schedulerStateLock
 - Additinnally, improve token checks in SelectionManager (due to a bug observed at Creos)
 - Add missing null checks on jobState
 - Use ReentrantReadWriteLock instead of synchronized for ClientJobState

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
@@ -509,6 +509,17 @@ public abstract class SelectionManager {
         return matched;
     }
 
+    private boolean isClientNodeUserAllPermission(Client client) {
+        try {
+            client.checkPermission(new NodeUserAllPermission(), client.getName() + " is not super-admin");
+            return true;
+        } catch (SecurityException e) {
+            // NO
+            logger.trace(e.getMessage());
+            return false;
+        }
+    }
+
     /**
      * Removes exclusion nodes and nodes not accessible for the client
      */
@@ -536,6 +547,10 @@ public abstract class SelectionManager {
         HashSet<Permission> clientPermissions = new HashSet<>();
         for (RMNode node : freeNodes) {
             try {
+                if (node.isProtectedByToken() && !nodeWithTokenRequested && !isClientNodeUserAllPermission(client)) {
+                    logger.debug("Node " + node.getNodeURL() + " is protected by token");
+                    continue;
+                }
                 if (!clientPermissions.contains(node.getUserPermission())) {
                     client.checkPermission(node.getUserPermission(),
                                            client + " is not authorized to get the node " + node.getNodeURL() +
--- a/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/UserIdentificationImpl.java
+++ b/scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/UserIdentificationImpl.java
@@ -57,26 +57,26 @@ public class UserIdentificationImpl extends UserIdentification {
     private Subject subject;
 
     /** Number of submit for this user */
-    private int submitNumber = 0;
+    private volatile int submitNumber = 0;
 
     /** Connection time of this user */
     private long connectionTime;
 
     /** last submit time */
-    private long lastSubmitTime = -1;
+    private volatile long lastSubmitTime = -1;
 
     /** Host name of this user. */
-    private String hostName;
+    private volatile String hostName;
 
     /** Tell if this user want to receive all events or only his events */
-    private boolean myEventsOnly = false;
+    private volatile boolean myEventsOnly = false;
 
     /** List of events that the user want to receive. */
-    private transient HashSet<SchedulerEvent> userEvents = null;
+    private transient volatile HashSet<SchedulerEvent> userEvents = null;
 
     /** Associated timerTask for user session management */
     //must be transient because useless on user side and TimerTask is not serializable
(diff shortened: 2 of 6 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
```diff
@@ -509,6 +509,17 @@ private void checkContentAuthorization(String content) {
         return matched;
     }
 
+    private boolean isClientNodeUserAllPermission(Client client) {
+        try {
+            client.checkPermission(new NodeUserAllPermission(), "Not Super-Admin");
+            return true;
+        } catch (SecurityException e) {
+            // NO
```
MaelAudren, line 517: NO ? could you add a more explicit message

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -1389,18 +1379,18 @@ public UserData getCurrentUserData() throws NotConnectedException {
...
+        for (UserAndCredentials userAndCredentials : identifications.values()) {
+            ListeningUser listeningUser = userAndCredentials.getListeningUser();
+            UserIdentificationImpl user = listeningUser.getUser();
+            users.add(new SchedulerUserInfo(user.getHostName(),
+                                            user.getUsername(),
+                                            user.getConnectionTime(),
+                                            user.getLastSubmitTime(),
+                                            user.getSubmitNumber()));
+        }
+        return users;
```
MaelAudren, line 1392: You can replace the upper block by:
```
return identifications.stream().map(userAndCredentials ->{
         ListeningUser listeningUser = userAndCredentials.getListeningUser();
         UserIdentificationImpl user = listeningUser.getUser();
         return new SchedulerUserInfo(user.getHostName(),
                                            user.getUsername(),
                                            user.getConnectionTime(),
                                            user.getLastSubmitTime(),
                                            user.getSubmitNumber()))
    }).collect(Collectors.toList());
```

MaelAudren (approved):

scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/UserIdentificationImpl.java
```diff
@@ -228,12 +228,20 @@ public TimerTask getSession() {
         return session;
     }
 
+    public synchronized void cancelSession() {
```
amouhoub, line 231: Add a JAVADOC method comment for this public method

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -993,7 +988,8 @@ private void dispatchSchedulerStateUpdated(SchedulerEvent eventType) {
             if (logger.isDebugEnabled()) {
                 logger.debug("event [" + eventType.toString() + "]");
             }
-            for (ListeningUser userId : identifications.values()) {
+            for (UserAndCredentials userAndCredentials : identifications.values()) {
+                ListeningUser userId = userAndCredentials.getListeningUser();
```
aminelouati, line 992: Maybe put the declaration of userId before the loop

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -1022,7 +1018,8 @@ private void dispatchJobSubmitted(JobState job) {
             if (logger.isDebugEnabled()) {
                 jlogger.debug(job.getJobInfo().getJobId(), " event [" + SchedulerEvent.JOB_SUBMITTED + "]");
             }
-            for (ListeningUser listeningUserId : identifications.values()) {
+            for (UserAndCredentials userAndCredentials : identifications.values()) {
+                ListeningUser listeningUserId = userAndCredentials.getListeningUser();
```
aminelouati, line 1022: same here

aminelouati (approved):

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
```diff
@@ -509,6 +509,17 @@ private void checkContentAuthorization(String content) {
         return matched;
     }
 
+    private boolean isClientNodeUserAllPermission(Client client) {
+        try {
+            client.checkPermission(new NodeUserAllPermission(), client.getName() + "is not super-admin");
```
mboussaa, line 514: add space before is not super-admin

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
aminelouati: To create a single ListeningUser object and initialize it each iteration.

fviale: it is not creation, it is declaration, which changes nothing if outside the loop. Each element of identifications contains a different user, so there is no interest in simply putting the declaration of userId outside the loop.

PR #3816, opened 2020-11-06
handle empty passwords

fviale:
as we observed a case where an empty password could wrongly allow arbitrary connections from user to some AD server, we need to prevent empty passwords in the ldap module code.


amouhoub (approved):

nebilbenmabrouk (approved):
LGTM, I hope that no one will use accidently the same password as the Fake password :)

fviale:
@nebilbenmabrouk hope so too but I estimate 1 chance on 1 billion :-)

MaelAudren (commented):
That would be better if this could configurable by the sysadmin. As Github is open source it would be easy for a hacker to find out the password and to test it with an account.

MaelAudren:
Instead of sending an emtpy password with ldap which could allow the connection. A bad password will be sent which will ensure that the password check is done.

PR #3825, opened 2020-11-27
Change package.loaded location

fviale:


tao-xinxiu (approved):

amouhoub (approved):

amouhoub:
@fviale what is the purpose of changing its location to a subfolder ?

aminelouati (approved):

fviale:
@amouhoub good question as it's not obvious. This is because of the need to containerize the PA server. When we restart the container, we need to keep the status of package.loaded. Unfortunately, it is not possible to mount a file in docker which does not exist yet. In that case, docker is going to create a directory. Thus we cannot do -v /otherpath/to/packages.loaded:/path/to/packages.loaded
We need to have a containing folder to mount, which will be populated by the container.

Additionnaly, we cannot mount the samples folder globally. If we do that, all the contents of the samples folder originally inside the container will be ignored and emptied. Meaning, the container will not be able to expand the proactive_examples.zip file (file disappeared !).

So it's a complex problem which can only be resolved with this change.

PR #3834, opened 2020-12-10
Fix-native-shell-pipe-issue

nebilbenmabrouk:


medou-boushab (approved):

common/common-api/src/main/java/org/ow2/proactive/utils/BoundedStringWriter.java
```diff
@@ -61,19 +61,21 @@ public BoundedStringWriter(PrintStream outputSink, final int maxSize) {
     /** {@inheritDoc} */
     @Override
     public void write(char[] c, int off, int len) {
-        int nextLen = this.content.length() + len;
-        if (nextLen <= this.maxSize) {
-            this.content.append(c, off, len);
-        } else { // nextLen > this.maxSize    		
-            if (len <= this.maxSize) {
-                this.content.delete(0, nextLen - this.maxSize);
+        if (len > 0 && off >= 0) {
```
mklkun, line 64: In this case we do not write anything if `off < 0`. Should we not just start from 0?

mklkun (approved):
One single question.

common/common-api/src/main/java/org/ow2/proactive/utils/BoundedStringWriter.java
nebilbenmabrouk: This what we are doing:
- We do not write anything if off < 0
- We write if off >=0

fviale (approved):

PR #3835, opened 2020-12-13
Add variables to multipart when submitting jobs via POST query

mbenguig:


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -1359,13 +1411,12 @@ public JobIdData submit(String sessionId, PathSegment pathSegment, MultipartForm
...
-            Map<String, List<InputPart>> formDataMap = multipart.getFormDataMap();
-
-            String name = formDataMap.keySet().iterator().next();
             File tmpJobFile = null;
             try {
 
-                InputPart part1 = multipart.getFormDataMap().get(name).get(0); // "file"
+                List<InputPart> jobVariablesListInputPart = multipart.getFormDataMap().remove(VARIABLES_KEY);
+
+                InputPart part1 = multipart.getFormDataMap().values().iterator().next().get(0); // "file"
```
fviale, line 1419: I don't understand why you still need to use the iterator here.

Why not simply do 
multipart.getFormDataMap().get(VARIABLES_KEY)
and
multipart.getFormDataMap().get(FILE_KEY)
?

mbenguig: You're right, since "file" is clearly specified as comment

mbenguig: After some tests, this key can either be "job.xml"  (submit from the studio) or "file" (submit from jobplanner)  ... the comment was wrong. And not sure there are no more keys. I'd prefer this prefer to not break anything.

fviale (approved):

PR #3837, opened 2020-12-18
Add WebProperties to rest scheduler/properties endpoint

fviale:
also, fix the NoVNC startup url.


mboussaa (approved):

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -2370,6 +2371,8 @@ public boolean changeStartAt(String sessionId, String jobId, String startAt) thr
         Map<String, Object> schedulerProperties;
         try {
             schedulerProperties = scheduler.getSchedulerProperties();
+            schedulerProperties.putAll(WebProperties.getPropertiesAsHashMap());
+            schedulerProperties.putAll(PortalConfiguration.getPropertiesAsHashMap());
```
MaelAudren, line 2375: Did you check that the properties do not have the same key ?

MaelAudren (approved):

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
fviale: yes they never have the same key

PR #3840, opened 2020-12-30
Fail fast acquireNodes request when ns not exist or not deployed

tao-xinxiu:
- Fail fast when REST acquire nodes request with a node source not exist or not deployed;
- Add the temporary tag of rest request id with a prefix "tmp:".


mcompastie (approved):
LGTM

AlexFalessi (approved):

nebilbenmabrouk (approved):

PR #3841, opened 2021-01-06
Enable the creation of nested folders in the global space

Hiba-Alili:


tools/LoadPackage.groovy
```diff
@@ -115,7 +115,19 @@ class LoadPackage {
             def file_src_path = file_src.absolutePath
             def file_name = file_src.getName()
             def file_dest = new File(target_dir_path, file_name)
-            def file_dest_path = file_dest.absolutePath
+            def file_dest_path
+            def dataspace_matcher = Pattern.compile("dataspace").matcher(file_relative_path)
```
fviale, line 119: I don't see the reason to use a pattern matcher here, this seems complicated and error prone

When I look at the packages metadata, I see that it contains relative paths like:

resources/dataspace/assets.txt
resources/dataspace/someFolder/someFile

So if the extracted archive contains 

/archive/temporary/location/resources/dataspace/someFolder/someFile

We want to copy this into

/global/space/location/someFolder/someFile

In order to do that, you can remove for example from the metadata relative path, the prefix resources/dataspace/ te determine the destination.

In order to create the destination parent folders, you can use:

import org.apache.commons.io.FileUtils;

destFile = new File(destinationPath)
FileUtils.forceMkDirParent(destFile)
FileUtils.copyFile(sourceFile, destFile)

fviale (commented):
the changes to admin.cred and the addition of proactive-examples.zip must not be commited

tools/LoadPackage.groovy
```diff
@@ -114,8 +115,10 @@ class LoadPackage {
             def file_src = new File(package_dir.absolutePath, file_relative_path)
             def file_src_path = file_src.absolutePath
             def file_name = file_src.getName()
-            def file_dest = new File(target_dir_path, file_name)
-            def file_dest_path = file_dest.absolutePath
+            def file_dest = file_relative_path.replace("resources/dataspace/", "")
```
fviale, line 118: this code is much simpler, thanks a lot

aminelouati (approved):

fviale (approved):

PR #3844, opened 2021-01-15
Improve unauthorized selection script behavior.

fviale:
Instead of throwing an exception, check script authorization gracefully.

Add additional test.

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
@@ -262,58 +262,60 @@ public abstract class SelectionManager {
         List<Node> matchedNodes;
         if (hasScripts) {
             // checking if all scripts are authorized
-            checkAuthorizedScripts(criteria.getScripts());
+            if (checkAuthorizedScripts(criteria.getScripts())) {
 
-            // arranging nodes for script execution
-            List<RMNode> arrangedNodes = arrangeNodesForScriptExecution(afterPolicyNodes,
-                                                                        criteria.getScripts(),
-                                                                        criteria.getBindings());
-            List<RMNode> arrangedFilteredNodes = arrangedNodes;
-            if (criteria.getTopology().isTopologyBased()) {
-                arrangedFilteredNodes = topologyNodesFilter.filterNodes(criteria, arrangedNodes);
-            }
+                // arranging nodes for script execution
+                List<RMNode> arrangedNodes = arrangeNodesForScriptExecution(afterPolicyNodes,
+                                                                            criteria.getScripts(),
+                                                                            criteria.getBindings());
+                List<RMNode> arrangedFilteredNodes = arrangedNodes;
+                if (criteria.getTopology().isTopologyBased()) {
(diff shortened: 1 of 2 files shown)


AlexFalessi (approved):

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
```diff
@@ -428,22 +430,26 @@ private static boolean electedToRunOnAllNodes(Criteria criteria) {
...
      * Checks is all scripts are authorized. If not throws an exception.
      */
-    private void checkAuthorizedScripts(List<SelectionScript> scripts) {
+    private boolean checkAuthorizedScripts(List<SelectionScript> scripts) {
         updateAuthorizedScriptsSignatures();
         if (authorizedSelectionScripts == null || scripts == null)
-            return;
+            return true;
+
+        boolean allAuthorized = true;
```
nebilbenmabrouk, line 438: "boolean allAuthorized = true;" should be before "return true;"

nebilbenmabrouk (approved):

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
```diff
@@ -262,58 +262,60 @@ private NodeSet doSelectNodes(Criteria criteria, Client client) {
...
+                        }
+
+                        List<RMNode> subset = arrangedFilteredNodes.subList(0,
+                                                                            Math.min(numberOfNodesForScriptExecution,
+                                                                                     arrangedFilteredNodes.size()));
+                        matchedNodes.addAll(runScripts(subset, criteria));
+                        // removing subset of arrangedNodes
+                        subset.clear();
+
+                        if (arrangedFilteredNodes.size() == 0) {
```
MaelAudren, line 309: Why not putting this condition in the while loop ? Usually it is better to avoid using break statement if you can put all conditions in the loop.

MaelAudren (approved):

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
MaelAudren: Moving the statement will add an useless allocation before finishing the method. Why are you pushing to do that @nebilbenmabrouk ?

nebilbenmabrouk: @MaelAudren because no instructions will be executed after the return statement

nebilbenmabrouk: @MaelAudren sorry I thought that both instructions are in the same block (because of the missing brackets).

Xiaoven:
```java
TestUtil.testCall(db, "return custom.answer(42,3.14,'foo',{a:1},[1],true,date(),datetime(),point({x:1,y:2})) as data", (row) -> assertEquals(9, ((List)row.get("data")).size()));
```

PR #3856, opened 2021-02-07
Change Signal API to send any signal only once

nebilbenmabrouk:
1- If a ready signal already exists, it will not be added again.
2- If any other signal already exists, it will not be added again.
3- If a signal is sent, its associated ready signal will be removed


fviale (commented):
Why not use HashSet instead of ArrayList ?

in

synchronization.putIfAbsent(SIGNALS_CHANNEL, jobId, new ArrayList<>());

This simplifies a lot the code and allows atomicity

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
```diff
@@ -121,7 +126,18 @@ public String checkForSignals(List<String> signalsSubList) throws SignalApiExcep
     public boolean sendSignal(String signalName) throws SignalApiException {
         try {
             init();
-            synchronization.compute(SIGNALS_CHANNEL, jobId, "{k, x -> x.add('" + signalName + "');x}");
+
+            // If the ready signal already exists, remove it
+            synchronization.conditionalCompute(SIGNALS_CHANNEL,
```
fviale, line 131: if you use HashSet, you can simplify this in one line

synchronization.compute(SIGNALS_CHANNEL, jobId, x -> x.remove('" + READY_PREFIX + signalName + "'); x -> x.add('" + signalName + "');x}");

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
```diff
@@ -134,10 +150,22 @@ public boolean sendSignal(String signalName) throws SignalApiException {
...
         try {
             init();
+            List<String> signals = getJobSignals();
             StringBuilder actions = new StringBuilder();
-            for (String signal : signalsSubList) {
+
+            // Filter signals to add only those signals that are not already sent
+            for (String signal : signalsSubList.stream()
+                                               .filter(signal -> !signals.contains(signal))
+                                               .collect(Collectors.toList())) {
```
fviale, line 159: again, if you use HashSet instead of ArrayList, the code will be much simpler

 actions.append("x.remove('" + READY_PREFIX + signal + "');x.add('" + signal + "');");

nebilbenmabrouk:
> Why not use HashSet instead of ArrayList ?
> 
> in
> 
> synchronization.putIfAbsent(SIGNALS_CHANNEL, jobId, new ArrayList<>());

OK I'll consider that.

amouhoub:
@nebilbenmabrouk Just one question, when you remove the associated "ready_" signals after you receive a signal,,, what about other "ready_" signals that were added by another task ? Do we support multiple signal sets (via multiple tasks) in parallel ? It doesn't look to be the case from the code right ? What about this scenario ?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
```diff
@@ -1733,11 +1733,11 @@ private JobInfo insertJobSignals(JobInfo jobInfo) {
 
             if (publicStore.containsKey(SIGNAL_ORIGINATOR, SIGNAL_TASK_ID, signalsChannel, jobid)) {
 
-                List<String> jobSignals = jobInfo.getSignals();
-                List<String> signalsToBeAdded = (List) publicStore.get(SIGNAL_ORIGINATOR,
-                                                                       SIGNAL_TASK_ID,
-                                                                       signalsChannel,
-                                                                       jobid);
+                Set<String> jobSignals = jobInfo.getSignals();
+                Set<String> signalsToBeAdded = (HashSet) publicStore.get(SIGNAL_ORIGINATOR,
```
fviale, line 1737: you should cast to Set

fviale (approved):
LGTM

nebilbenmabrouk:
> @nebilbenmabrouk Just one question, when you remove the associated "ready_" signals after you receive a signal,,, what about other "ready_" signals that were added by another task ? Do we support multiple signal sets (via multiple tasks) in parallel ? It doesn't look to be the case from the code right ? What about this scenario ?

@amouhoub Each job has a unique signal set, we DO NOT have multiple signal sets in parallel.
If a "ready_" signal is removed and another task sends another "ready_", it is the responsibility of the workflow implementer to ensure that signals are handled properly. Actually, in all our use cases, one signal is handled by only one task in the workflow. We do not have many tasks that deal with the same signal.

rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/data/JobInfoImpl.java
```diff
@@ -25,9 +25,7 @@
  */
 package org.ow2.proactive.scheduler.rest.data;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
amouhoub, line 28: Are you sure you need all utils here ? You just replaced all the three (List, AL & Map) with just two (Set & Hashset) ?

amouhoub (approved):

rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/data/JobInfoImpl.java
nebilbenmabrouk: Yes I'm sure :), thanks for the review

PR #3859, opened 2021-02-14
Remove ready signals when a signal is received in control templates

nebilbenmabrouk:
Control templates are: 

- Check_For_Signals

- Wait_For_Signals

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
@@ -110,7 +110,7 @@ public class SignalApiImpl implements SignalApi {
         HashSet<String> signals = getJobSignals();
 
         if (!signals.isEmpty()) {
-            return signalsSubSet.stream().filter(signals::contains).findFirst().get();
+            return signalsSubSet.stream().filter(signals::contains).findFirst().orElse(null);
         } else {
             return null;
         }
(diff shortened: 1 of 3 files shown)


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
```diff
@@ -108,17 +109,26 @@ public boolean isReceived(String signalName) throws SignalApiException {
...
         HashSet<String> signals = getJobSignals();
+        String receivedString = null;
         if (!signals.isEmpty()) {
-            return signalsSubSet.stream().filter(signals::contains).findFirst().orElse(null);
-        } else {
-            return null;
+            receivedString = signalsSubSet.stream().filter(signals::contains).findFirst().get();
+            // Remove ready signals if a signal is received
+            if (receivedString != null) {
+                removeManySignals(signalsSubSet.stream()
```
fviale, line 117: we mentioned that for this case, it is the template workflow which should remove the signals, not the api.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
```diff
@@ -108,17 +109,26 @@ public boolean isReceived(String signalName) throws SignalApiException {
...
         }
+        return receivedString;
     }
 
     @Override
     public boolean sendSignal(String signalName) throws SignalApiException {
         try {
             init();
+            if (!isReceived(READY_PREFIX + signalName)) {
+                throw new SignalApiException("The job is not ready for the signal " + signalName);
```
fviale, line 130: "Job " + jobId + " is not ready to receive signal " + signalName

fviale: please document in the javadoc that this exception is thrown when the ready signal is not there.

fviale (approved):

amouhoub (approved):

PR #3860, opened 2021-02-19
Return the set of job signals when a signal is sent

nebilbenmabrouk:


aminelouati (approved):

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
```diff
@@ -117,7 +117,7 @@ public String checkForSignals(Set<String> signalsSubSet) throws SignalApiExcepti
     }
 
     @Override
-    public boolean sendSignal(String signalName) throws SignalApiException {
+    public Set<String> sendSignal(String signalName) throws SignalApiException {
```
fviale, line 120: why not do the same for the other methods sendManySignals, removeSignal to have coherence in the api ?

nebilbenmabrouk: Because  **sendSignal**  is the only operation that can be performed in REST mode and needed by the portal.

fviale: it is not really an argument. The java api needs to have coherence in itself

fviale: at least sendManySignals should return the list of signals

fviale: In addition, I see nothing in this pull request related to the rest endpoint job/{jobid}/signals. So what are trying to achieve ?

nebilbenmabrouk: You're right, I made a mistake, **sendSignal** is not the method used by the REST endpoint. I'll change the PR.
Thanks @fviale

PR #3863, opened 2021-02-22
REST signal endpoint (i) returns the set of signals and (ii) throws an error if the ready signal does not exist

nebilbenmabrouk:
The add signal endpoint is changed to :
- return the set of signals 
- throw an error if the ready signal does not exist

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
@@ -66,8 +66,7 @@ public class SignalApiImpl implements SignalApi {
     private void init() throws SignalApiException {
         if (!isInitialized) {
             try {
-                // Initialize synchronization signals channel
-                synchronization.createChannelIfAbsent(SIGNALS_CHANNEL, true);
+                // Initialize job signals
                 synchronization.putIfAbsent(SIGNALS_CHANNEL, jobId, new HashSet<String>());
             } catch (IOException | InvalidChannelException e) {
                 throw new SignalApiException("Could not instantiate Signal API for the job " + jobId, e);
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
@@ -50,7 +50,7 @@ public class SignalApiImpl implements SignalApi {
 
     private static final String SIGNALS_CHANNEL = PASchedulerProperties.SCHEDULER_SIGNALS_CHANNEL.getValueAsString();
 
-    protected static final String READY_PREFIX = "ready_";
+    public static final String READY_PREFIX = "ready_";
 
     private SynchronizationWrapper synchronization;
 
@@ -120,9 +120,6 @@ public class SignalApiImpl implements SignalApi {
     public boolean sendSignal(String signalName) throws SignalApiException {
         try {
             init();
-            if (!isReceived(READY_PREFIX + signalName)) {
-                throw new SignalApiException("Job " + jobId + " is not ready to receive the signal " + signalName);
-            }
             // Remove the ready signal if it already exists, then add the signal if it does not exist
             synchronization.compute(SIGNALS_CHANNEL,
                                     jobId,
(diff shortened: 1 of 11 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
```diff
@@ -1915,20 +1918,37 @@ public void endActivity(Body body) {
...
             if (signals == null) {
                 signals = new HashSet<>();
             }
 
+            // Throw an error if: (i) the corresponding ready signal does not exist and (ii) the signal is not a ready signal
+            String readyPrefix = SignalApiImpl.READY_PREFIX;
+            if (!signals.contains(readyPrefix + signal) && !signal.startsWith(readyPrefix)) {
+                throw new SignalApiException("Job " + jobId + " is not ready to receive the signal " + signal);
+            }
+
```
fviale, line 1937: to improve atomicity you can do:
` Set<String> signals = (HashSet<String>) publicStore.get(SIGNAL_ORIGINATOR,
                                                                    SIGNAL_TASK_ID,
                                                                    signalsChannel,
                                                                    jobId);
if (!signals.contains(readyPrefix + signal) && !signal.startsWith(readyPrefix)) {
                throw new SignalApiException("Job " + jobId + " is not ready to receive the signal " + signal);
}                                                                    
`
return publicStore.compute(SIGNAL_ORIGINATOR, SIGNAL_TASK_ID, signalsChannel, jobId, "{ k, x ->x.remove('" + READY_PREFIX + signalName + "'); x.add('" + signalName + "'); x}");

fviale: additionnally, is it meaningful to allow sending a ready signal, like in the test signal.startsWith(readyPrefix) ?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
```diff
@@ -1915,20 +1918,31 @@ public void endActivity(Body body) {
 
     @Override
     @ImmediateService
-    public boolean addJobSignal(String sessionId, String jobId, String signal) {
+    public Set<String> addJobSignal(String sessionId, String jobId, String signal) throws SignalApiException {
         try {
+            publicStore.putIfAbsent(SIGNAL_ORIGINATOR, SIGNAL_TASK_ID, signalsChannel, jobId, new HashSet<String>());
```
nebilbenmabrouk, line 1923: @fviale 
I added this line because when the user wants to send a ready signal for the first time (i.e., the set of  job signals is null), we have to initialize the set of  job signals with a HashSet<String>.

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
```diff
@@ -66,8 +66,7 @@ public SignalApiImpl(String originator, TaskId taskId, SynchronizationInternal s
     private void init() throws SignalApiException {
         if (!isInitialized) {
             try {
-                // Initialize synchronization signals channel
-                synchronization.createChannelIfAbsent(SIGNALS_CHANNEL, true);
+                // Initialize job signals
```
nebilbenmabrouk, line 69: @fviale 
I removed the instruction **synchronization.createChannelIfAbsent(SIGNALS_CHANNEL, true);** because it is useless.
The signals channel is automatically created in the class **SchedulerFrontend**  in the method **initActivity()** -> **startSynchronizationService()**. Thus, avoiding this instruction each time the signal API is used.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
fviale: ok

fviale (approved):

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/signal/SignalApiImpl.java
fviale: ok

aminelouati (approved):

PR #3865, opened 2021-02-24
Deny sending empty signals

nebilbenmabrouk:


fviale (approved):
LGTM

scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/signal/SignalApiTest.java
```diff
@@ -172,6 +180,14 @@ public void testSendSignal() throws InvalidChannelException, SignalApiException
                                                                                    .count());
     }
 
+    @Test(expected = SignalApiException.class)
+    public void testSendSignalWithException() throws SignalApiException {
+        String signal = "";
+
+        // Send an empty signal then assert a SignalAPIException is thrown
+        signalApi.sendSignal(signal);
+    }
```
aminelouati, line 189: do you include whitespaced string (ex: "   ") as not allowed signals ? 
In that case maybe apply the trim() function ?

aminelouati (approved):
1 comment left

PR #3868, opened 2021-03-01
Fix jobsinfolist endpoint performance by using dbmanager

amouhoub:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
```diff
@@ -1634,15 +1634,9 @@ private void insertSignals(Page<JobInfo> jobsInfo) {
...
-        for (String jobId : jobsId) {
-            try {
-                jobsInfoList.add(this.getJobInfo(jobId));
-            } catch (UnknownJobException e) {
-                logger.warn("The job with job ID " + jobId + " couldn't be found");
-            }
-        }
-        return jobsInfoList;
+        frontendState.checkPermission("getJobs", "You don't have permissions to load jobs");
+        frontendState.checkPermission("getJobs", "You don't have permissions to load jobs");
```
fviale, line 1638: duplicated line

fviale (approved):
LGTM, don't forget to squash

aminelouati (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -273,6 +273,24 @@ public SchedulerDBManager(Configuration configuration, boolean drop) {
         return new Page<>(lJobs, totalNbJobs);
     }
 
+    public List<JobInfo> getJobs(final List<String> jobIds) {
+        List<Long> longJobIds = jobIds.stream().map(id -> Long.parseLong(id)).collect(Collectors.toList());
```
AlexFalessi, line 277: You could add `.distinct()` to evict duplicate ids.

fviale: makes sense, indeed

amouhoub: I'm afraid it might affect performance (see the distinct() documentation ) and currently it doesn't really matter as the frontend will handle this very well

AlexFalessi: To avoid using distinct you could also work with a Set.
`Set<Long> longJobIds = jobIds.stream().map(id -> Long.parseLong(id)).collect(Collectors.toSet());`

But if you are confident enough with front-end dealing with duplicates it works with me

PR #3875, opened 2021-03-15
Fix ReasyEasy client gzip limitation

fviale:


tao-xinxiu (approved):

rest/rest-server/src/main/webapp/WEB-INF/web.xml
```diff
@@ -18,7 +18,7 @@
     </context-param>
     <context-param>
         <param-name>resteasy.gzip.max.input</param-name>
-        <param-value>500000000</param-value>
+        <param-value>1000000000</param-value>
```
nebilbenmabrouk, line 21: Have you tested an archive with such a size in order to detect eventual performance issues ?

nebilbenmabrouk (approved):

rest/rest-server/src/main/webapp/WEB-INF/web.xml
fviale: I have not tested such a size. It is not an archive in case of gzip, but anything else. Archives like zip are not compressed during the transfer.

PR #3882, opened 2021-03-24
Further improvements in AOSynchronization concurrency

fviale:
Make waitUntil read requests an ImmediateService. waitUntil requests will run in a dedicated thread and execute the predicate periodically. waitUntilThen requests will remain executed by the active object runActivity (as they perform changes on the channel)

Additionally, improve SignalApiImpl initialization behavior to reduce the number of put requests sent to the synchronization api

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/synchronization/AOSynchronization.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/synchronization/AOSynchronization.java
@@ -101,6 +101,8 @@ public class AOSynchronization implements RunActive, InitActive, EndActive, Sync
     /** Schema used inside the JDBM database */
     private static final String STATUS_RECORD_NAME = "STORE";
 
+    private static long WAIT_STEP = 500L;
+
     /** HashMap storing the in-memory channels */
     private ConcurrentHashMap<String, Channel> inMemoryChannels;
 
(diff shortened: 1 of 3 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/synchronization/AOSynchronization.java
```diff
@@ -747,17 +773,43 @@ public boolean waitUntil(String originator, TaskId taskid, String channel, Strin
...
+         * it runs its predicate periodically until success or timeout
+         */
+        waitUntilStarted();
+        Channel chosenChannel = getChannel(channel);
+        Serializable value = chosenChannel.get(key);
+        long startTime = System.currentTimeMillis();
+        try {
+            while (!evaluateClosure(predicate, BiPredicate.class).test(key, value)) {
+                if (System.currentTimeMillis() - startTime > timeout) {
+                    throw new TimeoutException("Timeout of " + timeout + " ms expired while waiting for predicate");
```
nebilbenmabrouk, line 790: Throwing an exception may prevent the Signal API and the workflows from working correctly. Are you sure this is useful?

fviale: I did not modify the behavior of the api, only its implementation. In the previous implementation a TimeoutException would have been thrown if the timeout was exceeded. The signal api does not use timeouts.

nebilbenmabrouk (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/synchronization/AOSynchronization.java
nebilbenmabrouk: OK, thanks for the clarifications

tao-xinxiu (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/synchronization/AOSynchronization.java
```diff
@@ -101,6 +101,8 @@
     /** Schema used inside the JDBM database */
     private static final String STATUS_RECORD_NAME = "STORE";
 
+    private static long WAIT_STEP = 500L;
```
mboussaa, line 104: (optional) probably add comment above to explain

mboussaa (approved):

PR #3883, opened 2021-03-25
Fix issue in licence scheduling policy (when in error restart enabled)

mbenguig:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -171,26 +192,29 @@ private boolean acquireTokensForJob(JobDescriptorImpl job) {
...
                     return false;
                 }
 
-                if (!licenseSynchronization.containsJobId(requiredLicense, jobId) && !canGetToken(requiredLicense)) {
-                    logger.debug("Token for " + requiredLicense + " not available, keep job pending");
+                jobHoldsToken = licenseSynchronization.containsJobId(requiredLicense, jobId);
+                canGetToken = canGetToken(requiredLicense);
+                if (!jobHoldsToken && !canGetToken) {
+                    logger.debug("Job does not already hold a token for " + requiredLicense +
+                                 " and cannot get a new one, keep job pending");
```
fviale, line 207: good improvement, if a job does not already hold or can acquire a token for all required licences, then you immediately return false without acquiring tokens partially

aminelouati (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
mbenguig: This is a remark or a new condition to consider ?

fviale (approved):
LGTM

PR #3887, opened 2021-04-01
Add POST jobsInfoList endpoint

fviale:
Additionally, improve email notification template.

### DIFF
--- a/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
+++ b/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
@@ -276,6 +276,11 @@ public class SchedulerStateRest implements SchedulerRestInterface {
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


mboussaa (approved):

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -275,6 +275,19 @@ public boolean checkJobPermissionMethod(String sessionId, String method, String
         }
     }
 
+    @Override
```
amouhoub, line 278: @fviale Maybe you could simply call the same function as above (jobsInfoListPost(String sessionId, List<String> jobsId))
Because they are exactly the same, no diff at all, only the name changes.
After all, the GET & POST endpoints should behave the same and provide the same outputs for the same input which is a list of job ids. Only the HTTP method changes.

fviale: you're right, I will update my PR

medou-boushab (approved):

PR #3891, opened 2021-04-08
Fix result propagation for replicate skipped block

fviale:
- consider if/else/continuation dependency when searching for non-skipped parent tasks
 - use LinkedHashSet/Map to ensure parent task ordering
 - additionally, fix matching block search when skipping blocks

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
@@ -31,6 +31,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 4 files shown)


oanaBiancaSchiopu (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
```diff
@@ -1229,19 +1230,40 @@ public boolean equals(Object obj) {
...
+            }
+            if (joinedBranches != null) {
+                for (InternalTask parentTask : joinedBranches) {
+                    parentIds.add(parentTask.getId());
+                }
+            }
+
+            parentTasksResults = new LinkedHashMap<>();
+
+            if (internalTasksDependencies != null) {
```
amouhoub, line 1253: You have the same if twice if (internalTasksDependencies != null)
Maybe you can manage to merge the instructions inside them in a single if block

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
```diff
@@ -1229,19 +1230,40 @@ public boolean equals(Object obj) {
...
+                // Load parent task results from memory
+                for (InternalTask parentTask : internalTasksDependencies) {
```
amouhoub, line 1255: same for the loop

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
fviale: done

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/task/internal/InternalTask.java
fviale: done

PR #3892, opened 2021-04-09
Improve error reporting in script engines

fviale:
exception cause was stripped as it may not be serializable.

Instead, wrap the exception as a string


aminelouati (approved):

oanaBiancaSchiopu (approved):

tao-xinxiu (approved):

nebilbenmabrouk (approved):
LGTM

PR #3894, opened 2021-04-14
Add download task result endpoint

fviale:


AlexFalessi (approved):

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -1015,24 +1038,64 @@ private String getTaskResultValueAsStringOrExceptionStackTrace(TaskResult taskRe
...
+            resultMetadata = new HashMap<>();
+        }
+        if (destination == null) {
+            destination = DESTINATION_BROWSER;
+        }
+
+        Response.ResponseBuilder builder = Response.ok().entity(new ByteArrayInputStream(serializedValue));
+
+        if (resultMetadata.containsKey(METADATA_CONTENT_TYPE)) {
+            builder = builder.header(HttpHeaders.CONTENT_TYPE, resultMetadata.get(METADATA_CONTENT_TYPE));
```
amouhoub, line 1067: @fviale what is no content-type is provided ? Could you throw an exception with an explicit and clear error message ?

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -1015,24 +1038,64 @@ private String getTaskResultValueAsStringOrExceptionStackTrace(TaskResult taskRe
...
+        }
+        if (DESTINATION_FILE.equals(destination)) {
+            if (resultMetadata.containsKey(METADATA_FILE_NAME)) {
+                builder = builder.header(HttpHeaders.CONTENT_DISPOSITION,
+                                         "attachment; filename=" + resultMetadata.get(METADATA_FILE_NAME));
+            } else if (resultMetadata.containsKey((METADATA_FILE_EXTENSION))) {
```
amouhoub, line 1073: what if both METADATA_FILE_NAME &  METADATA_FILE_EXTENSION are provided in the following fashion :
fileName = "myFile", extension = "pdf" ?
In this case, you'll download a file called myFile without the proper extension. If it's an executable or any other type of file without a standard content-type then users will have to rename the file after it's downloaded to get the extension right.

Maybe you can simply do :

```
if (DESTINATION_FILE.equals(destination)) {
    String fileExtension= resultMetadata.containsKey(METADATA_FILE_EXTENSION)?resultMetadata.get(METADATA_FILE_EXTENSION):"";
    String fileName = resultMetadata.containsKey(METADATA_FILE_NAME)?resultMetadata.get(METADATA_FILE_NAME):"job_" + jobId + "_" + taskname + "_result";
    builder = builder.header(HttpHeaders.CONTENT_DISPOSITION,
                                          "attachment; filename= filename+fileExtension;
         }
```

fviale: METADATA_FILE_NAME takes precedence (it must include the extension)

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
fviale: no content-type is standard, it is when resultMetada is not filled. In that case I should return application/octet-stream (I will add this)

fviale: alternatively, I can return a byte stream of the result toString and set the content type to text/plain, it could make it easier to handle then.

amouhoub: Yes, I meant when resultMetada is not filled. Should we make it mandatory to fill it or return a bytearray by default ?

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
amouhoub: Maybe we can check when both are present but the filename does or doesn't contain an extension ? Just in case resultMetadata is misused.
Something like 'if (fileName.endsWith(fileExtension))' ?

fviale: It is explained in the documentation
https://doc.activeeon.com/latest/user/ProActiveUserGuide.html#_assigning_metadata_to_task_result

I am not going to handle misuse.

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
fviale: I updated the PR to handle everything related to result. Now this endpoint can be called directly without using any other endpoints (task failed, no metadata, etc).

amouhoub (approved):

PR #3895, opened 2021-04-15
Add a method waitForAll to Signal API

nebilbenmabrouk:


scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/signal/SignalApiTest.java
```diff
@@ -310,6 +310,34 @@ public void testWaitForAny() throws SignalApiException {
...
+    public void testWaitForAll() throws SignalApiException {
+        String signal_1 = "test_signal_11_1";
+        String signal_2 = "test_signal_11_2";
+        Set<String> signals = new HashSet<String>() {
+            {
+                add(signal_1);
+                add(signal_2);
+            }
+        };
+        long durationInMillis = 0;
```
amouhoub, line 323: Are you sure that you want to send signals at the same time with 0 latency ?

fviale (approved):
LGTM, but I don't understand the use case behind.

mklkun (approved):

scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/signal/SignalApiTest.java
nebilbenmabrouk: yes :)

nebilbenmabrouk:
> LGTM, but I don't understand the use case behind.

I agree, but it is a request from Denis.

PR #3898, opened 2021-04-16
Optimize token node selection

fviale:
1) tasks with the same token are equivalent
2) make sure propagated variables are updated when tasks are compared


nebilbenmabrouk (approved):
LGTM

oanaBiancaSchiopu (approved):

PR #3899, opened 2021-04-20
Clarify Produces annotation in scheduler and rm rest api

fviale:
- several values where incorrect or incomplete which can cause issues with frontend code.
 - use constants when possible


nebilbenmabrouk (approved):
LGTM

fviale:
@nebilbenmabrouk already :-) ?

amouhoub (approved):
We should have done this looooog ago !
It's never too late :D

PR #3905, opened 2021-04-26
Add token one-time authentication mechanism for retrieving dataspace file.

tao-xinxiu:
- Add a endpoint /token to generate a token for one-time authentication.
- The TokenStore stores the map between token and sessionId, the token is removed once it's used.
- Add a query param token for authentication in the request of retrieving the dataspace file. The token is used when sessionId is not specified in the request header.


amouhoub (approved):
LGTM

fviale (approved):
LGTM

PR #3919, opened 2021-05-14
AOSynchronization : Fix groovy memory leak v3

fviale:
Third attempt at fixing memory leak.

Use GroovyShell.evaluate(Reader, String) as this implementation cleans the generated class from ClassInfo

Added a unit test which reproduces the issue.

### DIFF
--- a/scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/synchronization/AOSynchronizationTestLeak.java
+++ b/scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/synchronization/AOSynchronizationTestLeak.java
@@ -1,0 +1,84 @@
+/*
+ * ProActive Parallel Suite(TM):
+ * The Open Source library for parallel and distributed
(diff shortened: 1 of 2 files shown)


scheduler/scheduler-server/src/test/java/org/ow2/proactive/scheduler/synchronization/AOSynchronizationTestLeak.java
```diff
@@ -0,0 +1,84 @@
...
+    private AOSynchronization synchronization;
+
+    @Before
+    public void init() throws IOException {
+        tempFolder = folder.newFolder();
+        synchronization = new AOSynchronization(tempFolder.getAbsolutePath());
+    }
+
+    /**
+     * Run a groovy closure many time and check that there is no leak in org.codehaus.groovy.reflection.ClassInfo
```
mboussaa, line 64: many times

mboussaa (approved):

PR #3920, opened 2021-05-18
improve load finished jobs

fviale:
- do not load unnecessary info (task dependencies, task variables, etc)
 - fix workflows tests which were relying on analysing dependencies after the job is finished

Additionally:

 - fix database error when loading results of skipped task
 - improve JobRemoveHandler

### DIFF
--- a/scheduler/scheduler-server/src/test/java/functionaltests/workflow/TWorkflowJobs.java
+++ b/scheduler/scheduler-server/src/test/java/functionaltests/workflow/TWorkflowJobs.java
@@ -28,14 +28,17 @@ package functionaltests.workflow;
 import static functionaltests.utils.SchedulerTHelper.log;
 
 import java.io.File;
+import java.io.Serializable;
 import java.util.*;
 import java.util.Map.Entry;
+import java.util.concurrent.TimeUnit;
 
 import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.ow2.proactive.scheduler.common.job.*;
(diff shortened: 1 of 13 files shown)


nebilbenmabrouk (approved):
LGTM

scheduler/scheduler-server/src/test/java/functionaltests/workflow/TWorkflowJobs.java
```diff
@@ -155,6 +158,46 @@ public static JobId testJobSubmission(SchedulerTHelper schedulerHelper, Job jobT
 
         Assert.assertEquals(jInfo.getJobId(), id);
         Assert.assertEquals(JobStatus.RUNNING, jInfo.getStatus());
+        return id;
+    }
+
+    public static JobId waitForJobFinishedAndCompareDependencies(JobId id, SchedulerTHelper schedulerHelper,
```
mbenguig, line 164: To my pov.. lack of comment in this function (to explain each step, the first "for Task t"  loop and the second one..

mbenguig (approved):

scheduler/scheduler-server/src/test/java/functionaltests/workflow/TWorkflowJobs.java
fviale: Indeed, there is an original lack of comment which I did not improve. I will add some explanation

fviale: I tried to improve comments, as requested

PR #3921, opened 2021-05-20
Use cached bounded thread pool for selection manager and node sources

fviale:
This reduces the number of threads constantly active

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
@@ -318,22 +318,30 @@ public class SchedulerFrontend implements InitActive, Scheduler, RunActive, EndA
                                                                                       PASchedulerProperties.SCHEDULER_CLIENT_POOL_NBTHREAD.getValueAsInt(),
                                                                                       120L,
                                                                                       TimeUnit.SECONDS,
-                                                                                      new NamedThreadFactory("ClientRequestsThreadPool"));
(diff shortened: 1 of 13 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontend.java
```diff
@@ -318,22 +318,30 @@ public void initActivity(Body body) {
                                                                                       PASchedulerProperties.SCHEDULER_CLIENT_POOL_NBTHREAD.getValueAsInt(),
                                                                                       120L,
                                                                                       TimeUnit.SECONDS,
-                                                                                      new NamedThreadFactory("ClientRequestsThreadPool"));
+                                                                                      new NamedThreadFactory("ClientRequestsThreadPool",
+                                                                                                             false,
+                                                                                                             3));
```
nebilbenmabrouk, line 323: is it possible to make the number of threads configurable? Same for the other thread pools

fviale: it is configurable actually (all thread pools are).  The configuration pa.scheduler.core.listener.threadnumber is stored in a static variable NB_THREADS

fviale: 3 is not the pool size, it is the thread priority

mboussaa (approved):

nebilbenmabrouk (approved):

PR #3930, opened 2021-05-31
Disable scheduling loop node partitioning

fviale:
Even though in general usage the current implementation works, it became very inefficient when there are nodes protected by tokens. In that case, if many tasks (10K more) needs to be scheduled, the scheduling loop is querying the rm several thousand times for a small number of nodes.

This change disable the node partitioning. When there are 10K tasks to be scheduled and all are "compatible", the scheduling loop send a single request to the RM, which returns in a single call the available nodes.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
@@ -323,7 +323,8 @@ public final class SchedulingMethodImpl implements SchedulingMethod {
 
         while (progressiveIterator.hasMoreElements() && !freeResources.isEmpty()) {
 
-            LinkedList<EligibleTaskDescriptor> taskRetrievedFromPolicy = new LinkedList<>(progressiveIterator.getNextElements(freeResources.size()));
+            LinkedList<EligibleTaskDescriptor> taskRetrievedFromPolicy = new LinkedList<>(progressiveIterator.getNextElements(PASchedulerProperties.SCHEDULER_POLICY_USE_FREE_NODES.getValueAsBoolean() ? freeResources.size()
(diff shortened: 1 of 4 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -323,7 +323,8 @@ private int selectAndStartTasks(Policy currentPolicy, Map<JobId, JobDescriptor>
 
         while (progressiveIterator.hasMoreElements() && !freeResources.isEmpty()) {
 
-            LinkedList<EligibleTaskDescriptor> taskRetrievedFromPolicy = new LinkedList<>(progressiveIterator.getNextElements(freeResources.size()));
+            LinkedList<EligibleTaskDescriptor> taskRetrievedFromPolicy = new LinkedList<>(progressiveIterator.getNextElements(PASchedulerProperties.SCHEDULER_POLICY_USE_FREE_NODES.getValueAsBoolean() ? freeResources.size()
+                                                                                                                                                                                                        : Integer.MAX_VALUE));
```
mboussaa, line 327: @fviale the Integer.MAX_VALUE is scarying. Could it make an issue here in the sublist ? https://github.com/ow2-proactive/scheduling/blob/a47870e681640fa0cff88146fa8a47a5d4ca580b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/VariableBatchSizeIterator.java#L55

fviale: Ok I will have a look

fviale: I made a change in VariableBatchSizeIterator to ensure it does not cause issues

mboussaa (approved):

PR #3932, opened 2021-06-04
Fix variables propagation issue

fviale:
During the workflow parsing, a job variable became converted to a task variable

Added unit and functional tests


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/StaxJobFactory.java
```diff
@@ -867,7 +867,7 @@ private void handleCancelJobOnErrorAttribute(CommonAttribute commonPropertiesHol
         LinkedHashMap<String, TaskVariable> updatedVariablesMap = new LinkedHashMap<>(variablesMap);
```
mbenguig, line 867: (non mandatory) i could have been great to rename all non relevant map name, by using jobVariables, jobSubmissionVariables, resolvedTaskVariables,..

mbenguig (approved):

mboussaa (approved):

PR #3934, opened 2021-06-14
Topology use hostname instead of InetAddress

fviale:
InetAddress was used as key to several maps.

This does not work properly as InetAddress are equals if their associated ipv4 address are equals thus 127.0.1.1 from different hosts are considered equals.

Instead of using InetAddress, we now use hostnames.


nebilbenmabrouk:
@fviale a global comment. What if different machines (with different IPs) have the same hostname? (like our trydev and jenkins)
I suggest to compare nodes using both **hostname** and **InetAddress**, thus, we obtain reliable comparisons.
What do you think?

fviale:
@nebilbenmabrouk and what if multiples nodes are deployed on the same machine but bound to different ip addresses (network interface) ?
There is no perfect solution here. and I prefer to consider multiple machines with the same hostname an infrastructure and not a proactive issue.

nebilbenmabrouk:
> @nebilbenmabrouk and what if multiples nodes are deployed on the same machine but bound to different ip addresses (network interface) ?
> There is no perfect solution here. and I prefer to consider multiple machines with the same hostname an infrastructure and not a proactive issue.

Fair enough :)

nebilbenmabrouk (approved):
LGTM

mboussaa (approved):

PR #3937, opened 2021-06-21
Add security headers to jetty responses

fviale:
A few recommended security headers are added dynamically by jetty server

Added web properties to allow configuration of these headers


aminelouati (approved):

mboussaa (approved):
complex to understand these properties but looks ok

PR #3946, opened 2021-07-16
Add new Global variables mechanism

fviale:
- relax-ng schema for global variables configuration
 - configuration parser (refreshed periodically)
 - global variables and/or generic info can be applied to submitted workflows according to xpath filters
 - added unit tests
 - Handled the special case of schedulerapi (SchedulerNodeClient). When a task submits a new workflow, we expect this new workflow to preserve by default the global variables values (if overridden).

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/globalvariables/GlobalVariablesParser.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/globalvariables/GlobalVariablesParser.java
@@ -1,0 +1,191 @@
+/*
+ * ProActive Parallel Suite(TM):
+ * The Open Source library for parallel and distributed
+ * Workflows & Scheduling, Orchestration, Cloud Automation
+ * and Big Data Analysis on Enterprise Grids & Clouds.
+ *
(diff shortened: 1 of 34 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/globalvariables/GlobalVariablesParser.java
```diff
@@ -0,0 +1,187 @@
...
+        try {
+            builder = builderFactory.newDocumentBuilder();
+        } catch (ParserConfigurationException e) {
+            logger.error("Error when configuring DocumentBuilder", e);
+            return answer;
+        }
+
+        try (StringReader reader = new StringReader(jobContent)) {
+            InputSource inputSource = new InputSource(reader);
+            //inputSource.setEncoding(FILE_ENCODING);
```
AlexFalessi, line 129: Forgotten comment here ?

fviale: yes I need to remove it, well spotted

AlexFalessi (approved):

tao-xinxiu (approved):

PR #3950, opened 2021-07-22
Add sessionid to scripts

fviale:
- PerUserConnectionRMProxiesManager creates a rest session for each user
 - This works only if the rest server is launched in the same JVM as the scheduler server
 - sessionids are propagated to selection (via RMProxy and Criteria), task (TimedDoAction), pre, post, forkenv, flow (TaskLaunchedInitializer) and clean scripts (RMProxy)
 - The script must be "cloned" before execution, the original Script object, reference by InternalTask must not be modified. Otherwise, it would break the late binding reference mechanism, allowing to edit the script definition on task restarts.


aminelouati (approved):
LGTM

AlexFalessi (approved):

PR #3966, opened 2021-09-16
Fix exporting job of old schemas, only export advanced when it's true.

tao-xinxiu:
Fix the bug when exporting a job which is using old schemas, it should not contain the new variable parameter "advanced". 
Now we only export the advanced parameter when it's set to true, which is the case needed to be handled.


fviale (approved):
LGTM

marounkoussaifi (approved):

PR #3967, opened 2021-09-19
Add Hidden variables

fviale:
- hidden variables are not displayed on the submission form
 - hidden variables can be displayed or hidden using SPEL expressions (dynamic forms)
 - JobInfoData : attach full variable definition (not just key and values)

### DIFF
--- a/rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/data/JobInfoImpl.java
+++ b/rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/data/JobInfoImpl.java
@@ -31,6 +31,7 @@ import org.ow2.proactive.scheduler.common.job.JobId;
 import org.ow2.proactive.scheduler.common.job.JobInfo;
 import org.ow2.proactive.scheduler.common.job.JobPriority;
 import org.ow2.proactive.scheduler.common.job.JobStatus;
+import org.ow2.proactive.scheduler.common.job.JobVariable;
 
 
 public class JobInfoImpl implements JobInfo {
@@ -78,6 +79,8 @@ public class JobInfoImpl implements JobInfo {
     private Map<String, String> genericInformation;
 
     private Map<String, String> variables;
+
+    private Map<String, JobVariable> detailedVariables;
 
     private Set<String> signals;
 
@@ -292,6 +295,15 @@ public class JobInfoImpl implements JobInfo {
 
     public void setVariables(Map<String, String> variables) {
         this.variables = variables;
+    }
+
+    @Override
+    public Map<String, JobVariable> getDetailedVariables() {
+        return detailedVariables;
+    }
+
+    public void setDetailedVariables(Map<String, JobVariable> detailedVariables) {
+        this.detailedVariables = detailedVariables;
     }
 
     @Override
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/ModelValidatorContext.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/ModelValidatorContext.java
@@ -28,7 +28,10 @@ package org.ow2.proactive.scheduler.common.job.factories.spi.model;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Set;
 
 import org.ow2.proactive.scheduler.common.Scheduler;
 import org.ow2.proactive.scheduler.common.SchedulerSpaceInterface;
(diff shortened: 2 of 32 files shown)


rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/data/JobInfoImpl.java
```diff
@@ -294,6 +297,15 @@ public void setVariables(Map<String, String> variables) {
         this.variables = variables;
     }
 
+    @Override
+    public Map<String, JobVariable> getVariablesFull() {
+        return variablesFull;
+    }
+
```
mboussaa, line 304: add @Override ?

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/ModelValidatorContext.java
```diff
@@ -191,10 +244,22 @@ public void updateTaskWithContext(Task task) {
          */
         private Boolean valid;
 
-        public SpELVariables(Map<String, Serializable> variables, Map<String, String> models) {
+        public SpELVariables(Map<String, Serializable> variables, Map<String, String> models, Set<String> groupsNames) {
             this.variables = variables;
             this.models = models;
             this.tempMap = new HashMap<>();
```
mboussaa, line 250: u can apply LinkedHashMap for tempMap as well

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskDataVariable.java
```diff
@@ -153,10 +156,27 @@ public void setGroup(String group) {
...
+            return advanced;
+        }
     }
 
     public void setAdvanced(Boolean advanced) {
         this.advanced = advanced;
     }
+
+    @Column(name = "VARIABLE_HIDDEN")
+    public Boolean getHidden() {
```
mboussaa, line 171: I am a bit confused about boolean and Boolean, I don't know the difference and if it is harmful to use sometimes one and sometimes another.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/ModelValidatorContext.java
```diff
@@ -176,6 +219,16 @@ public void updateTaskWithContext(Task task) {
          */
         private Map<String, String> models;
 
+        /**
+         * Can be used to hide/unhide variables
```
mboussaa, line 223: better to use hide/show as you did for method names "hideVar" and "showVar"

mboussaa (approved):
LGTM

rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/data/JobInfoImpl.java
fviale: this setter is not defined in the interface

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/ModelValidatorContext.java
fviale: ok

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/TaskDataVariable.java
fviale: Boolean is an object, and thus can take a null value.
boolean is a primitive type and thus can only take true/false value.

In the database we want Boolean object as the column can have null values (workflows submitted with an earlier scheduler version).

oanaBiancaSchiopu (approved):

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/ModelValidatorContext.java
```diff
@@ -261,5 +334,41 @@ public Boolean getValid() {
         public void setValid(Boolean valid) {
             this.valid = valid;
         }
+
+        public boolean hideVar(String variableName) {
```
tao-xinxiu, line 338: I feel a bit confused about these two funcs. They are used to change the variable state stored in the map, and always return true if it's a hidden variable (i.e., no matter what's its current state)? Could you add a comment?

tao-xinxiu (approved):
LGTM with two small comments.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/ModelValidatorContext.java
fviale: it always returns true because it does not perform a validation. It performs an action. Returning true allows to use these functions easily in a spel expression. e.g. showVar('var1) && hideVar('var2')

PR #3973, opened 2021-10-07
Use XSLT to update job descriptor schema version to latest

tao-xinxiu:
Use XSLT to transform job descriptor schema version to latest to fix schema version mismatch pb between global var and job descriptor.

This transformation uses Saxon transformer, because the default oracle XSLT transformer (Xalan and Xerces) will assign auto-generated namespace prefix to all the elements whose namespace is changed (i.e., all the elements in our case, because we need to change the namespace for the root element job). To keep other places keeps using previous Xalan and Xerces transformer, the system property "javax.xml.transform.TransformerFactory" is set in SchedulerStart (for Job2XMLTransformer) and JettyStarter (for scheduling-portal and catalog etc).


tao-xinxiu:
retest it please

PR #3994, opened 2021-12-01
Workflow var PA:CATALOG_OBJECT add params to filter kind and contentType

tao-xinxiu:
Add parameters (optional) to the workflow variable model PA:CATALOG_OBJECT model to filter kind and contentType.


oanaBiancaSchiopu (approved):

fviale (approved):
LGTM

PR #3995, opened 2021-12-03
set-default-values-when-sending-signals-with-variables

oanaBiancaSchiopu:


fviale:
Retest this please

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/DefaultModelJobValidatorServiceProvider.java
```diff
@@ -44,6 +44,7 @@
 import org.ow2.proactive.scheduler.common.task.Task;
 import org.ow2.proactive.scheduler.common.task.TaskVariable;
 import org.ow2.proactive.scheduler.common.util.VariableSubstitutor;
+import org.springframework.util.CollectionUtils;
```
fviale, line 47: is it possible to use something else than springframework api here ?

if it only does:

`(variableList != null && !variableList.isempty())`

I would prefer not using it

fviale (approved):
small remark, otherwise LGTM

fviale:
Retest this please

fviale:
Retest this please

tao-xinxiu (approved):

PR #4004, opened 2021-12-23
external endpoints, add enpoint name and icons

fviale:
also, improve some default error codes in ExceptionMapper

### DIFF
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
@@ -2442,15 +2442,21 @@ public interface SchedulerRestInterface {
      *            current session
      * @param jobId
      *            id of the job
+     * @param endpointName
+     *          name of the endpoint
      * @param externalEndpointUrl
      *          the external endpoint url to add to the job endpoint list
+     * @param endpointIconUri
+     *          uri of the icon associated with the endpoint. If this parameter is null, the workflow icon will be used.
      */
     @POST
     @Path("jobs/{jobid}/externalendpointurl")
-    @Consumes(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
     @Produces(MediaType.APPLICATION_JSON)
     void addExternalEndpointUrl(@HeaderParam("sessionid") String sessionId, @PathParam("jobid")
-    final String jobId, String externalEndpointUrl) throws RestException;
+    final String jobId, @FormParam("endpointName") String endpointName,
+            @FormParam("externalEndpointUrl") String externalEndpointUrl,
+            @FormParam("endpointIconUri") String endpointIconUri) throws RestException;
 
     /**
      *
--- a/rest/rest-client/src/test/java/functionaltests/SchedulerClientTest.java
+++ b/rest/rest-client/src/test/java/functionaltests/SchedulerClientTest.java
@@ -70,6 +70,7 @@ import org.ow2.proactive.scheduler.common.task.TaskInfo;
 import org.ow2.proactive.scheduler.common.task.TaskResult;
 import org.ow2.proactive.scheduler.common.task.TaskState;
 import org.ow2.proactive.scheduler.common.task.TaskStatus;
+import org.ow2.proactive.scheduler.job.ExternalEndpoint;
 import org.ow2.proactive.scheduler.rest.ISchedulerClient;
 import org.ow2.proactive.scheduler.rest.SchedulerClient;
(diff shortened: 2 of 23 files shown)


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
```diff
@@ -2442,15 +2442,21 @@ JobValidationData validateJobSignal(@HeaderParam("sessionid") String sessionId,
...
     @POST
     @Path("jobs/{jobid}/externalendpointurl")
-    @Consumes(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
     @Produces(MediaType.APPLICATION_JSON)
     void addExternalEndpointUrl(@HeaderParam("sessionid") String sessionId, @PathParam("jobid")
-    final String jobId, String externalEndpointUrl) throws RestException;
+    final String jobId, @FormParam("endpointName") String endpointName,
+            @FormParam("externalEndpointUrl") String externalEndpointUrl,
+            @FormParam("endpointIconUri") String endpointIconUri) throws RestException;
```
nebilbenmabrouk, line 2459: For the endpoint icon, during our discussions we agreed to do the same as for PSA endpoints (use the workflow, icon).
Do we really need the parameter `endpointIconUri`

rest/rest-client/src/test/java/functionaltests/SchedulerClientTest.java
```diff
@@ -161,8 +162,14 @@ public void testAddAndRemoveExternalEndpointUrl() throws Throwable {
...
-        Assert.assertTrue(jobInfo.getExternalEndpointUrls()
-                                 .containsAll(Arrays.asList("http://ccc.fr", "http://eee.fr", "http://bbb.fr")));
+        Assert.assertEquals(new ExternalEndpoint("ccc", "http://ccc.fr", "icon/ccc"),
+                            jobInfo.getExternalEndpointUrls().get("ccc"));
+        Assert.assertEquals(new ExternalEndpoint("bbb", "http://bbb.fr", "icon/bbb"),
+                            jobInfo.getExternalEndpointUrls().get("bbb"));
+        // the last endpoint has null icon on purpose (to check that it does not cause issues)
+        Assert.assertEquals(new ExternalEndpoint("eee", "http://eee.fr", null),
+                            jobInfo.getExternalEndpointUrls().get("eee"));
+        //Arrays.asList("http://ccc.fr", "http://eee.fr", "http://bbb.fr")));
```
mboussaa, line 172: to remove

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
fviale: it can be null, in which case the workflow icon will be used

rest/rest-client/src/test/java/functionaltests/SchedulerClientTest.java
fviale: yes, sorry

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
mboussaa: and if we have many endpoints and one single wf? they will have the same icon of the main wf which might also not have an icon

nebilbenmabrouk: @fviale @mboussaa Fair enough

nebilbenmabrouk (approved):
LGTM

mboussaa:
@fviale @nebilbenmabrouk could the endpoint have an optional description as well? so in the front, we can show it as a tooltip?

nebilbenmabrouk:
> @fviale @nebilbenmabrouk could the endpoint have an optional description as well? so in the front, we can show it as a tooltip?

@mboussaa for the sake of homogeneity with PSA endpoints, we do not need the endpoint description (at least for the moment)

mboussaa:
@nebilbenmabrouk @fviale Last question: We can have an endpoint that does not have a HTTP/HTTPS endpoint like tcp address. What do you think about that? otherwise it LGTM

mboussaa (approved):

fviale:
@mboussaa it is the same for psa services without http endpoints. In that case they are not "clickable", and it's a frontend decision, not backend. I agree with nebil, as psa service endpoints do not have descriptions, there is no reason to do it here (maybe in the future).

PR #4012, opened 2022-01-13
Add children count to JobInfo

fviale:
- when a job with a parentId is submitted, we increase the children count of the parent job
 - when a job with a parentId is removed, we decrease the children count of the parent job
 - also added the parentId to JobInfo
 - added a filter to revisionAndJobInfo endpoint, which allows to return all children of the given job
 - added db tests and rest client tests

### DIFF
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
@@ -202,6 +202,8 @@ public interface SchedulerRestInterface {
      *             Include only jobs with a project name that starts with projectName (case in-sensitive)
      * @param userName
      *             Include only jobs with a user name that matches exactly with userName (case in-sensitive)
+     * @param parentId
+     *             Include only children jobs of the given parent job id. This parameter is ignored if childJobs is set to false
      * @param sortParams
      *            string containing sort directives. It must contain a comma separated list of sort parameters.
      *            A sort parameter can contain ID,STATE,OWNER,PRIORITY,NAME,SUBMIT_TIME,START_TIME,IN_ERROR_TIME,FINISH_TIME,TOTAL_TASKS,
@@ -224,7 +226,8 @@ public interface SchedulerRestInterface {
             @QueryParam("childJobs") @DefaultValue("true") boolean childJobs,
             @QueryParam("jobName") @DefaultValue("") String jobName,
             @QueryParam("projectName") @DefaultValue("") String projectName,
-            @QueryParam("userName") @DefaultValue("") String userName, @QueryParam("sortParams") String sortParams)
+            @QueryParam("userName") @DefaultValue("") String userName,
+            @QueryParam("parentId") @DefaultValue("-1") Long parentId, @QueryParam("sortParams") String sortParams)
             throws RestException;
 
     /**
--- a/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/dto/JobInfoData.java
+++ b/rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/dto/JobInfoData.java
@@ -93,6 +93,10 @@ public class JobInfoData implements java.io.Serializable {
 
     private List<String> preciousTasks;
 
+    private Long parentId = null;
+
+    private int childrenCount = 0;
+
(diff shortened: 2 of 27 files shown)


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
```diff
@@ -202,6 +202,8 @@
      *             Include only jobs with a project name that starts with projectName (case in-sensitive)
      * @param userName
      *             Include only jobs with a user name that matches exactly with userName (case in-sensitive)
+     * @param parentId
+     *             Include only children jobs of the given parent job id. This parameter is ignored if childJobs == false
```
nebilbenmabrouk, line 206: This parameter is ignored if childJobs **is set to** false

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/dto/JobInfoData.java
```diff
@@ -93,6 +93,10 @@
 
     private List<String> preciousTasks;
 
+    private Long parentId = null;
```
nebilbenmabrouk, line 96: The default value of the parentId is set to -1 in the SchedulerRestInterface.java (line 230), and here it is null.
It would be better to have a homogenize the default value of parentId.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/JobInfo.java
```diff
@@ -305,4 +305,16 @@
      * @return non-empty result map test
      */
     boolean isResultMapPresent();
+
+    /**
+     * Return the parent id of the job
+     * @return parent id of the job, or null if the job as no parent
```
nebilbenmabrouk, line 311: @return parent id of the job, or null if the job **has** no parent

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/LiveJobs.java
```diff
@@ -370,6 +370,29 @@ void jobSubmitted(InternalJob job) {
         ClientJobState clientJobState = new ClientJobState(job);
         jobs.put(job.getId(), new JobData(job));
         listener.jobSubmitted(clientJobState);
+        if (job.getParentId() != null) {
+            // If a job has a parent, it means the parent job children count has been increased.
```
nebilbenmabrouk, line 374: If a job has a parent, it means **that** the parent job's children count has been increased

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
fviale: ok

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/dto/JobInfoData.java
fviale: I would like to keep it null in the database, I will look at the SchedulerRestInterface default

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/JobInfo.java
fviale: ok

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/LiveJobs.java
fviale: ok

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/dto/JobInfoData.java
fviale: @nebilbenmabrouk in SchedulerRestInterface the default value of parentId is in revisionAndJobInfo, to filter only jobs which are children of the given parentId. It is not the same semantic as JobInfoData

oanaBiancaSchiopu (approved):

amouhoub (approved):

PR #4014, opened 2022-01-17
Disable some models validations when variable is hidden

fviale:
When the variable is hidden, validation is disabled for models PA:CATALOG_OBJECT, PA:USER_FILE, PA:GLOBAL_FILE, PA:USER_FOLDER, PA:GLOBAL_FOLDER and PA:CREDENTIAL.

Disabling is mandatory for these models, as they require the user to select a file, credential or catalog object, which is not possible when the variable is hidden. Other models take default values which are required to be valid whether the variable is visible or not.


tao-xinxiu (approved):
LGTM

PR #4018, opened 2022-01-21
Update sshinfrav2

alijawadfahs:


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -51,10 +51,12 @@
 import org.ow2.proactive.resourcemanager.core.properties.PAResourceManagerProperties;
 import org.ow2.proactive.resourcemanager.exception.RMException;
 import org.ow2.proactive.resourcemanager.nodesource.common.Configurable;
-import org.ow2.proactive.resourcemanager.utils.CommandLineBuilder;
+import org.ow2.proactive.resourcemanager.nodesource.utils.InitScriptGenerator;
```
fviale, line 54: instead of completely replacing CommandLineBuilder usage in the node source, why not add extra feature to it ?

something like:
`clb.useStartupScript(String scriptContent)`

then `clb.buildCommandLine()` will use the script to generate the command line instead of building the command directly.

Also: 
 * the node source should contain a nodeJarUrl parameter
 * some properties such as -Dproactive.communication.protocol=%protocol% should not be set automatically

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/utils/InitScriptGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            //load configuration manager with the NodeSource properties file
+            nsConfig = NSProperties.loadConfig();
+        } catch (ConfigurationException e) {
+            logger.error("Exception when loading NodeSource properties", e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Getter
```
mboussaa, line 69: (optional) add javadoc to this getter?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -199,24 +200,37 @@ public void startNodeImpl(final HostTracker hostTracker, final int nbNodes, fina
...
         String cmdLine;
         String obfuscatedCmdLine;
-        try {
-            cmdLine = clb.buildCommandLine(true);
-            obfuscatedCmdLine = clb.buildCommandLine(false);
-        } catch (IOException e) {
-            throw new RMException("Cannot build the " + RMNodeStarter.class.getSimpleName() + "'s command line.", e);
-        }
+
+        cmdLine = initScriptGenerator.fillInAgentScriptProperties(this.startupScript,
```
mboussaa, line 208: I believe you should add try catch to cmdLine and obfuscatedCmdLine

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -199,24 +200,37 @@ public void startNodeImpl(final HostTracker hostTracker, final int nbNodes, fina
...
+                                                                            nodeName,
+                                                                            nbNodes,
+                                                                            "[OBFUSCATED_CRED]");
 
         // one escape the command to make it runnable through ssh
         if (cmdLine.contains("\"")) {
             cmdLine = cmdLine.replaceAll("\"", "\\\\\"");
         }
         final String finalCmdLine = cmdLine;
-
```
mboussaa, line 219: clean code of intellij does this ?

mboussaa (approved):

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
alijawadfahs: I guess this is spotlessJava

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
alijawadfahs: the command line builder used to throw an exception however with the new class no exception handling is required

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/utils/InitScriptGenerator.java
alijawadfahs: noted

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -173,23 +180,17 @@ public void startNodeImpl(final HostTracker hostTracker, final int nbNodes, fina
...
-        CommandLineBuilder clb = super.getDefaultCommandLineBuilder(getTargetOSObj());
 
         final boolean deployNodesInDetachedMode = PAResourceManagerProperties.RM_NODES_RECOVERY.getValueAsBoolean() ||
                                                   PAResourceManagerProperties.RM_PRESERVE_NODES_ON_SHUTDOWN.getValueAsBoolean();
         if (deployNodesInDetachedMode) {
             // if we do not want to kill the nodes when the RM exits or
             // restarts, then we should launch the nodes in background and
             // ignore the RM termination signal
-            clb.setDetached();
+            this.javaPath = "nohup " + this.javaPath;
```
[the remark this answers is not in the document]
alijawadfahs: yes the current implementation it's hardcoded to only use OperartingSystem.Unix, so technically the targetOs parameter is not used

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
alijawadfahs: The main Idea here is that we are imitating the work done in the new node source in nodesource-addons, CommandLineBuilder mainly creates the command to start the node agent which was replaced with NodeSource.properties, a configurable string that will be filled by simple calls. 
CommandLineBuilder is specific for a single type of call (start a proactive node agent), and for example if we want to use node.jar we have to recreate the whole class. 
In contrast, InitScriptGenerator is more elegant, the startup script will be imported according to the mode of deployment you want and then the variable will simply be passed to replace the properties. So for the next task completely replacing the CommandLineBuilder is necessary. 
PLEASE NOTE:  the branch is not ready to be merged, I pushed it so Isabelle can take a look and give me feedback, some changes are still necessary.

alijawadfahs: "the node source should contain a nodeJarUrl parameter" Yes this is left for the next step.

fviale: ok for the branch not ready to be merged, but you need to go in the right direction directly.

I believe the right direction is to use CommandLineBuilder and add inside the necessary extra features.

In that case the extra features is to build a command from a properties file with pattern replacement.

rm/rm-node/src/main/resources/NodeSource.properties
```diff
@@ -0,0 +1,11 @@
...
+#***********************************************************************************
+#ns.script.linux.jar.startup.scripts = mkdir -p /tmp/node && cd /tmp/node\n if ! type -p jre/bin/java; then wget -nv -N https://s3.amazonaws.com/ci-materials/Latest_jre/jre-8u281-linux-x64.tar.gz; tar -xf jre-8u281-linux-x64.tar.gz; mv jre1.8.0_281/ jre; fi\nwget -nv %nodeJarUrl%\nnohup jre/bin/java -jar node.jar -Dproactive.communication.protocol=%protocol% -Dpython.path=%jythonPath% -Dproactive.pamr.router.address=%rmHostname% -D%instanceIdNodeProperty%=%instanceId% -r %rmUrl% -s %nodeSourceName% %nodeNamingOption% -v %credentials% -w %numberOfNodesPerInstance% %additionalProperties% &
+ns.script.linux.agent.startup.scripts = %detachedModePrefix% %javaPath% %javaOptions% -Dfile.encoding=%fileEncoding% -cp .:%schedulingPath%/addons:%schedulingPath%/addons/*:%schedulingPath%/dist/lib/* org.ow2.proactive.resourcemanager.utils.RMNodeStarter -v %credentials% -n %nodeName% -s %nodeSourceName% -r %rmUrl% -w %numberOfNodesPerInstance% &
+
+# node default properties
+
+ns.default.suffix.rm.to.nodejar.url = /rest/node.jar
+ns.default.suffix.connector.iaas.url = /connector-iaas
```
fviale, line 10: remove this unused property

rm/rm-node/src/main/resources/NodeSource.properties
```diff
@@ -0,0 +1,11 @@
...
+# *************************Node startup script Properties***************************
+#***********************************************************************************
+#ns.script.linux.jar.startup.scripts = mkdir -p /tmp/node && cd /tmp/node\n if ! type -p jre/bin/java; then wget -nv -N https://s3.amazonaws.com/ci-materials/Latest_jre/jre-8u281-linux-x64.tar.gz; tar -xf jre-8u281-linux-x64.tar.gz; mv jre1.8.0_281/ jre; fi\nwget -nv %nodeJarUrl%\nnohup jre/bin/java -jar node.jar -Dproactive.communication.protocol=%protocol% -Dpython.path=%jythonPath% -Dproactive.pamr.router.address=%rmHostname% -D%instanceIdNodeProperty%=%instanceId% -r %rmUrl% -s %nodeSourceName% %nodeNamingOption% -v %credentials% -w %numberOfNodesPerInstance% %additionalProperties% &
+ns.script.linux.agent.startup.scripts = %detachedModePrefix% %javaPath% %javaOptions% -Dfile.encoding=%fileEncoding% -cp .:%schedulingPath%/addons:%schedulingPath%/addons/*:%schedulingPath%/dist/lib/* org.ow2.proactive.resourcemanager.utils.RMNodeStarter -v %credentials% -n %nodeName% -s %nodeSourceName% -r %rmUrl% -w %numberOfNodesPerInstance% &
```
fviale, line 5: can you rename it to ns.node.command.line.linux ?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -190,6 +196,16 @@ public void startNodeImpl(final HostTracker hostTracker, final int nbNodes, fina
...
         clb.setNodeName(nodeName);
         clb.setNumberOfNodes(nbNodes);
+
+        // set the stratup script retrieved from NodeSource.properties
+        if (this.startupScript != null && !this.startupScript.isEmpty()) {
+            clb.setStartupScript(this.startupScript);
+        } else {
+            // in this case CommandlineBuilder will automatically fallback to creating the command as a list
+            // as a result, user's stratup script is disregarded and replaced by a default one
+            logger.error("Unable to correctly retrieve the startupScript for the node source " + nodeSource.getName());
```
fviale, line 206: why is there a logger error here ? is the startup script never null if we did not enter it ?

rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/NSProperties.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.commons.configuration2.builder.FileBasedConfigurationBuilder;
+import org.apache.commons.configuration2.builder.fluent.Parameters;
+import org.apache.commons.configuration2.builder.fluent.PropertiesBuilderParameters;
+import org.apache.commons.configuration2.convert.DisabledListDelimiterHandler;
+import org.apache.commons.configuration2.convert.ListDelimiterHandler;
+import org.apache.commons.configuration2.ex.ConfigurationException;
+import org.apache.log4j.Logger;
+
+
+public class NSProperties {
```
fviale, line 41: can you use a different name for this class ? For example NSPropertiesLoader

As there is already a NSProperties class in https://github.com/ow2-proactive/nodesource-addons/blob/master/infrastructures/infrastructure-common/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/util/NSProperties.java

Even if it's not the same package, there could be one day a conflict

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -109,6 +111,9 @@
     @Configurable(description = "Options for the java command launching the node on the remote hosts", sectionSelector = 3)
     protected String javaOptions;
 
+    @Configurable(textArea = true, description = "startup script to launch the ProActive nodes (optional)", sectionSelector = 2)
+    protected String startupScript = initScriptGenerator.getDefaultLinuxStartupScript();
```
fviale, line 115: please add this new parameter in the comments section of the configure method

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
fviale: I still don't understand why there is a logger.error, I would suggest rather:
logger.info("Node source " + nodeSource.getName() + " is not using a startup script");

rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/InitScriptGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+     * @param startupScript the Linux startup script added by the user in the GUI.
+     * @param javaPath the path to the bin java.
+     * @param schedulingPath the path to the rm home.
+     * @param rmUrl the Url of the RM (pamr://0/ by default).
+     * @param fileEncoding the file encoding (UTF-8 by default).
+     * @param javaOptions the options to be passed to the Java command.
+     * @param nodeName the name of the node.
+     * @param nodeSourceName the name of the node source.
+     * @param credentials the RM credentials or an obfuscated credentials String .
+     * @param detached if set to true, the node agent will ignore the RM termination signal.
```
tao-xinxiu, line 96: This comment is a bit confusing for me. Does `nohup` make agent ignore RM termination signal? Will it impact remove the node source? And what about nodes started through jar?

tao-xinxiu, line 96: This comment is a bit confusing for me. Does `nohup` make agent ignore RM termination signal? Will it impact remove the node source? And what about nodes started through jar?

rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/CommandLineBuilder.java
```diff
@@ -252,8 +264,14 @@ public String getSourceName() {
      * @throws java.io.IOException if you supplied a ProActive Configuration file that doesn't exist.
      */
     public String buildCommandLine(boolean displayCredentials) throws IOException {
-        List<String> command = this.buildCommandLineAsList(displayCredentials);
-        return Joiner.on(' ').join(command);
+        if (!deploymentMode.equals("NoStartup")) {
```
tao-xinxiu, line 267: To confirm, `NoStartup` means no extra startup script from the user? Could you please comment it?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -109,6 +111,12 @@
     @Configurable(description = "Options for the java command launching the node on the remote hosts", sectionSelector = 3)
     protected String javaOptions;
 
+    @Configurable(description = "Agent, Jar or NoStartup. Please refer to the manual", sectionSelector = 1)
+    protected String deploymentMode = "NoStartup";
+
+    @Configurable(textArea = true, description = "Startup script that will execute before ProActive node agents", sectionSelector = 3)
+    protected String startupScript = "";
```
tao-xinxiu, line 118: before ProActive node agents or jar? Maybe just "before ProActive node startup"?

tao-xinxiu (changes requested):
Generally looks good to me. Some comments need to be added.
And I'm a bit confused about the mode `NoStartup`, in this case, we don't deploy any nodes? What's the use case for this mode?

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -109,6 +111,12 @@
     @Configurable(description = "Options for the java command launching the node on the remote hosts", sectionSelector = 3)
     protected String javaOptions;
 
+    @Configurable(description = "Agent, Jar or NoStartup. Please refer to the manual", sectionSelector = 1)
```
fviale, line 114: It is difficult to explain and understand here what Agent, Jar and NoStartup means.

It would be easier if we had two extra parameters startupScriptStandard and startupScriptWithNodeJarDownload

In that case deploymentMode could be: useStartupScript, useNodeJarStartupScript and Autogenerated. Which seems easier to explain

rm/rm-node/src/main/java/org/ow2/proactive/resourcemanager/utils/InitScriptGenerator.java
```diff
@@ -35,55 +35,84 @@
...
-    public static final String NODE_SOURCE_NAME_PROPERTY = "%nodeSourceName%";
+    private static final String NODE_JAR_URL_HOST_NAME_PROPERTY = "%hostName%";
 
-    public static final String CREDENTIALS_PROPERTY = "%credentials%";
+    private static final String NODE_JAR_URL_PORT_PROPERTY = "%port%";
 
-    public static final String NUMBER_OF_NODES_PER_INSTANCE_PROPERTY = "%numberOfNodesPerInstance%";
+    private static final String NODE_JAR_URL_DIR_PROPERTY = "%nodeJarDir%";
+
+    private static final String NODE_JAR_DIR = "rest/node.jar";
```
tao-xinxiu, line 77: NODE_JAR_DIR -> NODE_JAR_URL_PATH ? (DIR seems more to be the disk location on machine for me)

tao-xinxiu (approved):
LGTM, one last small comment.

rm/rm-node/src/main/resources/NodeCommandLine.properties
```diff
@@ -0,0 +1,7 @@
+#***********************************************************************************
+# *************************Node startup script Properties***************************
+#***********************************************************************************
+ns.node.jar.command.line.linux = mkdir -p /tmp/node && cd /tmp/node\n if ! type -p jre/bin/java; then wget -nv -N %javaUrl%; tar -xf jre-8u281-linux-x64.tar.gz; mv jre1.8.0_281/ jre; fi\nwget -nv -O node.jar %nodeJarUrl%\n%detachedModePrefix% jre/bin/java -jar node.jar %javaOptions% -Dpython.path=%jythonPath% -r %rmUrl% -n %nodeName% -s %nodeSourceName% -w %numberOfNodesPerInstance% -v %credentials% &
```
fviale, line 4: I don't think it's relevant to have a javaUrl pattern while subsequent commands depend on the java package name and folder. You should remove the pattern and put the url explicitely in the script

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -109,6 +109,18 @@
     @Configurable(description = "Options for the java command launching the node on the remote hosts", sectionSelector = 3)
     protected String javaOptions;
 
+    @Configurable(description = "autoGenerated, useStartupScript or useNodeJarStartupScript. Please refer to the manual", sectionSelector = 1)
```
fviale, line 112: we should decide which parameters are advanced, I would suggest all of the new parameters should be.

fviale: Please refer to the manual => Please refer to the infrastructure documentation

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/nodesource/infrastructure/SSHInfrastructureV2.java
```diff
@@ -109,6 +109,18 @@
...
+    protected String deploymentMode = "autoGenerated";
+
+    @Configurable(description = "The URL from which the node.jar will be downloaded. Used only when useNodeJarStartupScript mode is selected", sectionSelector = 3)
+    protected String nodeJarUrl = InitScriptGenerator.createNodeJarUrl();
+
+    @Configurable(textArea = true, description = "Standard Startup script that will start the ProActive node agents. Used only when useStartupScript mode is selected", sectionSelector = 3)
+    protected String startupScriptStandard = InitScriptGenerator.getDefaultLinuxStandardCommand();
+
+    @Configurable(textArea = true, description = "Startup script that will download a node.jar and start ProActive Agent using it. Used only when useNodeJarStartupScript mode is selected", sectionSelector = 3)
```
fviale, line 121: Startup script that will download a Jave Runtime Environment, a jar archive containing all ProActive libraries (node.jar) and start ProActive Agent using it. Used only when useNodeJarStartupScript mode is selected.

rm/rm-node/src/main/resources/NodeCommandLine.properties
fviale: additionnally why is not 8u312b07 used, like in https://github.com/ow2-proactive/nodesource-addons/blob/master/infrastructures/infrastructure-common/src/main/resources/NodeSource.properties ?

fviale:
Retest this please

fviale:
Retest this please

tao-xinxiu:
Retest this please

PR #4020, opened 2022-01-28
Thread leak issue in ExtendedSchedulerPolicy

fviale:
- make ScheduledThreadPoolExecutor static to avoid multiple instanciation
 - improve SchedulingThread by using a CountDownLatch, this allows other threads to notify SchedulingThread without acquiring a lock

Also, in install_base.sh, delete previous version jars in subfolders of the addons directory

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingThread.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingThread.java
@@ -27,6 +27,8 @@ package org.ow2.proactive.scheduler.core;
 
 import java.util.Timer;
 import java.util.TimerTask;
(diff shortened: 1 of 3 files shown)


nebilbenmabrouk (approved):
LGTM, and thanks for fixing also the problem related to the upgrade with the same version

AlexFalessi (approved):

fviale:
@nebilbenmabrouk actually the fix I did does not concern the upgrade with the same version, but a regular upgrade. I noticed the issue when we were looking at the subfolders of addons, remember ?

nebilbenmabrouk:
> @nebilbenmabrouk actually the fix I did does not concern the upgrade with the same version, but a regular upgrade. I noticed the issue when we were looking at the subfolders of addons, remember ? There were jars coming from old versions

You're right, I saw that also

aminelouati (approved):

PR #4022, opened 2022-02-03
replace-system-variables

oanaBiancaSchiopu:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -2039,6 +2044,34 @@ public void newJobSubmitted(final InternalJob job) {
         });
     }
 
+    private void replaceSystemVariables(Serializable job) {
```
fviale, line 2047: I understand the reason why you give a Serializable job parameter, but the code would be more elegant by creating two private methods, one which takes InternalJob and the other one which takes JobData

oanaBiancaSchiopu: I was thinking about this too but there would be 2 methods with the exact same code, the only difference would be the method parameter InternalJob/JobData

oanaBiancaSchiopu: if you think that it would look better, I can do the change

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -2039,6 +2044,34 @@ public void newJobSubmitted(final InternalJob job) {
...
+        variableReplacement.put(SchedulerVars.PA_JOB_ID.name(), internalJob.getId());
+        variableReplacement.put(SchedulerVars.PA_JOB_NAME.name(), internalJob.getName());
+        variableReplacement.put(SchedulerVars.PA_USER.name(), internalJob.getOwner());
+
+        internalJob.getVariables().values().forEach(jobVariable ->
+                variableReplacement.put(jobVariable.getName(), jobVariable.getValue()));
+        internalJob.getVariables().values().forEach(runtimeVariable -> {
+            String originalValue = runtimeVariable.getValue();
+            runtimeVariable.setValue(VariableSubstitutor.filterAndUpdate(originalValue, variableReplacement));
+        });
```
AlexFalessi, line 2072: I think you can do both processes in a single `forEach()` to avoid streaming twice

AlexFalessi (approved):
Small non blocking comment

PR #4025, opened 2022-02-11
Create endpoint for submit with body parameters variables

oanaBiancaSchiopu:


fviale (approved):
LGTM
is there no conflicts with the other jobs POST method that produces and consumes application/json ?

oanaBiancaSchiopu:
> LGTM is there no conflicts with the other jobs POST method that produces and consumes application/json ?

no, because they have different parameters

    @POST
    @Path("{path:jobs}")
    @Produces(MediaType.APPLICATION_JSON)
    JobIdData submitFromUrl(@HeaderParam("sessionid") String sessionId, @HeaderParam("link") String url,
            @PathParam("path") PathSegment pathSegment, @Context UriInfo contextInfos) throws JobCreationRestException,
            NotConnectedRestException, PermissionRestException, SubmissionClosedRestException, IOException;

rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
```diff
@@ -1539,6 +1539,31 @@ JobIdData submitFromUrl(@HeaderParam("sessionid") String sessionId, @HeaderParam
             throws JobCreationRestException, NotConnectedRestException, PermissionRestException,
             SubmissionClosedRestException, IOException;
 
+    /**
+     * Submits a workflow to the scheduler from a workflow URL, creating hence a
```
mboussaa, line 1543: u have copied/pasted the endpoint description from the previous one, we should explain the difference between the original endpoint and the new one in the description

oanaBiancaSchiopu: the other 2 endpoints have the same description. In fact, they do the same thing, just the parameters are different. I have changed the parameters description:
     * @param jsonBody
     *            a json with the variables of the workflow

mboussaa: ok.thx

mboussaa (approved):

PR #4030, opened 2022-03-15
Add cache mechanism to workflow and script download

fviale:
Additionally, improve selection scripts sessionid usage.


rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -2692,16 +2706,27 @@ public Response index() throws URISyntaxException {
...
             HttpResourceDownloader httpResourceDownloader = HttpResourceDownloader.getInstance();
-            return httpResourceDownloader.getResource(sessionId, workflowUrl, String.class);
+            downloadedWorkflow = httpResourceDownloader.getResource(sessionId, workflowUrl, String.class);
         } else {
             URL nonHttpURL = new URL(workflowUrl);
-            return IOUtils.toString(nonHttpURL.openStream(), Charsets.UTF_8);
+            downloadedWorkflow = IOUtils.toString(nonHttpURL.openStream(), Charsets.UTF_8);
         }
+        workflowCache.put(key, downloadedWorkflow);
+        return downloadedWorkflow;
```
nebilbenmabrouk, line 2729: Should we check that 'downloadedWorkflow' is not null before returning it ? otherwise throw an exception

nebilbenmabrouk (approved):
LGTM, minor comment

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
fviale: httpResourceDownloader.getResource() and downloadedWorkflow = IOUtils.toString() don't seem to be able to return null, they would throw exceptions. But I can add a null check and throw a meaningful exception just in case

mklkun (approved):
LGTM

PR #4033, opened 2022-04-01
Add housekeeping configuration of jobs containing errors

fviale:
Also, improved adjusting the housekeeping after job recovery (when enabled)

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/LiveJobs.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/LiveJobs.java
@@ -812,6 +812,16 @@ class LiveJobs {
         }
     }
 
+    public static boolean isJobWithErrors(InternalJob job) {
+        switch (job.getStatus()) {
+            case FAILED:
+            case CANCELED:
+                return true;
+            default:
+                return job.getJobInfo().getNumberOfFailedTasks() > 0 || job.getJobInfo().getNumberOfFaultyTasks() > 0;
+        }
+    }
+
     TerminationData finishInErrorTask(JobId jobId, String taskName) throws UnknownTaskException, UnknownJobException {
         JobData jobData = lockJob(jobId);
         if (jobData == null) {
@@ -851,7 +861,10 @@ class LiveJobs {
                 job.terminate();
                 jlogger.debug(job.getId(), "terminated");
                 jobs.remove(job.getId());
-                terminationData.addJobToTerminate(job.getId(), job.getGenericInformation(), job.getCredentials());
+                terminationData.addJobToTerminate(job.getId(),
+                                                  job.getGenericInformation(),
+                                                  job.getCredentials(),
+                                                  isJobWithErrors(job));
             }
 
             task.setInErrorTime(-1);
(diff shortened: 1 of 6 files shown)


mboussaa (approved):
LGTM

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/LiveJobs.java
```diff
@@ -812,6 +812,21 @@ private void setJobStatusAfterInErrorTaskRestart(InternalJob job) {
         }
     }
 
+    public static boolean isJobWithErrors(InternalJob job) {
+        switch (job.getStatus()) {
+            case FAILED:
+            case CANCELED:
+                return true;
+            default:
+                if (job.getJobInfo().getNumberOfFailedTasks() > 0 || job.getJobInfo().getNumberOfFaultyTasks() > 0) {
```
tao-xinxiu, line 821: The whole if-else part could be simplified to `return job.getJobInfo().getNumberOfFailedTasks() > 0 || job.getJobInfo().getNumberOfFaultyTasks() > 0`

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -1128,6 +1129,29 @@ void terminateJobHandling(final JobId jobId, final Map<String, String> jobGeneri
         }
     }
 
+    private long getRemoveDelayFromGenericInfo(JobId jobId, Map<String, String> jobGenericInfo, boolean isJobWithErrors,
+            long removeDelay) {
```
tao-xinxiu, line 1133: This function seems not need the variable `removeDelay` as input?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
```diff
@@ -1128,6 +1129,29 @@ void terminateJobHandling(final JobId jobId, final Map<String, String> jobGeneri
...
+        if (jobGenericInfo != null) {
+            if (jobGenericInfo.containsKey(GENERIC_INFO_REMOVE_DELAY)) {
+                try {
+                    removeDelay = Tools.parsePeriod(jobGenericInfo.get(GENERIC_INFO_REMOVE_DELAY));
+                } catch (Exception e) {
+                    logger.error("Error when parsing generic information " + GENERIC_INFO_REMOVE_DELAY + " for job " +
+                                 jobId, e);
+                }
+            }
+            if (isJobWithErrors && jobGenericInfo.containsKey(GENERIC_INFO_REMOVE_DELAY_ON_ERROR)) {
```
tao-xinxiu, line 1143: Could the two cases both happen? `else if` might be a bit more clear?

tao-xinxiu (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/LiveJobs.java
fviale: indeed

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingService.java
fviale: yes both case can happen. A user can specific both generic info

PR #4038, opened 2022-05-06
Add tenant filtering

fviale:
- tenants can be defined through a configuration file or a ldap attribute
 - a scheduler property allows to enable/disable job filtering by tenant.
 - allow filtering by tenant in revisionAndJobsInfo
 - allow node source restriction by tenants


rm/rm-api/src/main/java/org/ow2/proactive/resourcemanager/core/properties/PAResourceManagerProperties.java
```diff
@@ -138,6 +138,9 @@ public enum PAResourceManagerProperties implements PACommonProperties {
     /** Resource Manager group file name */
     RM_GROUP_FILE("pa.rm.defaultgroupfilename", PropertyType.STRING, "config/authentication/group.cfg"),
 
+    /** Resource Manager tenant file name */
+    RM_TENANT_FILE("pa.rm.defaulttenantfilename", PropertyType.STRING, "config/authentication/tenant.cfg"),
```
mboussaa, line 142: I don't understand why having config/authentication/tenant.cfg for RM and Scheduler separately?

mboussaa (approved):
1/4 LGTM

aminelouati (approved):

PR #4045, opened 2022-06-16
Improve getJobs performance

fviale:
Add a statusRank column to JobData used to sort statuses (pending first, then all running states, then all finished states)

Bootstrap method to fill the statusRank column on startup for an existing older version database

Added indexes that benefit from both statusRank and jobId reverse sort
Added additional indexes for common use cases (myJobs, parentJobsOnly)

Added a custom hibernate interceptor for HSQLDB which uses indexes on ORDER BY columns only if the hint USING INDEX is appended to the query

Added the hsqldb option sql.nulls_order=false to let null values be sorted according to ASCENDING or DESCENDING order

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/DBJobDataParameters.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/DBJobDataParameters.java
@@ -32,7 +32,7 @@ import java.util.Set;
 
 import org.ow2.proactive.db.SortParameter;
 import org.ow2.proactive.scheduler.common.JobSortParameter;
-import org.ow2.proactive.scheduler.common.job.JobStatus;
+import org.ow2.proactive.scheduler.common.job.JobState;
 
 
 /**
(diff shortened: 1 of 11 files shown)


tao-xinxiu (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/DBJobDataParameters.java
```diff
@@ -85,17 +85,17 @@ public class DBJobDataParameters {
...
+            newStatusRanks.add(JobState.PENDING_RANK);
         }
         if (running) {
-            newStatus.addAll(SchedulerDBManager.RUNNING_JOB_STATUSES);
+            newStatusRanks.add(JobState.RUNNING_RANK);
         }
         if (finished) {
-            newStatus.addAll(SchedulerDBManager.FINISHED_JOB_STATUSES);
+            newStatusRanks.add(JobState.FINISHED_RANK);
         }
```
nebilbenmabrouk, line 97: Is it possible to use `else if ` to avoid executing additional if statements?

nebilbenmabrouk (approved):
LGTM

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/DBJobDataParameters.java
fviale: no because it is not exclusive conditions. you can have all possible combinations of pending, running and finished

PR #4049, opened 2022-06-24
Change ssh2 libraries

nebilbenmabrouk:
- Add bcutil-jdk15to18, bctls-jdk15to18 and bcpg-jdk15to18 to scheduler-examples
- Exclude the old transitive trilead-ssh2 patch of org.tmatesoft.svnkit


rest/rest-cli/build.gradle
```diff
@@ -9,7 +9,9 @@ dependencies {
     compile project(":common:common-api")
     compile project(":common:common-http")
     compile project(':rest:rest-api')
-    compile group: 'org.tmatesoft.svnkit', name: 'svnkit', version: '1.9.0'
+    compile(group: 'org.tmatesoft.svnkit', name: 'svnkit', version: '1.9.0'){
```
fviale, line 12: how is svnkit used in this project ?

If you exclude this dependency, you need to manually add the replacement in this project as it does not depends on programming-extensions-pnps

alijawadfahs (approved):

PR #4053, opened 2022-07-03
SchedulingMethodImpl: parallelize createExecution

fviale:
- use the existing thread pool to run createExecution(s) in parallel
- start the database application and deploy the task launcher outside of a job lock (to deploy launchers in parallel for multiple tasks of the same job)

Also, in RMCore, allow setting multiple nodes as busy and use it in SelectionManager

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/threading/TimeoutThreadPoolExecutor.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/threading/TimeoutThreadPoolExecutor.java
@@ -26,14 +26,7 @@
 package org.ow2.proactive.threading;
 
 import java.util.List;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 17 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/threading/TimeoutThreadPoolExecutor.java
```diff
@@ -45,13 +38,31 @@ public TimeoutThreadPoolExecutor(int corePoolSize, int maximumPoolSize, long kee
...
+
+        BlockingQueue<Runnable> queue = new LinkedTransferQueue<Runnable>() {
+            @Override
+            public boolean offer(Runnable e) {
+                return tryTransfer(e);
+            }
+        };
+        TimeoutThreadPoolExecutor threadPool = new TimeoutThreadPoolExecutor(0,
+                                                                             nThreads,
+                                                                             120L,
```
tao-xinxiu, line 51: Does this parameter keepAliveTime (`120L`) need to be configurable? Or put it as a public static constant if it might be changed in the future?

tao-xinxiu (approved):
LGTM. One tiny comment.

scheduler/scheduler-server/src/main/java/org/ow2/proactive/threading/TimeoutThreadPoolExecutor.java
fviale: @tao-xinxiu 

I have used the same configuration in various other thread pools, without making the keepAliveTime configurable.

It could be configurable yes, but I would need to change it in various places.

nebilbenmabrouk (approved):
LGTM

PR #4054, opened 2022-07-05
Fix node recovery

fviale:
- add owner parameter to updateMultipleRMNodeStatus transaction
 - depending on usage info content (no usage info, single usage info or multiple usage info), add usage info the same transaction or add it in delayed transactions

### DIFF
--- a/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
+++ b/rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
@@ -350,22 +350,22 @@ public abstract class SelectionManager {
 
         // the nodes are selected, now mark them as busy.
         int counter = 0;
+        int usageInfoCriteriaSize = criteria.getListUsageInfo() != null ? criteria.getListUsageInfo().size() : 0;
         int listCapacity = selectedNodes.size() +
                            (selectedNodes.getExtraNodes() != null ? selectedNodes.getExtraNodes().size() : 0);
         List<String> selectedNodesUrls = new ArrayList<>(listCapacity);
         List<Map<String, String>> usageInfoList = new ArrayList<>(listCapacity);
         for (Node node : selectedNodes) {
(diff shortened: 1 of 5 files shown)


rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
```diff
@@ -350,13 +350,14 @@ private NodeSet doSelectNodes(Criteria criteria, Client client) {
...
+        int usageInfoCriteriaSize = criteria.getListUsageInfo() != null ? criteria.getListUsageInfo().size() : 0;
         int listCapacity = selectedNodes.size() +
                            (selectedNodes.getExtraNodes() != null ? selectedNodes.getExtraNodes().size() : 0);
         List<String> selectedNodesUrls = new ArrayList<>(listCapacity);
         List<Map<String, String>> usageInfoList = new ArrayList<>(listCapacity);
         for (Node node : selectedNodes) {
-            if (criteria.getListUsageInfo() != null) {
-                if (criteria.getListUsageInfo().size() == 1) {
+            if (usageInfoCriteriaSize != 0) {
+                if (usageInfoCriteriaSize == 1) {
```
tao-xinxiu, line 360: The structure could be a bit more clear with less nested if:
```
if (usageInfoCriteriaSize == 1) { ...
} else if (usageInfoCriteriaSize != 0) { ...
} else { ...
}
```

tao-xinxiu (approved):

rm/rm-server/src/main/java/org/ow2/proactive/resourcemanager/selection/SelectionManager.java
fviale: thanks, I will improve it

fviale: else if (usageInfoCriteriaSize > 0) :-)

PR #4055, opened 2022-07-07
Fix pause issues in SchedulingMethodImpl

fviale:
As pauseJob or pauseTask requests can be received asynchronously, we need to check for the task status each time the job is locked.

Added a test which replicates the issue


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
```diff
@@ -917,91 +854,101 @@ protected boolean createExecution(NodeSet extraNodes, Node node, InternalJob job
...
-                task.getExecutableContainer().setNodes(extraNodes);
+                    //set nodes in the executable container
+                    task.getExecutableContainer().setNodes(extraNodes);
+
+                    tlogger.debug(task.getId(), "deploying");
 
-                tlogger.debug(task.getId(), "deploying");
+                    // Dynamically adjust the start-task-timeout according to the number dependency tasks in a merge.
+                    // above 500 parent tasks, it is worth adjusting.
+                    if (taskDescriptor.getParents().size() > 500) {
```
nebilbenmabrouk, line 880: 500 can be a configurable parameter?

nebilbenmabrouk (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulingMethodImpl.java
fviale: this is unrelated to the PR

PR #4071, opened 2022-09-08
Add tags attribute at job level

mbenguig:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/JobFilterCriteria.java
```diff
@@ -26,6 +26,7 @@
 package org.ow2.proactive.scheduler.common;
 
 import java.io.Serializable;
+import java.util.Set;
```
fviale, line 29: remove this useless import

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/JobState.java
```diff
@@ -288,7 +288,7 @@ private TaskStatesPage getTaskStatesPage(int offset, int limit, List<TaskState>
      * To get the list of available tags in a job.
      * @return the list of tags.
      */
-    public List<String> getTags() {
+    public List<String> getTaskTags() {
```
fviale, line 291: instead of renaming this existing method, which returns indeed legacy tags assigned to tasks, I would suggest that you call your new method and associated attribute getWorkflowTags

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/Job.java
```diff
@@ -86,6 +88,11 @@ public abstract class Job extends CommonAttribute {
      */
     protected String projectName = "";
 
+    /**
+     * Tags of this job
+     */
+    protected Set<String> tags = new HashSet<>();
```
fviale, line 94: you should not put this new attribute in CommonProperties as it would be assigned to both Job and Task.

Instead create a workflowTags attribute in the Job class, with associated getter and setter

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/job/InternalJob.java
```diff
@@ -1080,6 +1080,11 @@ public void setProjectName(String projectName) {
         jobInfo.setProjectName(projectName);
     }
 
+    @Override
```
fviale: this method should not be added

rest/rest-cli/src/test/java/org/ow2/proactive_grid_cloud_portal/cli/cmd/sched/ListJobCommandTest.java
```diff
@@ -125,6 +125,7 @@ public void testAllJobsReturned() throws Exception {
                                     null,
                                     null,
                                     null,
+                                    null,
```
fviale, line 128: there should not be an extra parameter to revisionAndJobsInfo

scheduler/scheduler-client/src/main/java/org/ow2/proactive/scheduler/job/ClientJobState.java
```diff
@@ -94,6 +94,7 @@ public ClientJobState(JobState jobState) {
         this.name = jobState.getName();
         this.description = jobState.getDescription();
         this.projectName = jobState.getProjectName();
+        this.workflowTags = jobState.getWorkflowTags();
```
fviale, line 97: I don't think this is needed

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SchedulerFrontendState.java
```diff
@@ -543,6 +543,7 @@ InternalJob createJob(Job userJob, UserIdentificationImpl ident)
         job.setTenant(ident.getTenant());
         // route project name inside job info
         job.setProjectName(job.getProjectName());
+        job.setWorkflowTags(job.getWorkflowTags());
```
fviale, line 546: I don't think this is needed

fviale (approved):
Two small comments, otherwise LGTM

fviale:
To see if we can define a regexp to control the tags XML attribute

tao-xinxiu (approved):

PR #4088, opened 2022-10-18
Add multi-ldap login module

fviale:
The ldap module used is associated with a user domain.
Different domains will use different ldap configuration

Also, ensure that domain information is properly read and propagated through the server and nodes.

### DIFF
--- a/common/common-api/src/main/java/org/ow2/proactive/authentication/ConnectionInfo.java
+++ b/common/common-api/src/main/java/org/ow2/proactive/authentication/ConnectionInfo.java
@@ -42,6 +42,8 @@ public class ConnectionInfo implements Serializable {
 
     private String login;
 
+    private String domain;
+
     private String password;
 
     private File credentialFile;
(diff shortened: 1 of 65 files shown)


aminelouati (approved):

mboussaa (approved):

common/common-api/src/main/java/org/ow2/proactive/authentication/ConnectionInfo.java
```diff
@@ -51,13 +53,32 @@ public class ConnectionInfo implements Serializable {
     /**
      * @param url            the REST server URL
```
nebilbenmabrouk, line 54: the REST server URL => LDAP server URL

common/common-api/src/main/java/org/ow2/proactive/authentication/ConnectionInfo.java
```diff
@@ -102,4 +131,40 @@ public void setCredentialFile(File credentialFile) {
...
+     * @param fullLogin the login to parse
+     * @return the domain name, null if no domain is specified
+     * @since Scheduling 3.0.1
+     */
+    public static String parseDomain(String fullLogin) {
+        if (fullLogin.contains("\\")) {
+            String domain = fullLogin.substring(0, fullLogin.indexOf("\\"));
+            if ("".equals(domain.trim())) {
+                return null;
+            }
```
nebilbenmabrouk, line 148: This block is useless. It is included in the instruction:
`return "".equals(domain.trim()) ? null : domain;`

config/rm/settings.ini
```diff
@@ -194,6 +194,11 @@ pa.rm.auth.pubkey.path=config/authentication/keys/pub.key
 # else, (if the path is absolute) it is directly interpreted
 pa.rm.ldap.config.path=config/authentication/ldap.cfg
 
+# Multi-domain LDAP configuration
+# a comma-separated list of domain_name:configuration_path
```
nebilbenmabrouk, line 198: a comma-separated list of  **the pair** _domain_name:configuration_path_

common/common-api/src/main/java/org/ow2/proactive/authentication/ConnectionInfo.java
fviale: no, here it is the rest server url

common/common-api/src/main/java/org/ow2/proactive/authentication/ConnectionInfo.java
fviale: you're right, this code is a blind copy/paste

PR #4108, opened 2022-11-21
Better handling of oracle varbinary columns

fviale:
This is to avoid java.sql.SQLException: Stream has already been closed errors when using oracle db


mboussaa (approved):
We should make sure it is ok on PG and MySQL too

PR #4115, opened 2022-12-13
SchedulerFrontend.insertSignals catch all exceptions

fviale:
If any exception occur in the keySet method, signal will be discarded


fviale:
Retest this please

PR #4123, opened 2022-12-23
Add bucket and object names in CATALOG_OBJECT model validator

mklkun:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/validator/CatalogObjectValidator.java
```diff
@@ -175,6 +181,26 @@ private boolean matchKindAndContentType(CommonHttpResourceDownloader.ResponseCon
                                                             expectedContentType));
             }
         }
+        if (StringUtils.isNotEmpty(expectedBucketName)) {
+            String catalogObjBucketName = jsonNode.path("bucket_name").asText();
+            if (!catalogObjBucketName.contains(expectedBucketName) ||
+                (expectedBucketName.startsWith("%") && !catalogObjBucketName.startsWith(expectedBucketName)) ||
```
fviale, line 187: if expectedBucketName starts with % the catalog object name can be any word ENDING with expectedBucketName minus the starting % character.

The contains will return false if catalogObjBucketName = pre_toto_post and expectedBucketName = %toto% 

Also, your use of OR expression means that you will throw a ValidationExpression at the first predicate that returns true.

Morality, you should first analyse the expected bucket name to see what kind of expression it is and then perform only 1 type of test.

Also consider introducing private methods as the test for bucket name and object name is similar

scheduler/scheduler-api/src/test/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/factory/CatalogObjectParserValidatorTest.java
```diff
@@ -70,11 +70,20 @@ public void testCatalogObjectParserValidatorOKWithKindAndContentTypeInModel()
         Assert.assertEquals(validValue, new CatalogObjectParserValidator(model).parseAndValidate(validValue));
     }
 
+    @Test
+    public void testCatalogObjectParserValidatorOKWithBucketAndName()
```
fviale, line 74: you should add more unit tests

fviale (approved):

aminelouati (approved):

PR #4126, opened 2023-01-11
Top workflows consuming the most cpu

oanaBiancaSchiopu:


mbenguig (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/job/InternalJob.java
```diff
@@ -1074,6 +1074,14 @@ public void setSubmittedTime(long submittedTime) {
         jobInfo.setSubmittedTime(submittedTime);
     }
 
+    public void setCumulatedCoreTime(long cumulatedCoreTime) {
```
fviale, line 1077: this method should also be synchronized

fviale (approved):
LGTM, minor comment

PR #4127, opened 2023-01-13
Upgrade jetty version

fviale:
- upgraded also atmosphere framework to be compatible with jetty 9.4
 - implement necessary api changes
 - add property to control response header size, and increase default request header size
 - prevent LinkageError issues in cloud-automation-service when manipulating jetty api


fviale:
@nebilbenmabrouk yes, this is what I mentioned, programming PR must be merged first in order to allow the scheduling PR to pass

PR #4130, opened 2023-01-19
top-workflows-number-of-used-nodes

oanaBiancaSchiopu:


aminelouati (approved):
I do not see any functional test in this PR. Do you plan to do that in a separate PR?
Not sure about adding a system test as well.
Thanks

oanaBiancaSchiopu:
> I do not see any functional test in this PR. Generally, all public methods should be tested (except getter and setter). Do you plan to do that in a separate PR? Not sure about adding a system test as well. Thanks

@aminelouati I would prefer to add the tests in a separate PR because I would have to do it for all "Top workflows" endpoints.

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/JobInfo.java
```diff
@@ -130,6 +130,12 @@ public interface JobInfo extends Serializable {
      */
     long getCumulatedCoreTime();
 
+    /**
+     * Returns the number of nodes used by the job
```
fviale, line 134: used by the job since the beginning of its execution

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/JobState.java
```diff
@@ -438,6 +438,15 @@ public long getCumulatedCoreTime() {
         return getJobInfo().getCumulatedCoreTime();
     }
 
+    /**
+     * To get the numberOfNodes
```
fviale, line 442: Returns the number of nodes used by the job since the beginning of its execution

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/LiveJobs.java
```diff
@@ -640,6 +640,7 @@ void taskStarted(InternalJob job, InternalTask task, TaskLauncher launcher, Stri
 
         // set the different informations on task
         job.startTask(task);
+        jobData.job.incrementNumberOfNodes();
```
fviale, line 643: unfortunately, it's more complicated than this, as tasks can use multiple nodes

fviale: you can use something like

`jobData.job.increaseNumberOfNodesUsed(task.getExecuterInformation().getNodes().getTotalNumberOfNodes())`

fviale (approved):

PR #4150, opened 2023-03-17
add-stats-submission-mode-endpoint

oanaBiancaSchiopu:


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/dto/JobsSubmissionModeData.java
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * If needed, contact us to obtain a release under GPL Version 2 or 3
+ * or a different license than the AGPL.
+ */
+package org.ow2.proactive_grid_cloud_portal.scheduler.dto;
+
+public class JobsSubmissionModeData implements java.io.Serializable {
```
fviale, line 28: we should not have a dedicated class for this statistic. 

The returned format should be in the form Map<String, Integer>

where keys are "studio", "rest-api", etc and can vary based on what is found in the database

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -1026,15 +1042,43 @@ public List<FilteredTopWorkflow> getTopWorkflowsWithIssues(int numberOfWorkflows
...
+
+            String selectSubQuery = "select submissionMode, count(*) as numberOfJobs FROM JobData where startTime > 0 and finishedTime > 0  ";
+            String subQueryGroupByStatement = "group by submissionMode order by numberOfJobs desc";
+            Query query = getTopWorkflowsQuery(session,
+                                               workflowName,
+                                               bucketName,
+                                               user,
+                                               tenant,
+                                               startTime,
+                                               endTime,
```
fviale, line 1058: startTime and endTime are using finishedTime column inside this method, consider adding an extra parameter to this method to allow switching between finishedTime, submittedTime, startTime, ...

fviale (approved):
LGTM

AlexFalessi (approved):

PR #4159, opened 2023-04-24
Add NODE_SOURCE generic info

fviale:
- allows selecting nodes which belong to the given node source.
 - added functional test and unit test.


andrewssobral (approved):

andrewssobral:
@fviale now we can replace all NODE_ACCESS_TOKEN and NODESOURCENAME by NODE_SOURCE ? (ai workflows)

fviale:
@andrewssobral 

I need 2 other PRs before you can do this.

aminelouati (approved):
What happen if a user define both NODE_SOURCE and NODE_ACCESS_TOKEN at the same time? In this case, NODE_ACCESS_TOKEN should be evaluated first right?

PR #4165, opened 2023-05-09
fix global variables filters

fviale:
- use contains instead of equality as system tests create a service with multiple transitions.


andrewssobral (approved):

fviale:
Retest this please

PR #4167, opened 2023-05-26
Fix not finished job issue when submitting multiple jobs

mbenguig:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/SubmitHandler.java
```diff
@@ -91,6 +91,7 @@ public void run() {
         for (InternalJob job : jobs) {
             if (job != null) {
                 service.getJobs().jobSubmitted(job, timingLogger);
+                frontendState.jobSubmitted(job, ident);
```
fviale, line 94: I believe you need to add
if (forntendState != null) {
}
This is for unit tests

fviale (approved):

fviale (approved):

PR #4171, opened 2023-06-05
create-labels-endpoints

oanaBiancaSchiopu:


rest/rest-api/src/main/java/org/ow2/proactive_grid_cloud_portal/common/SchedulerRestInterface.java
```diff
@@ -2825,4 +2825,116 @@ void addExternalEndpointUrl(@HeaderParam("sessionid") String sessionId, @PathPar
...
+    @Path("labels/jobs/removelabel")
+    @Consumes(value = MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    void removeJobLabels(@HeaderParam("sessionid") String sessionId, List<String> jobIds) throws RestException;
+
+    /**
+     *
+     * Gets all labels.
+     *
+     * @param sessionId
```
fviale, line 2865: you should decribe with @returns what is returned by the various endpoints

rest/rest-client/src/main/java/org/ow2/proactive/scheduler/rest/data/DataUtility.java
```diff
@@ -299,4 +301,14 @@ public static TaskId taskId(JobId jobId, TaskIdData taskIdData) {
         return createTaskId(jobId, taskIdData.getReadableName(), taskIdData.getId());
     }
 
+    public static JobLabelInfo toJobLabelInfo(JobLabelInfoData jobLabel) {
+        return new JobLabelInfo(jobLabel.getId(), jobLabel.getLabel());
+    }
+
+    public static List<JobLabelInfo> toJobLabelsInfo(List<JobLabelInfoData> jobLabel) {
+        List<JobLabelInfo> jobLabelInfo = new LinkedList<>();
```
fviale, line 309: better to use an ArrayList with the same size as the input.

List<JobLabelInfo> jobLabelInfo = new ArrayList<>(jobLabel.size());

fviale:
Retest this please

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3149,4 +3153,130 @@ public Map<Long, List<String>> getPreciousTaskNames(List<String> jobsIds) {
...
+                                   .collect(Collectors.toList());
+        });
+    }
+
+    public List<JobLabelInfo> newLabels(List<String> labels) throws LabelConflictException {
+        List<JobLabelInfo> jobLabelsInfo = new LinkedList<>();
+        for (String label : labels) {
+            if (checkIfLabelExists(label)) {
+                throw new LabelConflictException(label);
+            }
```
AlexFalessi, line 3176: I think this should be done on all the list before sending sql queries.
Here some labels can be created before reaching an already existing label, which throws the exception and the next labels that can be good will not be created.

So either we reject the whole creations or a silent failure type which i don't think is what we want here

oanaBiancaSchiopu: I agree. I will refactor this

AlexFalessi (approved):

PR #4174, opened 2023-06-14
add-label-validation-exception

oanaBiancaSchiopu:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3168,12 +3169,15 @@ public List<JobLabelInfo> getLabels() {
...
     }
 
-    public List<JobLabelInfo> newLabels(List<String> labels) throws LabelConflictException {
+    public List<JobLabelInfo> newLabels(List<String> labels) throws LabelConflictException, LabelValidationException {
         List<JobLabelInfo> jobLabelsInfo = new LinkedList<>();
         for (String label : labels) {
             if (checkIfLabelExists(label)) {
                 throw new LabelConflictException(label);
             }
+            if (label.length() > 20 || !label.matches("^[a-zA-Z0-9_/-]*$")) {
```
fviale, line 3178: could you add a constant to store the regexp ?

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/core/properties/PASchedulerProperties.java
```diff
@@ -564,6 +564,10 @@ negative number indicates that session is infinite (value specified in second) *
 
     public static final String PA_SCHEDULER_PROPERTIES_RELATIVE_FILEPATH = "config/scheduler/settings.ini";
 
+    public static final String LABEL_REGEX = "^[a-zA-Z0-9_/-]*$";
+
+    public static final int LABEL_MAX_LENGTH = 20;
```
fviale, line 569: you need to define a enum value (check the other properties defined in this enum class).

Also, you need to add it in config/settings.ini

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3168,12 +3171,15 @@ public List<JobLabelInfo> getLabels() {
...
     }
 
-    public List<JobLabelInfo> newLabels(List<String> labels) throws LabelConflictException {
+    public List<JobLabelInfo> newLabels(List<String> labels) throws LabelConflictException, LabelValidationException {
         List<JobLabelInfo> jobLabelsInfo = new LinkedList<>();
         for (String label : labels) {
             if (checkIfLabelExists(label)) {
                 throw new LabelConflictException(label);
             }
+            if (label.length() > LABEL_MAX_LENGTH.getValueAsInt() || !label.matches(LABEL_REGEX.getValueAsString())) {
```
fviale, line 3180: as PASchedulerProperties uses lazy loading and a configuration file, it's not a good practice to use import static.

It might work, but I would prefer if you use it directly with e.g. PASchedulerProperties.LABEL_MAX_LENGTH.getValueAsInt()

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -27,6 +27,7 @@
 
 import static org.ow2.proactive.scheduler.core.db.types.FilterJobsMode.FINISHED_AND_STARTED;
 import static org.ow2.proactive.scheduler.core.db.types.FilterJobsMode.SUBMITTED_ONLY;
+import static org.ow2.proactive.scheduler.core.properties.PASchedulerProperties.LABEL_MAX_LENGTH;
```
fviale, line 30: you need to remove the import static

fviale (approved):

AlexFalessi (approved):

PR #4180, opened 2023-06-23
add-logs-for-label-operations

oanaBiancaSchiopu:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3257,16 +3256,19 @@ public JobLabelInfo updateLabel(String labelId, String newLabel)
...
 
     public void deleteLabel(String labelId) throws LabelNotFoundException {
         if (!checkIfLabelIdExists(Long.parseLong(labelId))) {
             throw new LabelNotFoundException(labelId);
         }
+        String label = getLabelById(Long.parseLong(labelId));
         executeReadWriteTransaction(session -> session.getNamedQuery("deleteLabel")
                                                       .setParameter("labelId", Long.parseLong(labelId))
                                                       .executeUpdate());
+        logger.info("Label " + label + " set been deleted");
```
aminelouati, line 3271: "Label " + label + " set been deleted" --> Label " + label + " has been deleted"

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3300,6 +3303,7 @@ public void removeJobLabels(List<String> jobIds) {
                                                                                                .setParameter("jobIdList",
                                                                                                              longIds)
                                                                                                .executeUpdate()));
+            logger.info("Label set been removed from jobs " + jobIds);
```
aminelouati, line 3306: this logger message is not clear

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3192,6 +3189,7 @@ public List<JobLabelInfo> newLabels(List<String> labels) throws LabelConflictExc
             JobLabel jobLabel = executeReadWriteTransaction(session -> {
                 JobLabel labelData = JobLabel.createJobLabel(label);
                 session.save(labelData);
+                logger.info("Label " + label + " has been created");
```
fviale, line 3192: it is better to put this log message inside the frontend caller method.
This way, you can include in the log message the user which created the label

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3211,6 +3209,7 @@ public List<JobLabelInfo> setLabels(List<String> labels) throws LabelValidationE
             JobLabel jobLabel = executeReadWriteTransaction(session -> {
                 JobLabel labelData = JobLabel.createJobLabel(label);
                 session.save(labelData);
+                logger.info("Label " + label + " set been set");
```
fviale, line 3212: same remark as above

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3257,16 +3256,19 @@ public JobLabelInfo updateLabel(String labelId, String newLabel)
                                                       .setParameter("newLabel", newLabel)
                                                       .setParameter("labelId", Long.parseLong(labelId))
                                                       .executeUpdate());
+        logger.info("Label " + newLabel + " set been updated");
```
fviale, line 3259: same

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
fviale: same remark as above

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3284,6 +3286,7 @@ public void setLabelOnJobIds(String labelId, List<String> jobIds) throws LabelNo
                                                                                                .setParameter("jobIdList",
                                                                                                              jobIdSubList)
                                                                                                .executeUpdate()));
+            logger.info("Label " + label + " set been set of jobs " + jobIds);
```
fviale, line 3289: same remark as above

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
fviale: same remark as above

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
oanaBiancaSchiopu: For some operation we don't have the label name in the frontend caller method. We receive the Id as request param

fviale: alright, in that case, you could add the intiator as a parameter to the database methods. This way you would have both information when printing the logs.

fviale (approved):
LGTM

PR #4186, opened 2023-07-05
Add submitted time params

oanaBiancaSchiopu:


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -607,6 +620,22 @@ private int getTotalNumberOfJobs(final DBJobDataParameters params) {
                     queryString.append("and parentId = :parentId ");
                 }
 
+                if (params.getLabel() != null && !params.getLabel().isEmpty()) {
+                    queryString.append("and label like :label ");
+                }
+
+                if (params.getStatus() != null && !params.getStatus().isEmpty()) {
+                    queryString.append("and status like :status ");
```
fviale, line 628: I believe the status is an enum type and thus is stored as an integer in the database (see above the code that you wrote in the getJobs method).
Thus, you cannot use like in this expression.

oanaBiancaSchiopu: Here I have just given the parameter name. 
This is where I set the value:
query.setParameter("status", JobStatus.valueOf(params.getStatus().toUpperCase()));

fviale: you are creating a SQL query string, this query string will be invalid if you use "status like :status"

fviale: I believe you need to have in DBJobDataParameters status defined as an int (obtained by JobStatus.valueOf(status.toUpperCase())))) and then use:
`queryString.append("and status = :status ");`

fviale (approved):
LGTM

PR #4194, opened 2023-07-12
Add domain information and control

fviale:
- Added configuration settings to control which domains can be used during a login
 - store domain information in the JaaS subject
 - when using Multi-Ldap authentication, domain == tenant
 - In windows operating system, the domain list is by default the current windows domain
 - store domain information along the JobData in the scheduler database
 - added endpoints that return the list of configured domains


oanaBiancaSchiopu (approved):

nebilbenmabrouk (approved):
LGTM

PR #4196, opened 2023-07-26
Add nodeTokens selection script binding

fviale:
Allows to retrieve the tokens list from a selection script (useful when there are multiple tokens defined, and the task execution must be restrained to a specific configuration)

implemented modifications in ProbablisticSelectionManager to take into account the fact that the node tokens may vary over time.

Also, allow RMRest model/tokens endpoint to return the list of tokens defined inside a node source userAccessType (and not only present inside running nodes).


andrewssobral (approved):
✅

PR #4198, opened 2023-08-02
Add robots.txt to the root of the server homepage

amouhoub:


fviale:
did we say to remove also the one in dist/war/rest ?

amouhoub:
> did we say to remove also the one in dist/war/rest ?

@fviale Strange, I don't see it in the code...

fviale (approved):

mklkun (approved):

PR #4202, opened 2023-08-17
change-error-message-when-invalid-domain

oanaBiancaSchiopu:


common/common-client/src/main/java/org/ow2/proactive/authentication/AuthenticationImpl.java
```diff
@@ -172,7 +172,7 @@ public Subject authenticate(Credentials cred) throws LoginException {
             }
             //Nature of exception is hidden for user, we don't want to inform
             //user about the reason of non authentication
-            throw new LoginException("Authentication failed");
+            throw new LoginException(e.getMessage());
```
fviale, line 175: why not do:
throw new LoginException("Authentication failed" + e.getMessage(), e);

fviale (approved):

PR #4219, opened 2023-10-24
replace-variables-model-in-validate-signal-request

oanaBiancaSchiopu:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/factories/spi/model/DefaultModelJobValidatorServiceProvider.java
```diff
@@ -119,8 +121,19 @@ public void validateVariables(List<JobVariable> variableList, Map<String, Serial
                                                                                                                      variableReplacement));
             updatedVariables.put(jobVariable.getName(), jobVariable.getValue());
         });
+
+        Map<String, Serializable> updatedVariablesModel = new LinkedHashMap<>();
```
fviale, line 125: I think you can use a single map with type <String, String>

fviale (approved):

amouhoub (approved):

PR #4245, opened 2024-02-16
Add Groovy class cleanup mechanism

fviale:
- dependency to groovy added into common-client
 - count running groovy scripts and take a snapshot of existing classes when no script is running
 - Use of a scheduled background thread to perform actual cleanup.
 - cleanup calls System.gc() to effectively remove classes from the JVM memory
 - cleanup is disabled inside the ProActive server (only enabled inside nodes). Cleanup can be disabled at will via a java property.

### DIFF
--- a/common/common-client/build.gradle
+++ b/common/common-client/build.gradle
@@ -6,6 +6,9 @@ dependencies {
     compile "org.objectweb.proactive:programming-extension-annotation:${programmingVersion}"
 
     compile project(':common:common-api')
+
+    // this dependency is added to allow groovy class cleanup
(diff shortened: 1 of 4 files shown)


nebilbenmabrouk (approved):
LGTM, just one question: What about other JVM languages like Java ? Do we do the same ?

fviale:
@nebilbenmabrouk no this is only a groovy issue (AFAIK), and the fix only applies to groovy class loading.

fviale:
retest this please

AlexFalessi (approved):

aminelouati (approved):

PR #4251, opened 2024-02-27
install_base.sh: Copy config/authentication/truststore from previous to default

mboussaa:


tools/install_base.sh
```diff
@@ -92,9 +92,14 @@ reuse_accounts()
 
     echo "Retrieving keystore files from previous scheduler installation"
     /bin/cp $OLD_PADIR/config/web/keystore $PA_ROOT/default/config/web/
+
+    echo "Retrieving truststore files from previous scheduler installation"
```
fviale, line 96: print the echo message inside the if

fviale: forget about it, it's ok like it is

fviale (approved):
LGTM

PR #4264, opened 2024-05-07
Be able to kill a set of jobs even it contains a terminated job

aminelouati:


fviale (approved):

fviale:
in the javadoc, change the sentence:
"@return list of locked jobs for each job id provided, otherwise null"

PR #4266, opened 2024-05-22
Add priority list in UserData

fviale:
- include the list of priorities that this user is allowed to enable


AlexFalessi (approved):

aminelouati (approved):

AlexFalessi:
Is the user type permission taken into account with the endpoint  /rest/scheduler/jobs/permission/methods/ ?

PR #4267, opened 2024-05-30
Return an error if a user tries to update the priority of a job that …

AlexFalessi:
…has a bad value

### DIFF
--- a/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
+++ b/rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
@@ -2475,9 +2475,11 @@ public class SchedulerStateRest implements SchedulerRestInterface {
             throws RestException, JobAlreadyFinishedRestException {
         try {
             Scheduler s = checkAccess(sessionId, PATH_JOBS + jobId + "/priority/byname/" + priorityName);
-            s.changeJobPriority(jobId, JobPriority.findPriority(priorityName));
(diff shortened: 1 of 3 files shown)


nebilbenmabrouk (approved):
LGTM

rest/rest-server/src/main/java/org/ow2/proactive_grid_cloud_portal/scheduler/SchedulerStateRest.java
```diff
@@ -2435,11 +2437,25 @@ public void schedulerChangeJobPriorityByValue(String sessionId, String jobId, St
             throws RestException, JobAlreadyFinishedRestException {
         try {
             Scheduler s = checkAccess(sessionId, PATH_JOBS + jobId + "/priority/byvalue" + priorityValue);
-            s.changeJobPriority(jobId, JobPriority.findPriority(Integer.parseInt(priorityValue)));
-        } catch (SchedulerException e) {
-            throw RestException.wrapExceptionToRest(e);
+            JobPriority newPriority = Arrays.stream(JobPriority.values())
```
fviale, line 2440: a more elegant solution would be to add a findPriorityExact static method in JobPriority that throws an IllegalArgumentException if not found

AlexFalessi: Ok i'll implement this.

AlexFalessi:
@fviale I'll rollback the usage of a static map if you don't like it.

fviale:
@AlexFalessi +1 to remove the map :-)

fviale (approved):

AlexFalessi:
retest this please

PR #4270, opened 2024-06-13
Add findRootTasks method in TaskFlowJob

fviale:


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/TaskFlowJob.java
```diff
@@ -170,11 +165,41 @@ public Set<Task> findTerminalTasks() {
...
+
+        for (Task task : tasks.values()) {
+
+            FlowScript flowScript = task.getFlowScript();
+            if (flowScript != null && flowScript.getActionType().equals(FlowActionType.IF.toString())) {
+                if (flowScript.getActionTarget() != null) {
+                    // remove from root all tasks that are target of this task's IF, ELSE or CONTINUATION branches
+                    rootTasks.removeIf(t -> t.getName().equals(flowScript.getActionTarget()));
+                    rootTasks.removeIf(t -> t.getName().equals(flowScript.getActionTargetElse()));
+                    rootTasks.removeIf(t -> t.getName().equals(flowScript.getActionContinuation()));
```
AlexFalessi, line 192: You can do it in a single `removeIf()`, this will avoid to potentially call two more uncessary removeIf. Also you can read the removeIf result and if true add a continue to bypass second if.

nebilbenmabrouk (approved):
LGTM

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/job/TaskFlowJob.java
fviale: right

fviale: though both if are independant, I will do the first change you mention

AlexFalessi: Yes you are right. In any case it would not have been a valuable change as you would still have to test the `removeIf ` result.

PR #4276, opened 2024-07-02
add NodeUsageSchedulingPolicy

fviale:
This scheduling policy allows handling jobs submitted with the MAX_NODES_USAGE generic information.

When such job is submitted, the policy ensures that no more than MAX_NODES_USAGE Nodes can be used in parallel by this job and all of its sub jobs.


mbenguig (approved):

aminelouati (approved):
Are you considering all the hierarchy of sub-jobs or only the first level ?

fviale:
@aminelouati all the hierarchy

PR #4279, opened 2024-07-10
Prevent duplicates when creating labels. 

AlexFalessi:
Also ensure front-end backward compatiblity by returning unique results when fetching labels from DB that have same label name

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/Scheduler.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/Scheduler.java
@@ -2099,7 +2099,7 @@ public interface Scheduler extends SchedulerUsage, ThirdPartyCredentials, Servic
      */
     @RoleAdmin
     List<JobLabelInfo> setLabels(List<String> labels)
-            throws NotConnectedException, PermissionException, LabelValidationException;
+            throws NotConnectedException, PermissionException, LabelValidationException, LabelConflictException;
 
     /**
      * Update the given label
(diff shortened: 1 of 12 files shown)


scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/Scheduler.java
```diff
@@ -2099,7 +2099,7 @@ List<JobLabelInfo> createLabels(List<String> labels)
      */
     @RoleAdmin
     List<JobLabelInfo> setLabels(List<String> labels)
-            throws NotConnectedException, PermissionException, LabelValidationException;
+            throws NotConnectedException, PermissionException, LabelValidationException, LabelConflictException;
```
fviale, line 2102: why not reuse LabelValidationException ?

AlexFalessi: It seemed to me that it's not really a label validation exception, which is thrown when there is a formatting issue on a label (special characters).
Should I use the LabelValidationException ?

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
```diff
@@ -3227,35 +3230,37 @@ private boolean isLabelValid(String label) {
     public List<JobLabelInfo> newLabels(List<String> labels, String username)
             throws LabelConflictException, LabelValidationException {
         List<JobLabelInfo> jobLabelsInfo = new LinkedList<>();
-        for (String label : labels) {
+        Set<String> uniqueLabels = new HashSet<>(labels);
```
fviale, line 3233: you should use linkedhashset also here to preserve the order

fviale (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/SchedulerDBManager.java
AlexFalessi: OK done.

aminelouati (approved):

PR #4283, opened 2024-07-16
Refactor POST and DELETE credentials to take key as form param. Remov…

AlexFalessi:
…e @Encode annotation and manual decode() for automatic decoding of key.


fviale (approved):

nebilbenmabrouk (approved):
LGTM

PR #4284, opened 2024-07-19
Improve performance of ExtendedSchedulerPolicy

fviale:
- use an LRU cache for getStartAtValue (this prevents costly operations over the same job and task)
 - changed the key of delayedJobsWakeUpCache to be the startAt time instead of job id
 - added timer loggers to monitor performances when the logger is in debug

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/ExtendedSchedulerPolicy.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/ExtendedSchedulerPolicy.java
@@ -35,6 +35,7 @@ import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.collections4.map.LRUMap;
+import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
 import org.objectweb.proactive.utils.NamedThreadFactory;
 import org.ow2.proactive.scheduler.common.JobDescriptor;
@@ -44,6 +45,9 @@ import org.ow2.proactive.scheduler.core.properties.PASchedulerProperties;
 import org.ow2.proactive.scheduler.descriptor.EligibleTaskDescriptor;
 import org.ow2.proactive.scheduler.descriptor.EligibleTaskDescriptorImpl;
 import org.ow2.proactive.scheduler.descriptor.JobDescriptorImpl;
+import org.ow2.proactive.scheduler.util.MultipleTimingLogger;
+
+import com.google.common.annotations.VisibleForTesting;
 
 
 /**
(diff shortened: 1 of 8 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/ExtendedSchedulerPolicy.java
```diff
@@ -121,21 +138,28 @@ public LinkedList<EligibleTaskDescriptor> getOrderedTasks(List<JobDescriptor> jo
...
     }
 
-    private void scheduleWakeUp(final String taskOrJobId, Date now, Date startAtDate) {
-        if (!startAtCache.containsKey(taskOrJobId)) {
+    private void scheduleWakeUp(Date now, Date startAtDate) {
+        final long scheduledTime = startAtDate.getTime();
+        if (!delayedJobsWakeUpCache.containsKey(scheduledTime)) {
             executor.schedule(() -> {
-                startAtCache.remove(taskOrJobId);
+                delayedJobsWakeUpCache.remove(scheduledTime);
```
AlexFalessi, line 153: Not necessary as you are in `if (!delayedJobsWakeUpCache.containsKey(scheduledTime))`

AlexFalessi (approved):
Small comments

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/ExtendedSchedulerPolicy.java
fviale: hehe, look at the code, it does not do it immediately, it schedules the operation.
When the operation is performed, the key must be removed

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/license/LicenseSchedulingPolicy.java
```diff
@@ -156,71 +164,81 @@ private void removeFinishedTasksAndReleaseTokens(String requiredLicense) {
...
                 return true;
+            } else { // == 0
+                logger.debug("Removing all terminated jobs/tasks and check again if there are remaining licenses");
+                removeFinishedJobsAndReleaseTokens(currentRequiredLicense);
+                removeFinishedTasksAndReleaseTokens(currentRequiredLicense);
+
+                if (licenseSynchronization.getRemainingTokens(currentRequiredLicense) > 0) {
+                    logger.debug("There are remaining licenses after releasing tokens, return true");
+                    return true;
+                }
```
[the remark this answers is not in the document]
fviale: I didn't touch this code, only wrapped it with timing loggers.

and yes I believe we can enter it if getRemainingTokens return 0

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/policy/ExtendedSchedulerPolicy.java
AlexFalessi: Oh yes sorry ...

PR #4285, opened 2024-07-22
Fix updateStartAt

fviale:
When updating the startAt value of a pending job, the cache used by ExtendedSchedulerPolicy need to be updated.

Accordingly, the StartAtUpdater must be able to retrieve the policy instance and ask it to update the startAt value of a given job or task.

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/StartAtUpdater.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/StartAtUpdater.java
@@ -32,19 +32,36 @@ import java.util.Set;
 
 import org.ow2.proactive.scheduler.common.task.TaskId;
 import org.ow2.proactive.scheduler.common.util.ISO8601DateUtil;
+import org.ow2.proactive.scheduler.core.SchedulingService;
 import org.ow2.proactive.scheduler.core.db.SchedulerDBManager;
 import org.ow2.proactive.scheduler.job.InternalJob;
 import org.ow2.proactive.scheduler.policy.ExtendedSchedulerPolicy;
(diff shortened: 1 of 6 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/StartAtUpdater.java
```diff
@@ -56,30 +73,39 @@ public boolean updateStartAt(InternalJob job, String startAt, SchedulerDBManager
...
+                    schedulerPolicy.updateStartAt(td.getId().toString(), startAt);
+                }
             }
         }
 
         return updatedTasks;
     }
 
-    private boolean resetJobGenericInformation(InternalJob job, String startAt) {
+    private boolean resetJobGenericInformation(InternalJob job, String startAt,
```
mboussaa, line 97: I don't see a case when it will be false

mboussaa: ok, only when isValidStartAt(genericInformation, startAt) returns false the fucntion will return false

mboussaa (approved):

mbenguig (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/StartAtUpdater.java
fviale: I don't understand to which code you refer to ...

fviale: @mboussaa can I merge my PR as it is ?

PR #4286, opened 2024-07-25
Add startAt to JobData and Jobinfo

fviale:
startAt is stored in the database and searchable

ensure that changeStartAt request updates the value stored in JobData and the generic info

add a changeStartAt request that accepts multiple jobs


AlexFalessi (approved):

fviale:
Retest this please

PR #4287, opened 2024-07-29
Add description to JobInfo

fviale:


scheduler/scheduler-server/src/test/java/performancetests/metrics/TaskCreationTimeTest.java
```diff
@@ -57,7 +57,7 @@ public class TaskCreationTimeTest extends PerformanceTestBase {
      */
     @Parameterized.Parameters
     public static Collection<Object[]> data() {
-        return Arrays.asList(new Object[][] { { 20000, 100000 } });
+        return Arrays.asList(new Object[][] { { 20000, 110000 } });
```
mbenguig, line 60: 20000 could have be explained in a very short comment .. but not blocking

mbenguig (approved):

amouhoub (approved):

scheduler/scheduler-server/src/test/java/performancetests/metrics/TaskCreationTimeTest.java
fviale: this increase is because my last commit added one column to JobData which makes this performance test slightly slower

PR #4291, opened 2024-08-26
SchedulerNodeClient: when connection fails, use the scheduler rest pu…

fviale:
…blic url when configured

When connection to the scheduler fails (for example, when the scheduler is behind a reverse proxy), and a scheduler rest public url is configured, the scheduler client will attempt to connect using the public url.

### DIFF
--- a/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/client/SchedulerNodeClient.java
+++ b/scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/client/SchedulerNodeClient.java
@@ -30,6 +30,7 @@ import static org.ow2.proactive.scheduler.common.SchedulerConstants.PARENT_JOB_I
 import java.io.File;
 import java.io.Serializable;
 import java.net.URL;
+import java.security.KeyException;
 import java.util.Collections;
 import java.util.Date;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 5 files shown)


scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/client/SchedulerNodeClient.java
```diff
@@ -136,19 +144,56 @@ public void connect() throws Exception {
      * @throws Exception
      */
     public void connect(String url) throws Exception {
+        connect(url, SchedulerNodeClient.this.schedulerRestUrl, SchedulerNodeClient.this.schedulerRestPublicUrl);
+    }
+
+    private void connect(String url, String backupUrl1, String backupUrl2) throws Exception {
```
nebilbenmabrouk, line 150: Why do  we need a **backupUrl2** ? _(I see it is set to null)_

nebilbenmabrouk (approved):
LGTM (with a small question)

scheduler/scheduler-node/src/main/java/org/ow2/proactive/scheduler/task/client/SchedulerNodeClient.java
fviale: it is used when we do schedulerapi.connect(url)

```
public void connect(String url) throws Exception {
        connect(url, SchedulerNodeClient.this.schedulerRestUrl, SchedulerNodeClient.this.schedulerRestPublicUrl);
 }
```

PR #4298, opened 2024-11-05
Fix possible ConcurrentModificationException that can occur when seri…

fviale:
…alizing SchedulerUsers

### DIFF
--- a/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/SchedulerUsers.java
+++ b/scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/SchedulerUsers.java
@@ -25,6 +25,9 @@
  */
 package org.ow2.proactive.scheduler.common;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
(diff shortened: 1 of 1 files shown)


ankicabarisic (approved):
looks good

you can eventually consider change `(HashSet)` to `(HashSet<UserIdentification>) `at the cast in `readObject` to avoid raw types and improve type safety.

fviale:
@ankicabarisic ok, done

scheduler/scheduler-api/src/main/java/org/ow2/proactive/scheduler/common/SchedulerUsers.java
```diff
@@ -98,4 +101,12 @@ public void update(UserIdentification user) {
             users.add(user);
         }
     }
+
+    private void readObject(ObjectInputStream aInputStream) throws ClassNotFoundException, IOException {
+        users = Collections.synchronizedSet((Set) aInputStream.readObject());
+    }
+
+    private void writeObject(ObjectOutputStream aOutputStream) throws IOException {
+        aOutputStream.writeObject(new HashSet<>(users));
```
AlexFalessi, line 110: These methods are not used ?

PR #4300, opened 2024-11-08
Change passwords stored in login.cfg to hash/salt format

fviale:
- new implementation uses the library https://github.com/patrickfav/bcrypt which implements OpenBSD Blowfish 2 password hashing algorithm. This algorithm has the advantage of not using salt tables (which cannot be used in our context).
 - note that Blowfish supports passwords of a maximum of 72 bytes
 - Add command to regenerate passwords, which includes automatic conversion to the new format
 - this command is called automatically from install.sh when appropriate


fviale:
PR: https://github.com/ow2-proactive/scheduling/pull/4300

ankicabarisic (approved):

tools/install_base.sh
```diff
@@ -61,18 +61,42 @@ escape_rhs_sed ()
...
+    NUMERIC_VERSION=$(echo ${1/-SNAPSHOT})
+    echo ${NUMERIC_VERSION%.*}
+}
+
+# an upgrade of the password stored format occurs when the new version is 14.1 or greater and the old version is 14.0 or lower
+is_legacy_hash_upgrade()
+{
+  NEW_VERSION_LEGACY_SETTING=$( grep 'pa.rm.legacy.encryption=' $PA_ROOT/default/config/rm/settings.ini | head -n 1 | cut -d '=' -f2 )
+  OLD_VERSION_LEGACY_SETTING=$( grep 'pa.rm.legacy.encryption=' $PA_ROOT/previous/config/rm/settings.ini | head -n 1 | cut -d '=' -f2 )
+  if [ "$NEW_VERSION_LEGACY_SETTING" == "" ]; then
```
mboussaa, line 87: if [ -z "$NEW_VERSION_LEGACY_SETTING" ]; then

tools/install_base.sh
```diff
@@ -61,18 +61,42 @@ escape_rhs_sed ()
...
+    NEW_VERSION_LEGACY_SETTING="true"
+  fi
+  if [ "$OLD_VERSION_LEGACY_SETTING" == "" ]; then
```
mboussaa, line 90: same

tools/install_base.sh
```diff
@@ -61,18 +61,42 @@ escape_rhs_sed ()
...
+    OLD_VERSION_LEGACY_SETTING="true"
+  fi
+  if [ "$NEW_VERSION_LEGACY_SETTING" == "false" ] & [ "$OLD_VERSION_LEGACY_SETTING" == "true" ]; then
+    return 0
+  else
```
mboussaa, line 95: the else involves:
false && false
true && true
true && false
false && false
I don't think it is good ...

tools/install_base.sh
fviale: ok

tools/install_base.sh
```diff
@@ -61,18 +61,42 @@ escape_rhs_sed ()
...
+{
+  NEW_VERSION_LEGACY_SETTING=$( grep 'pa.rm.legacy.encryption=' $PA_ROOT/default/config/rm/settings.ini | head -n 1 | cut -d '=' -f2 )
+  OLD_VERSION_LEGACY_SETTING=$( grep 'pa.rm.legacy.encryption=' $PA_ROOT/previous/config/rm/settings.ini | head -n 1 | cut -d '=' -f2 )
+  if [ "$NEW_VERSION_LEGACY_SETTING" == "" ]; then
+    NEW_VERSION_LEGACY_SETTING="true"
+  fi
+  if [ "$OLD_VERSION_LEGACY_SETTING" == "" ]; then
+    OLD_VERSION_LEGACY_SETTING="true"
+  fi
+  if [ "$NEW_VERSION_LEGACY_SETTING" == "false" ] & [ "$OLD_VERSION_LEGACY_SETTING" == "true" ]; then
```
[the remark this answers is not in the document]
fviale: && does not work here, I don't understand why.

with &
`fviale@Olivia:~$ ./test.sh
false
true
0` => correct

with &&

`fviale@Olivia:~$ ./test.sh
false
true
1` => not correct

tools/install_base.sh
```diff
@@ -759,6 +785,9 @@ if ls $PA_ROOT/default/addons/*.jar > /dev/null 2>&1; then
         if [[ "$OLD_VERSION" != "$NEW_VERSION" ]]; then
            rm -f  $PA_ROOT/default/addons/*$OLD_VERSION*.jar
            rm -f  $PA_ROOT/default/addons/*/*$OLD_VERSION*.jar
+           OLD_VERSION_MAJOR=$(compute-major-minor OLD_VERSION)
+        else
+           OLD_VERSION_MAJOR=NEW_VERSION_MAJOR
```
mboussaa, line 790: I don't understand these OLD_VERSION_MAJOR= 

OLD_VERSION_MAJOR is nowhere used

tools/install_base.sh
fviale: we need to upgrade when the new version has pa.rm.legacy.encryption set to false and the old version has it set to true or absent

fviale: legacy.encryption is set to true when we enable backward compatibility mode with the old encryption mechanism

tools/install_base.sh
fviale: true, I used this before I decided to rely on looking at the config/rm/settingsini configuration, I can remove it.

tools/install_base.sh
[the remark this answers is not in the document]
fviale: Similarly when I use the following:

`if [ "$NEW_VERSION_LEGACY_SETTING" == "false" -a "$OLD_VERSION_LEGACY_SETTING" == "true" ]; then
    return 0
  else
    return 1
  fi`
it returns 1 in the above example, if you can help me understand ...

config/authentication/login.cfg
```diff
@@ -1,16 +1,16 @@
...
+nsadmin2:$2a$12$ml3UHl8QeUxgjMFcAGTB8OGDvl4tGYZA7LBBKgEEyQ4x7Qp0lMwTS
+provider:$2a$12$a94e.v2qzMPKp7nn/CMMFu3G28WdvWSmWs98fPbVJeiJPpm6WCfc.
+radmin:$2a$12$/uMxFHAPidOcmjemfmv/DOLEUTx5L3cJKV/wKS76yeKugzB21t106
+rm:$2a$12$reLZNyffRQmZ1ACZNn85x.rosnt4AXEnsn/tfEq6bxw0mCT2Qwbl6
+scheduler:$2a$12$.cmyzTeCiFo8pxr6k9mCF.rNqwUMS2OY2zwlRh1lTQbjkqePd9O5q
+subadmin:$2a$12$/V4AQYyeR2U9EyCQ8hq52.BN/Xul1vSzIy0FuGil6/xsuoRjufUw2
+test:$2a$12$BsCEuxiiGFiPXh2Bu3EzfeVlV09O2nIfXhq28HuLOUmruRfbhUUZ.
+test_executor:$2a$12$LmF/0h1e8AEWV6YQkLqNxOZ4IvquPOR5Nf/yw2/KuEJaPS2bSo1..
+user:$2a$12$GWJnc6LaQNNHoCtusN626efZiCEWCA5rBPA7jBeLiWgERmxOlTs8C
+watcher:$2a$12$ko6Hf7SX.dqTY.zYKFUBLOrM1Kx8Ujuvb8uGAVKUTgvRmzZYCXdUG
```
mboussaa, line 16: Many users are not uptodate with the doc like test users:
https://trydev2.activeeon.com/doc/admin/ProActiveAdminGuide.html#_groups

mboussaa (approved):

config/authentication/login.cfg
fviale: it's the doc that's not uptodate :-)

PR #4304, opened 2024-11-18
support the execution of stored procedure with and without parameters

aminelouati:


AlexFalessi (approved):

AlexFalessi:
So the stored procedures we want to execute have to be created before using some SQL scripts right ?

fviale (approved):

PR #4309, opened 2024-12-06
Add endpoint to change logo. Image must be PNG or JPEG

AlexFalessi:
Add an endoint to change the enteprise logo on the portals.

Image must:
- Be PNG or JPEG
- Have a max size of 1500 x 1500

The scheduler will check MIME type (deducted by reading first bytes of data) and size

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/LogoValidator.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/LogoValidator.java
@@ -41,7 +41,7 @@ public class LogoValidator {
     private static final int MAX_HEIGHT = 1500;
 
     /**
-     *   Checks that the input image is of the correct format (PNG or JPEG), size and MIME type.
+     *   Checks that the input image is of the correct format (PNG or Baseline JPEG), size and MIME type.
      *
      *   @param byteArrayImage the PNG or JPEG image as a byte array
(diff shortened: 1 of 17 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/LogoValidator.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public static byte[] checkImage(byte[] byteArrayImage) throws IOException {
+
+        ImageInfo imageInfo = Imaging.getImageInfo(byteArrayImage);
+        BufferedImage image = Imaging.getBufferedImage(byteArrayImage);
+
+        byte[] result = byteArrayImage;
+
+        // If the image is a JPEG, convert it to PNG
+        if (imageInfo.getFormat() == ImageFormats.JPEG) {
```
fviale, line 57: for code clarity, it's better to use a switch case

fviale (approved):
Minor comment

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/helpers/LogoValidator.java
AlexFalessi: Actually I can't as `imageInfo.getFormat()` returns an `public interface ImageFormat`

PR #4317, opened 2025-01-17
Add shadow account mechanism

fviale:
A shadow account is an internal account (FileLoginModule) which masks a third-party account (ldap, keycloak). The shadow account uses a randomly generated password.

 - a shadow account is created whenever the corresponding property is set to true in either keycloak or ldap configuration file
 - this shadow account's credentials are used instead of the original credentials in either the scheduler, rm or rest server
 - moved logic of create credentials to FileLoginModule, the created credentials are the shadow credentials when shadow account is configured
 - after the shadow account has been created, whenever a user logs in again from ldap or keycloak, a new account is not created and the original shadow account is reused. Groups associated with the shadow account are updated if necessary.


fviale:
Retest this please

oanaBiancaSchiopu (approved):

common/common-client/src/main/java/org/ow2/proactive/authentication/LDAPLoginModule.java
```diff
@@ -29,8 +29,7 @@
 import java.security.SecureRandom;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
-import java.util.Hashtable;
-import java.util.Map;
+import java.util.*;
```
AlexFalessi, line 32: import whole util package ?

AlexFalessi (approved):
Good for me, small non blocking comment.

common/common-client/src/main/java/org/ow2/proactive/authentication/LDAPLoginModule.java
fviale: probably because I added some classes for tests that I removed after, but it's not a big deal

PR #4319, opened 2025-01-28
SchedulerDBManager: improve loadJobs performance

fviale:
- increase default batch size to 1000
 - load jobs from the database in parallel. Number of threads is configurable by a new property pa.scheduler.core.loadjobs.nbthreads
 - increase default database connection timeout


AlexFalessi (approved):

fviale:
Retest this please

PR #4321, opened 2025-02-07
Store script urls in database by reference

fviale:
When this option is enabled, script urls which contains a catalog url will be stored in the db by using $PA_CATALOG_REST_URL/path and this expression will be resolved to a local url when the job is loaded from the db.

To use this option, it is mandatory to set the catalog url manually in config/scheduler/settings.ini.

This is to prevent issues in HA active/passive of two servers using the same database

### DIFF
--- a/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/ScriptData.java
+++ b/scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/ScriptData.java
@@ -40,6 +40,8 @@ import org.hibernate.type.SerializableToBlobType;
 import org.ow2.proactive.db.DatabaseManagerException;
 import org.ow2.proactive.scheduler.common.task.flow.FlowActionType;
 import org.ow2.proactive.scheduler.common.task.flow.FlowScript;
+import org.ow2.proactive.scheduler.core.properties.PASchedulerProperties;
+import org.ow2.proactive.scheduler.task.SchedulerVars;
 import org.ow2.proactive.scripting.InvalidScriptException;
 import org.ow2.proactive.scripting.Script;
 import org.ow2.proactive.scripting.SimpleScript;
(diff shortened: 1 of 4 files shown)


scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/ScriptData.java
```diff
@@ -186,13 +188,37 @@ protected static void initCommonAttributes(ScriptData scriptData, Script<?> scri
...
         if (script.getScriptUrl() != null) {
-            scriptData.setURL(script.getScriptUrl().toExternalForm());
+            scriptData.setURL(getDBUrl(script.getScriptUrl()));
         }
         if (script.getParameters() != null) {
             scriptData.setScriptParameters(Arrays.asList(script.getParameters()));
         }
     }
 
+    public static String getDBUrl(URL scriptUrl) {
```
AlexFalessi, line 198: Can be protected or with no modifier

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/ScriptData.java
```diff
@@ -186,13 +188,37 @@ protected static void initCommonAttributes(ScriptData scriptData, Script<?> scri
...
+        String answer = scriptUrl.toExternalForm();
+        String catalogUrl = PASchedulerProperties.CATALOG_REST_URL.getValueAsString();
+        if (PASchedulerProperties.STORE_CATALOG_REF_IN_DB.getValueAsBoolean() && catalogUrl != null &&
+            answer.contains(catalogUrl)) {
+            return answer.replace(catalogUrl, "$" + SchedulerVars.PA_CATALOG_REST_URL.name());
+        }
+        return answer;
+    }
+
+    public static URL getScriptUrl(String dbUrl) throws MalformedURLException {
```
AlexFalessi, line 208: Same

AlexFalessi (approved):

scheduler/scheduler-server/src/main/java/org/ow2/proactive/scheduler/core/db/ScriptData.java
fviale: ok

mboussaa (approved):

PR #4325, opened 2025-03-13
Add group manager read and write endpoints

fviale:
Also:
 - added RoleAdminPermission (ability to read and write all groups) and RoleReaderPermission (ability to read its own group permissions)
 - regenerated security.java.policy-server using the api
 - in order to bypass cyclic dependencies between rest-server and scheduler-server, JAASParser is initialized inside scheduler-server and passed to CommonRest as a parameter interface
 - small unit test


fviale:
Retest this please

